Issue #2, opened 2016-01-03
Parse command line args


dbrgn:
Probably using the docopt crate.

PR #5, opened 2016-01-06
Argument Parsing with Docopt

dbrgn:
- Implemented parsing and render / update commands.
- Moved formatter to separate module.


dbrgn:
Fixes #2.

PR #4, opened 2016-01-06
Rething caching

dbrgn:
Right now I'm trying to follow the caching conventions used by the node client (main directory ~/.tldr/cache`, containing `LICENSE.md` and the `pages/` directory).

The problem is that this convention is not written down anywhere. If the two clients make different decisions, we could end up overwriting the cache of each other.

I thought about proposing a standard, but in the end it's probably the easiest if we simply dump the git export into our own cache (following the XDG standard at `~/.cache/tldr-rs` for POSIX systems). We could either dump the files into that directory directly, or alternatively put them into a `checkout` subdir.

The huge advantage of this is that we can scrap the `index.json` file [which is unreliable](https://github.com/tldr-pages/tldr/issues/487) and get rid of a lot of complicated code targeted at deserializing that file and copying the pages to the cache.

Instead, our own cache would probably make more sense, and it will always be up to date. And during the cache rebuilding, we could also drop commands not supported by the current architecture.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,6 @@ dependencies = [
  "log 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-serialize 0.3.16 (registry+https://github.com/rust-lang/crates.io-index)",
  "tar 0.3.2 (git+https://github.com/dbrgn/tar-rs?branch=pax_header)",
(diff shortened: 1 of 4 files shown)


dbrgn:
16 lines added, 122 lines removed, that's the way I like it :)

This code is actually approved by my cats:

PR #19, opened 2016-04-02
Fix cache updating

### LINKED ISSUE #18: tldr-rs warns "Cache wasn't updated in 30 days" regardless of when cache was updated
It never used to do this, so I suppose I may have ticked over the 30-day timer. But I have updated frequently since I installed the program, so it looks like `tldr --update` doesn't properly reset the timer.
```
# walter @ warg in ~ [20:44:06] 
$ tldr --update
Successfully updated cache.

# walter @ warg in ~ [20:44:24] 
$ tldr sl
Cache wasn't updated in 30 days.


  Steam locomotive running through your terminal.

  - Let a steam locomotive run through your terminal:

  sl

  - The train burns, people scream:

  sl -a

  - Let the train fly:

  sl -F
```

dbrgn:
For some reason, the cache was not actually being updated because the
`Archive.unpack()` method doesn't overwrite any files.

So now we clear the cache directory before unpacking the archive.

Fixes #18.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -53,7 +53,7 @@ impl Cache {
         Ok(Archive::new(decoder))
     }
 
-    /// Update the pages cache. Return the number of cached pages.
+    /// Update the pages cache.
     pub fn update(&self) -> Result<(), TldrError> {
         // First, download the compressed data
         let response = try!(self.download());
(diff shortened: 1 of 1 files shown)


dbrgn:
Build fails only on Beta / OS X due to a problem in tar-rs. We can ignore that for now.

PR #28, opened 2016-04-18
Highlighting installation instructions

igorshubovych:


dbrgn:
Thanks! It doesn't make a difference though, does it? The bash-rendered version doesn't include colors either, as it doesn't include any variables or something like that.

README.md
```diff
@@ -51,13 +51,17 @@ High level project goals:
...
+```bash
+$ cargo install tealdeer
+```
 
 ...or from source (see section "Building" in this README).
 
 If you're an Arch Linux user, you can also install the package from the AUR:
 
+```bash
     $ yaourt -S tealdeer-git
```
SShrike: The indentation is a little screwy here, you should probably fix it.

igorshubovych:
http://take.ms/RuT81

dbrgn:
Ah, I see! Didn't notice. The reason is that the line was indented 3 instead of 4 characters.

Can you change the style back to using indentation only (fixing the 3 spaces to 4 spaces), instead of using the bash backticks? Then I'll merge :) The bash backticks don't result in any syntax highlighting, but make the text file harder to read.

dbrgn:
Thanks @igorshubovych!

PR #29, opened 2016-04-18
Use rust-xdg for cache creation

### LINKED ISSUE #10: Evaluate using xdg-basedir crate
The [xdg-basedir](https://kiran-kp.github.io/xdg-basedir/xdg_basedir/index.html) crate could make the $XDG_CACHE lookup (usually `~/.cache`) more standards compliant.

A `get_dir` function could be conditionally-compiled for unix and non-unix platforms. On unix, the $XDG_CACHE lookup would be used, on Windows a directory in the user profile that doesn't contain a dot would be used.

SShrike:
Pretty small PR that closes #10.

I believe this PR also fixes a bug you might not of noticed, where if a user changed `$XDG_CACHE_HOME`, tealdeer would still create the cache under `~/.cache`. Correct me if I'm wrong though.

I've also gone through and removed the spaces like [you said](https://github.com/dbrgn/tealdeer/issues/10#issuecomment-211429317).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,17 +17,18 @@ logging = ["env_logger"]
 dev = ["clippy"]
 
 [dependencies]
-ansi_term = "^0.7"
 log = "^0.3"
+xdg = "^2.0"
 tar = "^0.4"
-flate2 = "^0.2"
 curl = "^0.2"
-rustc-serialize = "^0.3"
-docopt = "^0.6"
 time = "^0.1"
+flate2 = "^0.2"
+docopt = "^0.6"
 walkdir = "^0.1"
-env_logger = { version = "^0.3", optional = true }
+ansi_term = "^0.7"
+rustc-serialize = "^0.3"
 clippy = {version = "~0.0.37", optional = true}
+env_logger = { version = "^0.3", optional = true }
 
 [dev-dependencies]
 tempdir = "^0.3"
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -49,7 +49,7 @@ impl Cache {
         // Otherwise, fall back to $XDG_CACHE_HOME/tealdeer.
         let xdg_dirs = match BaseDirectories::with_prefix("tealdeer") {
             Ok(dirs) => dirs,
-            Err(_) => return Err(CacheError("$HOME does not exist or is not defined.".into())),
+            Err(_) => return Err(CacheError("Could not determine XDG base directory.".into())),
         };
         Ok(xdg_dirs.get_cache_home())
     }
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -47,7 +47,7 @@ impl Cache {
         };
 
         // Otherwise, fall back to $XDG_CACHE_HOME/tealdeer.
-        let xdg_dirs = match BaseDirectories::with_prefix("tealdeer") {
+        let xdg_dirs = match BaseDirectories::with_prefix(::NAME) {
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -17,17 +17,18 @@ logging = ["env_logger"]
...
-docopt = "^0.6"
 time = "^0.1"
+flate2 = "^0.2"
+docopt = "^0.6"
 walkdir = "^0.1"
-env_logger = { version = "^0.3", optional = true }
+ansi_term = "^0.7"
+rustc-serialize = "^0.3"
 clippy = {version = "~0.0.37", optional = true}
+env_logger = { version = "^0.3", optional = true }
```
dbrgn: Haha, ordering by length. Is that some convention?

src/cache.rs
```diff
@@ -46,11 +46,12 @@ impl Cache {
             }
         };
 
-        // Otherwise, fall back to ~/.cache/tealdeer
-        let home_dir = try!(env::home_dir().ok_or(
-            CacheError("Could not determine home directory".into())
-        ));
-        Ok(home_dir.join(".cache").join("tealdeer"))
+        // Otherwise, fall back to $XDG_CACHE_HOME/tealdeer.
+        let xdg_dirs = match BaseDirectories::with_prefix("tealdeer") {
```
dbrgn: Maybe we could use `with_prefix(::NAME)``?

src/cache.rs
```diff
@@ -36,6 +35,7 @@ impl Cache {
         // $TLDR_RS_CACHE_DIR env variable.
```
dbrgn: Oh, we should probably rename that too (in a separate PR)...

SShrike: That was the next thing on my list, renaming everything from `tldr` to `tealdeer`.

src/cache.rs
```diff
@@ -46,11 +46,12 @@ impl Cache {
...
 
-        // Otherwise, fall back to ~/.cache/tealdeer
-        let home_dir = try!(env::home_dir().ok_or(
-            CacheError("Could not determine home directory".into())
-        ));
-        Ok(home_dir.join(".cache").join("tealdeer"))
+        // Otherwise, fall back to $XDG_CACHE_HOME/tealdeer.
+        let xdg_dirs = match BaseDirectories::with_prefix("tealdeer") {
+            Ok(dirs) => dirs,
+            Err(_) => return Err(CacheError("$HOME does not exist or is not defined.".into())),
```
dbrgn: The XDG cache dir could also be defined depending on the `XDG_CACHE_HOME` variable or on other fallbacks. Maybe make this more generic?

    return Err(CacheError("Could not determine XDG base directory.".into()))

src/cache.rs
dbrgn: I thought I did that, but missed some places...

Cargo.toml
SShrike: Didn't even realise I did that haha, it's personal preference really. I'm not really sure how most Rustaceans organise their `Cargo.toml`.

src/cache.rs
SShrike: I intend to rename things like `$TLDR_RS_CACHE_DIR`, `TldrError`, etc.

dbrgn: :+1:

src/cache.rs
SShrike: Good call.

SShrike:
Obligatory :tada:

PR #34, opened 2017-08-19
Add bash completion

silentboatman:
To use bash completion: Rename 'bash_tealdeer' to 'tldr' and move it
to /usr/share/bash-completion/completions/ .

### DIFF
--- a/bash_tealdeer
+++ b/bash_tealdeer
@@ -1,0 +1,30 @@
+# tealdeer bash completion
+
+_tealdeer()
+{
+	local cur prev words cword
+	_init_completion || return
+
+	case $prev in
+		-h|--help|-v|--version|-l|--list|-u|--update|-c|--clear-cache)
+			return
+			;;
+		-f|--render)
+			_filedir
+			return
+			;;
+		-o|--os)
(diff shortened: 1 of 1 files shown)


dbrgn:
This is nice! Some questions:

- The file contains some of the completions hardcoded in. Would there be a way to generate this through docopt?
- Are there any conventions on how these files are handled in projects? We probably can't distribute this file through `cargo install`, but maybe we can make the life for packaging as easy as possible.

silentboatman:
I haven't worked with docopt. But the only way that I can think of to get rid of the hard coded os types is to add a flag to tealdeer that lists them. Something like `tldr --list-os`.

Alot of programs add their bash completion file directly to the [bash-completion projet](https://github.com/scop/bash-completion). But that's maybe not the best solution in this case, since there are several tldr programs with (I suppose) different flags..

I haven't seen any standard place were people put their completion files.. but here's some different packages and were they put it in:

- pacman _/scripts/completion/_
- exa _/contrib/_
- systemd _/shell-completion/bash/_
- task in _/scripts/bash/_
- ninja _/misc/_
- colord _/data/_

dbrgn:
Sorry for the long delay, and thanks!

PR #35, opened 2017-11-26
cargo upgrade --all

anderejd:
Upgraded all dependencies using `cargo upgrade --all` (https://crates.io/crates/cargo-edit) and fixed the resulting build errors by moving to serde for deserialization.

`cargo test` fails both on master and this branch so I did not bother with that in this PR (the special command line test invoker does not find the executable (on macOS, latest rust stable, maybe use this: https://github.com/killercup/assert_cli ?)).


Cargo.toml
```diff
@@ -1,34 +1,42 @@
...
+walkdir = "2.0.1"
+xdg = "2.1.0"
+
+[dependencies.clippy]
+optional = true
+version = "0.0.174"
+
+[dependencies.env_logger]
+optional = true
+version = "0.4.3"
```
dbrgn: Did you use some kind of auto-formatting tool on this file? Can you change this back to the inline-style? I prefer that :)

src/types.rs
```diff
@@ -12,27 +10,6 @@ pub enum OsType {
...
-                "other" => Ok(OsType::Other),
-                _ => Err(d.error(&format!("Invalid OS type: '{}'. Choose one of 'linux', \
-                                           'osx', 'sunos' or 'other'.", lowercase)))
-            }
-        })
-    }
-}
-
-
-
```
dbrgn: Nice, makes things simpler :)

src/types.rs
```diff
@@ -77,30 +54,31 @@ impl LineType {
...
-        assert_eq!(json::decode::<OsType>("\"Linux\"").unwrap(), Linux);
-        assert_eq!(json::decode::<OsType>("\"LINUX\"").unwrap(), Linux);
-    }
+    // REVIEW: Not sure how to do case-insensitive deserialization with serde
+    //         json. Ok to ditch support for this?
+    //#[test]
+    //fn test_os_type_decoding_uppercase() {
+    //    assert_eq!(from_str::<OsType>("\"Linux\"").unwrap(), Linux);
+    //    assert_eq!(from_str::<OsType>("\"LINUX\"").unwrap(), Linux);
+    //}
```
dbrgn: I think we're looking for this, right? https://github.com/serde-rs/serde/issues/586

I'm not sure how important it is to have case insensitive matching, but the workaround by dtolnay seems quite low-effort to implement, what do you think?

Cargo.toml
anderejd: Ah yes, `cargo upgrade` does not preserve formatting and always outputs in this format as far as I know. I can massage this file into inline shape later today.

src/types.rs
anderejd: Looks simple enough, but would also be nice to avoid the extra code if it's not needed. I can add the workaround if you would like to keep the support for case-insensitive input?

anderejd:
I tried to implement the serde workaround for enums and failed, but found out docopt already provides case insensitive deserialization of enums :)
https://github.com/docopt/docopt.rs/blob/62cf32bd22f96941cec998adda5550ac475d5d5e/src/dopt.rs#L921

dbrgn:
Sorry, I forgot about this PR...

> I tried to implement the serde workaround for enums and failed, but found out docopt already provides case insensitive deserialization of enums :)

Very nice :) Thanks for the changes!

dbrgn:
> maybe use this: https://github.com/killercup/assert_cli

Good suggestion, will integrate that before merging this PR.

PR #41, opened 2018-02-11
Add Windows CI

dbrgn:
Does it compile on Windows?

Refs #39.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+os: Visual Studio 2017
+
+environment:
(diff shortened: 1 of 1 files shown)


dbrgn:
Hm, needs changes because it does not conform to the XDG specification.

messense:
FYI, Travis CI supports Windows now.

https://docs.travis-ci.com/user/reference/windows/

dbrgn:
@messense thanks for the hint, much nicer than appveyor :)

PR #43, opened 2018-07-08
Feature/improve colour contrast

Voultapher:
## What
- Change default colour scheme to high contrast using terminal default colour
- Highlight program name in code examples

Before:
screenshot_20180708_173111
After:
screenshot_20180708_173320

## Why
Right now the only reason I use a painfully slow python version of tldr is because its colour scheme makes quickly glancing at a tldr and seeing what I'm looking for much easier than when looking at the monotone low contrast produced by the current version of tealdeer.
The current contrast of most likely dark background to teal is challenging for certain people with colour blindness. I understand that the project is called **teal**dear and removing the teal colour might make the pun obsolete.
If you prefer the old colour scheme, I suggest making it an option. People that want another scheme could do `alias tldr=tldr --color=teal`.

## Implementation Notes
I chose to go with string indexing over `split` and `flat_map` as I wanted to avoid a performance regression. In a micro benchmark, `format_code` is faster and more consistent than `format_braces` whilst doing more work. Yet it is seems to be only a small fraction of the overall time, which means I wouldn't be opposed to rewriting it in a more *readable* fashion.
Pulling the command from the title isn't pretty, the command name is present when called regularly, however when using the option `--render` we would have to either introduce a new parameter to specify the command, which would increase usage burden and could introduce user errors, or accept different behavior for regular use vs `--render` which I find particular bad. This solution doesn't change the current interface and behaves the same way in regular use and `--render`. However I have not tested if all tldr files correctly specify the program title in their title.


Voultapher:
Also if this is accepted we should update the README example picture.

dbrgn:
Hi. Thanks for your pull request!

> I understand that the project is called tealdear and removing the teal colour might make the pun obsolete.

That was actually not intentional, I never noticed :slightly_smiling_face:

Regarding the proposed change, first of all, do you use some customizations regarding your terminal colors? In a regular dark-background xterm session, the output looks like this:

img

Quite high-contrast to me, I'm not sure why it looks so uniform on your terminal.

In contrast, your version looks like this:

img

To me that's less easy to parse (except that highlighting the program name is a great idea).

Voultapher:
> do you use some customizations regarding your terminal colors
Yes. I use konsole with a custom color scheme.

I think when it comes to color schemes, we can't make one size fits all. The best we can do is something that is good in *most* cases. To get a better understanding I tried out all standard konsole color schemes with the new highlighting.
screenshot_20180714_092235
screenshot_20180714_092312
screenshot_20180714_092341
screenshot_20180714_092425
screenshot_20180714_092453
screenshot_20180714_092519
screenshot_20180714_092556
screenshot_20180714_092612
screenshot_20180714_092627
screenshot_20180714_092653
screenshot_20180714_092717
screenshot_20180714_092729
screenshot_20180714_092747
screenshot_20180714_092802
screenshot_20180714_092815

Most of them are fine. The explicit white for the description creates problems in the caption for some of the bright schemes. My understanding is that we want a clear contrast between the example text and the example code. Coming back to not being able to make it perfect for everyone, I believe the default color is likely to always be clearly readable. If we stick to giving either the example text or example code another color, we greatly increase the risk of making the 2 styles incompatible. By highlighting the program name in example code, I believe there is a noticeable difference between example text and code. Red is often high contrast to the default color, and even when it is not only the program name is hard to read and the rest of the information which is likely more unknown is still legible.
I'd like the description to stand out, yet sticking to what I said about adding non default color increasing the potential for poor contrast, I suggest giving it the default color and somehow highlighting it differently.
2 examples
screenshot_20180714_095347
screenshot_20180714_100017

Voultapher:
@dbrgn this has been without activity for 2 weeks now. How should we proceed?

dbrgn:
@Voultapher sorry, I already typed a reply but never sent it off. I'm a bit tight on free time at the moment.

I think the best solution in this case would be theme support (in the sense of a config file where you can specify colors from an ANSI color palette), so that people with non-standard terminal colors can customize the output themselves. However, it should not impact performance in case the defaults are used.

Right now we don't have any configfile support. Maybe checking for a configfile (using [dirs](https://github.com/soc/dirs-rs) or [directories](https://github.com/soc/directories-rs)) when starting up would be an option. If no config file exists, just carry on. If a config file exists, parse it (I think I'd prefer toml format that's being parsed into structs) and apply a theme if it exists.

Voultapher:
Ok, will look into it.

dbrgn:
@Voultapher hehe, actually last week I switched to alacritty by default (instead of uxterm) which uses a slightly different theme with less contrast between green and yellow, for example, so I'm more sympathetic to your feature request now :smile:

Voultapher:
Ok, syntax config support and tests for it are done, concerning performance implication of this feature:

`brach name`:

time in seconds
instruction count

`master`:

0.010360291
19,279,851

0.012931068
17,382,742

0.011129954
12,447,344

0.009349510
17,045,595

0.009982335
18,797,206

`feature/improve-colour-contrast | no config file`:

0.011297944
17,443,583

0.011954686
18,474,816

0.011131505
18,740,495

0.010944560
16,565,537

0.010744000
17,389,713

`feature/improve-colour-contrast | with config file`:

0.010893010
18,202,644

0.009908770
15,949,160

0.010217479
16,634,313

0.010714000
18,750,531

0.011990885
14,365,538

This should not be a noticeable slowdown, no matter if a syntax config is used or not. Btw, serde is dope! Imo error categories, propagation and handling are not perfect, still I'm happy with the current state. Take a look and tell me if there is something to improve.

dbrgn:
Very nice to see these performance numbers :+1: Serde is blazing fast.

I'll try to have a look next week if possible, I'll keep this tab open. If I forget for some reason, feel free to remind me in the last week of August, when I have a few free days :slightly_smiling_face:

Voultapher:
Friendly ping.

dbrgn:
> Friendly ping.

Thanks! Yes, it's still on my todo list :slightly_smiling_face: The weather early this week was great, so I spent it paragliding instead of sitting in front of my computer :wink:

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl From<RawConfig> for Config {
+    fn from(raw_config: RawConfig) -> Config {
+        Config{
+            highlight_style: Style::from(raw_config.highlight_style),
+            description_style: Style::from(raw_config.description_style),
+            example_text_style: Style::from(raw_config.example_text_style),
+            example_code_style: Style::from(raw_config.example_code_style),
+            example_variable_style: Style::from(raw_config.example_variable_style),
```
dbrgn: Note: You can probably use `raw_style.into()` instead of `Style::from(raw_style)` here. Makes the code a bit shorter.

src/error.rs
```diff
@@ -11,3 +14,9 @@ impl From<CurlError> for TealdeerError {
         TealdeerError::UpdateError(format!("Curl error: {}", err.to_string()))
     }
 }
+
+impl From<IoError> for TealdeerError {
+    fn from(err: IoError) -> TealdeerError {
+        TealdeerError::ConfigError(format!("Io error: {}", err))
+    }
+}
```
dbrgn: Hm, I'm not sure if we should convert `IoError`s to `ConfigError`, since an `IoError` could also occur in cache handling code for example. Maybe it would be better to explicitly convert with `.map_err(|e| ConfigError(...))`?

src/formatter.rs
```diff
@@ -1,36 +1,75 @@
...
-            LineType::Title(_) => debug!("Ignoring title"),
-            LineType::Description(text) => println!("  {}", text),
-            LineType::ExampleText(text) => println!("  {}", Colour::Green.paint(text)),
-            LineType::ExampleCode(text) => println!("      {}", &format_braces(&text)),
+            LineType::Title(title) => {
+                debug!("Ignoring title");
+
+                // This is safe as long as the parsed title is only the command,
+                // and tokenizer yields values in order of appearance.
+                command = title;
```
dbrgn: Hm, I'm not too happy about that, but the other options aren't really better (e.g. extracting the name from the filename, which isn't guaranteed to be correct either).

Maybe add a `debug!("Detected command name: {}", command)` call?

src/formatter.rs
```diff
@@ -1,36 +1,75 @@
...
+            },
+            LineType::Description(text) => println!("  {}", format_description(&text, &config)),
+            LineType::ExampleText(text) => println!("  {}", config.example_text_style.paint(text)),
```
dbrgn: For consistency I'd add a `format_text` function here too, even if it's just a one-liner.

src/main.rs
```diff
@@ -110,9 +117,22 @@ fn print_page(path: &Path) -> Result<(), String> {
...
+    let config = match Config::new() {
+        Ok(config) => config,
+        Err(ConfigError(msg)) => {
+            println!("Failed to create config: {}", msg);
+            process::exit(1);
+        }
+        Err(_) => {
+            println!("Unknown error while creating config");
+            process::exit(1);
+        }
```
dbrgn: You could use [`eprintln!`](https://stackoverflow.com/questions/27588416/how-to-send-output-to-stderr#27590832) in those two cases. Same thing for the other errors below that result in `process::exit(1)`.

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+}
+
+impl Config {
+    pub fn new() -> Result<Config, TealdeerError> {
```
dbrgn: I would prefer not to call this method `new`, since I would not expect a call to `Config::new` to have the side effect of doing file system lookups.

What about renaming it to `load`?

README.md
```diff
@@ -61,13 +61,15 @@ These are the clients I tried but failed to compile or run:
...
-        -c --clear-cache    Clear the local cache
+        -h --help            Show this screen
+        -v --version         Show version information
+        -l --list            List all commands in the cache
+        -f --render <file>   Render a specific markdown file
+        -o --os <type>       Override the operating system [linux, osx, sunos]
+        -u --update          Update the local cache
+        -c --clear-cache     Clear the local cache
+        -d --display-config  Show config directory path
+        -s --seed-config     Create a basic config
```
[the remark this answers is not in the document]
Voultapher: I agree.

README.md
```diff
@@ -83,6 +85,37 @@ These are the clients I tried but failed to compile or run:
 
         $ tldr --render /path/to/file.md
 
+### Style Customization
+
+The tldr page syntax highlighting can be customized with a config file.
+Creating the config file can be done manually or with the help of tldr.
+```
+tldr -s
+```
```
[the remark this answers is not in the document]
Voultapher: Habit, will do.

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+    Red,
+    Green,
+    Yellow,
+    Blue,
+    Purple,
+    Cyan,
+    White,
+}
+
+impl From<RawColour> for Colour {
```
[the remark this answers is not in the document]
Voultapher: Looked at it when writing it, came to the same conclusion, making it explicit seems clearer.

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+use ansi_term::{Colour, Style};
+use toml;
+use xdg::BaseDirectories;
+
+use error::TealdeerError::{self, ConfigError};
+
+pub const SYNTAX_CONFIG_FILE_NAME: &'static str = "syntax.toml";
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RawColour {
```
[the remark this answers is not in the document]
Voultapher: AnsiTerm also exposes color, as alias to colour, how about chaning all mentions of colour to color in the project?

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+            RawColour::Yellow => Colour::Yellow,
+            RawColour::Blue => Colour::Blue,
+            RawColour::Purple => Colour::Purple,
+            RawColour::Cyan => Colour::Cyan,
+            RawColour::White => Colour::White,
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
Voultapher: True, usually I lean towards more derives than less, because third party users, even within a crate might have use cases that depend on some derives, however here there is no user of ord, plus this type does not get exposed. I'll remove them.

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub example_code_style: RawStyle,
+    pub example_variable_style: RawStyle,
+}
+
+impl RawConfig {
+    fn new() -> RawConfig {
+        let mut raw_config = RawConfig::default();
+
+        raw_config.highlight_style.foreground = Some(RawColour::Red);
+        raw_config.example_variable_style.underline = true;
```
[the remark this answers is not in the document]
Voultapher: Not sure, what do we gain from having the caller specify them?

src/error.rs
Voultapher: That's a lot of extra code, how about moving this impl to only be visible for the config module. Does that even work?

src/formatter.rs
Voultapher: Imo not a valuable abstraction, will do if you insist.

src/config.rs
Voultapher: Sounds good.

Voultapher:
> The first words of the description text is highlighted too. Furthermore, the commands are duplicated tartar instead of tar. I do like tartar, but this should be fixed stuck_out_tongue (Btw, you also copied that error into the test expectations.)

Wow, no idea how I missed that :stuck_out_tongue:

src/config.rs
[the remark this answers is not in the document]
dbrgn: When creating a new `RawConfig` instance, I would expect it to be uninitialized. Otherwise the responsibility of the `RawConfig` type (encapulating style configuration) and the calling program (configuring the style of the output) are mixed -- the container all of a sudden sets styling. I'd separate those clearly.

src/error.rs
dbrgn: I don't think that works. In the future, [catch](https://doc.rust-lang.org/unstable-book/language-features/catch-expr.html) will probably help in these cases.

If you want we can leave the impl in there for now. But I'm a bit uneasy about it, since when we add more I/O related code to cache handling or something like that, an error might wrongly be reported as a `ConfigError` when forgetting to properly convert the error.

src/formatter.rs
dbrgn: Ok, skip it if you don't think it's worth it :)

src/config.rs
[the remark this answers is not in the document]
Voultapher: That's what default does, if we move it to the caller site we'd repeat the same thing 3 times.

src/config.rs
```diff
@@ -0,0 +1,207 @@
...
+use std::io::{Read, Write};
+use std::path::PathBuf;
+
+use ansi_term::{Colour, Style};
+use toml;
+use xdg::BaseDirectories;
+
+use error::TealdeerError::{self, ConfigError};
+
+pub const SYNTAX_CONFIG_FILE_NAME: &'static str = "syntax.toml";
```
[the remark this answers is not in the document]
Voultapher: Personal preference is to have special purpose config files, because it isolates handling code. With a shared config file there is always the consideration that unrelated logic is influenced. Of course there is a limit, 100 config files are not useful, so far it looks like there will be 2 or so config files with #44 also doing configuration.

Voultapher:
From my point of view there are only 2 outstanding items, multipurpose config files and `RawConfig` default initialization, both waiting for your feedback. Is there something else unresolved?

src/config.rs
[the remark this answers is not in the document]
dbrgn: Ok, let's keep it for now, but then at least add a docstring indicating that some values will be initialized. Something like this:

```
/// Create a new `RawConfig` instance with some values initialized to certain default values:
///
/// - Foreground highlight style is "red"
/// - Example variable style is "underlined"
```

src/config.rs
```diff
@@ -113,22 +113,28 @@ pub struct Config {
...
+            example_text_style: raw_config.example_text_style.into(),
+            example_code_style: raw_config.example_code_style.into(),
+            example_variable_style: raw_config.example_variable_style.into(),
         }
     }
 }
 
+fn map_io_err_to_config_err(e: IoError) -> TealdeerError {
+    ConfigError(format!("Io Error: {}", e))
+}
```
dbrgn: Good approach, with the helper function! :+1:

src/formatter.rs
```diff
@@ -3,29 +3,42 @@
...
-                parts.push(config.example_code_style.paint(example_code));
-                parts.push(config.example_variable_style.paint(example_variable));
-            }
-            else {
-                parts.push(config.example_code_style.paint(between_variables));
-            }
+            highlight_command(&command, &example_code, &config, &mut parts);
+            parts.push(config.example_variable_style.paint(example_variable));
+        }
+        else {
```
dbrgn: Nitpick: I prefer the `else` to be on the same line as the closing bracket.

tests/inkscape-with-config.expected
```diff
@@ -0,0 +1,28 @@
+
+  [32mAn[0m SVG (Scalable Vector Graphics) editing program.
+  [32mUse[0m -z to not open the GUI and only process files in the console.
```
dbrgn: Hm, I still find it strange that `An` and `Use` are highlighted. Could we avoid that?

src/config.rs
```diff
@@ -0,0 +1,215 @@
...
+
+use ansi_term::{Color, Style};
+use toml;
+use xdg::BaseDirectories;
+
+use error::TealdeerError::{self, ConfigError};
+
+pub const SYNTAX_CONFIG_FILE_NAME: &'static str = "syntax.toml";
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
```
dbrgn: Can you add a `#[serde(rename_all = "lowercase")]` annotation? Then we can put lowercase color names in the configuration file. Capitalized colors seems strange, since everything else is lowercase.

src/config.rs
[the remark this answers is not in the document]
dbrgn: I'd go the other route, if that's OK for you. Start with a single config file, that's much easier to handle for the users, and also for the handling inside the code (you get a single configuration object, and have to deal with a single path, instead of having to handle multiple paths and have to do file-exists checks for every file etc).

If the config gets too big, it can still be split. But I don't see how the tealdeer config file should ever grow overly complex.

Inside the config, we could put all color configuration inside a `style` section for grouping. I guess it would look like this:

```toml
[style]

    [style.highlight]
    foreground = "blue"
    underline = false
    bold = true

    [style.description]
    underline = false
    bold = false

    [style.example_text]
    underline = false
    bold = false

    [style.example_code]
    underline = false
    bold = false

    [style.example_variable]
    underline = true
    bold = false
```

Other sections could then be added for non-style configuration (like #44).

Actually when I look at the example, I have to say that the nested syntax looks kind of ugly... I took a look at the [alacritty](https://github.com/jwilm/alacritty) config, which uses YAML and it looks much nicer, roughly like this:

```yaml
# Configuration for tealdeer

# Configure styling of the different parts of the output.
style:

  # Style used for highlighting the command name
  highlight:
    foreground: blue
    underline: false
    bold: true

  # Style used for the description text
  description:
    underline: false
    bold: false

  # Style used for the example text
  example_text:
    underline: false
    bold: false

  # Style used for the example code
  example_code:
    underline: false
    bold: false

  # Style used for variables inside the example code
  example_variable:
    underline: true
    bold: false
```

What do you think? Thanks to serde, switching to YAML would probably be easy.

tests/inkscape-with-config.expected
Voultapher: The idea was to visually distinguish the description from example text, to allow it to be found with a quick glance due to its high contrast unique pattern. I'm open to alternatives, not separating it from example text would seems like a missed opportunity.

src/config.rs
Voultapher: Good idea, done.

tests/inkscape-with-config.expected
dbrgn: Well, it's the first thing after the headline. And it's not indented. Adding a configurable style for it would be OK, but highlighting the first word seems confusing to me :)

dbrgn:
Thanks for the adjustments!

> Also should we warn the user if `tldr --seed-config` is being run and a file is already present in that location?

That's a good idea. I think even showing an error message à la "A configuration file already exists at <PATH>, no action was taken" with a non-0 exit code would be fine.

> Everything else looks good.

What do you think about the YAML proposal above? Shouldn't be too hard to try thanks to libraries like [serde-yaml](https://github.com/dtolnay/serde-yaml), and I think it would improve readability. (I'd be fine wtih TOML too though.)

Voultapher:
> What do you think about the YAML proposal above? Shouldn't be too hard to try thanks to libraries like serde-yaml, and I think it would improve readability. (I'd be fine wtih TOML too though.)

I'll repost what I wrote earlier about toml vs serde, it seems you didn't see it:

Personally prefer toml over yaml for confgs like this, also if we make the style config a member of the config object, there is no need to introduce a block. Looks like this:

```toml
[style.highlight]
foreground = "blue"
underline = false
bold = true

[style.description]
underline = false
bold = false

[style.example_text]
underline = false
bold = false

[style.example_code]
underline = false
bold = false

[style.example_variable]
underline = true
bold = false
```

To add to it. The yaml version would look something like this:
```yaml
style:

  highlight:
    foreground: blue
    underline: false
    bold: true

  description:
    underline: false
    bold: false

  example_text:
    underline: false
    bold: false

  example_code:
    underline: false
    bold: false

  example_variable:
    underline: true
    bold: false
```
Not sure that is better.

tests/inkscape-with-config.expected
Voultapher: It already has a configurable style, however it is indented on the same level as the example text, still I agree that the split word highlighting may be confusing and could collide with a custom style. So I'll take it out.

Voultapher:
Friendly ping.

dbrgn:
OK let's keep TOML for now :)

I'll do some more adjustments (like making the default style like the previous default) and will then merge the branch!

Thanks a lot @Voultapher for your patience :)

dbrgn:
Merged in 278a2bbac9ea8434f546df017fc935e1d4013200!

Issue #36, opened 2017-11-29
support pagination


tamird:
Currently output will happily flow when the terminal is smaller than the output.

dbrgn:
Isn't that what tools like `less` and `more` are for?

tamird:
Yes, and tools like git that do this properly automatically detect when a
pager is needed; they don't make the user do it.

dbrgn:
True.

It's not a priority for me personally, but I'd accept pull requests :)

PR #44, opened 2018-08-21
#36 add -p --pager flag to page the output using pager crate

jdvr:
I just added the [pager crate](https://crates.io/crates/pager) and a flag (`-p --pager`) to enable or disable it.

BTW, I have tested it with Rust 1.28.0 and it works well, maybe you can update.

I would need help for windows and macOS testers.

refs #36

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -109,8 +109,15 @@ struct RawStyleConfig {
 }
 
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+struct RawDisplayConfig {
+    #[serde(default)]
+    pub use_pager: bool,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 struct RawConfig {
     style: RawStyleConfig,
+    display: RawDisplayConfig,
 }
 
 impl RawConfig {
@@ -123,6 +130,7 @@ impl RawConfig {
         raw_config.style.example_code.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.underline = true;
+        raw_config.display.use_pager = false;
 
         raw_config
     }
@@ -138,8 +146,14 @@ pub struct StyleConfig {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
+pub struct DisplayConfig {
+    pub use_pager: bool,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub struct Config {
     pub style: StyleConfig,
+    pub display: DisplayConfig,
 }
 
 impl From<RawConfig> for Config {
@@ -151,6 +165,9 @@ impl From<RawConfig> for Config {
                 example_text: raw_config.style.example_text.into(),
                 example_code: raw_config.style.example_code.into(),
                 example_variable: raw_config.style.example_variable.into(),
+            },
+            display: DisplayConfig {
+                use_pager: false,
             },
         }
     }
@@ -193,6 +210,9 @@ impl Config {
                     example_text: Style::default(),
                     example_code: Style::default(),
                     example_variable: Style::default(),
+                },
+                display: DisplayConfig {
+                    use_pager: false,
                 }
             }
         })
--- a/src/config.rs
+++ b/src/config.rs
@@ -116,7 +116,9 @@ struct RawDisplayConfig {
 
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 struct RawConfig {
+    #[serde(default)]
     style: RawStyleConfig,
+    #[serde(default)]
     display: RawDisplayConfig,
 }
 
@@ -134,7 +136,7 @@ impl RawConfig {
 
         raw_config
     }
-} // impl RawConfig
+}
 
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct StyleConfig {
--- a/src/config.rs
+++ b/src/config.rs
@@ -188,7 +188,7 @@ impl Config {
         let config_file_path = get_config_path()
             .map_err(|e| ConfigError(format!("Could not determine config path: {}", e)))?;
 
-        // Load config
+        // Load raw config
         let raw_config: RawConfig = if config_file_path.exists() && config_file_path.is_file() {
             let mut config_file =
                 fs::File::open(config_file_path).map_err(map_io_err_to_config_err)?;
@@ -202,24 +202,23 @@ impl Config {
             RawConfig::new()
         };
 
-        Ok(if enable_styles {
-            Self::from(raw_config)
-        } else {
-            Self {
-                style: StyleConfig {
-                    command_name: Style::default(),
-                    description: Style::default(),
-                    example_text: Style::default(),
-                    example_code: Style::default(),
-                    example_variable: Style::default(),
-                },
-                display: DisplayConfig {
-                    use_pager: false,
-                }
-            }
-        })
+        // Convert to config
+        let mut config = Self::from(raw_config);
+
+        // Potentially override styles
+        if !enable_styles {
+            config.style = StyleConfig {
+                command_name: Style::default(),
+                description: Style::default(),
+                example_text: Style::default(),
+                example_code: Style::default(),
+                example_variable: Style::default(),
+            };
+        }
+
+        Ok(config)
     }
-} // impl Config
+}
 
 /// Return the path to the config directory.
 ///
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -162,6 +168,10 @@ fn main() {
     let args: Args = Docopt::new(USAGE)
                             .and_then(|d| d.deserialize())
                             .unwrap_or_else(|e| e.exit());
+    
+    if args.flag_pager {
+        Pager::new().setup();
```
dbrgn: Maybe we should add the `.skip_on_notty()` option?

src/main.rs
```diff
@@ -72,6 +76,7 @@ Options:
     -o --os <type>      Override the operating system [linux, osx, sunos]
     -u --update         Update the local cache
     -c --clear-cache    Clear the local cache
+    -p --pager    User `more` command to page output
```
dbrgn: Can you align the help text with the other help text lines (using spaces)?

Also, I wonder if the pager behavior should be the default... Maybe a `--no-pager` option would be better? (CC @tamird)

src/main.rs
jdvr: I thought the same, but according to the rust compiler that options is enable by default.

```
warning: use of deprecated item 'pager::Pager::skip_on_notty': 'skip_on_notty' is default now
   --> src/main.rs:173:22
    |
173 |         Pager::new().skip_on_notty().setup();
    |                      ^^^^^^^^^^^^^
    |
    = note: #[warn(deprecated)] on by default

```

src/main.rs
jdvr: My fist implementation what following that but the thing is, since we are delegating the output to `more` the usability of it is weird because after print a command page, you have to hit  `Q` to close the pager.

image

You are stuck there until you press `q`

dbrgn: Oh, shouldn't the pager only activate when the output is too large for the screen? That's what git does by default I think.

On the other hand the pager crate can't know that, so maybe we need to implement that ourselves...

jdvr: Maybe I didn't explain myself well. Pager only pages when the output is large, but when it actives, you have to hit `q` to get out of there.

jdvr:
@dbrgn is there any pending issue about this ?

src/main.rs
dbrgn: You're right, then it's better to remove it again :)

dbrgn:
> @dbrgn is there any pending issue about this ?

No, I just needed to find some free time to review :slightly_smiling_face: 

My main issue with this is currently that the default pager is `more`, which does not support arrow up/down keys. I think most tools default to `less` which is more user friendly in that regard.

In my opinion the `pager` crate should first probe for `less`, and if found, invoke it with the `-r` or `-R` option for color support. Only then it should fall back to `more`. But maybe `more` is part of POSIX and `less` isn't?

We could force `less` to be used by using `Pager::with_pager("less -r").setup()`, but then we lose the ability to set another pager with the `PAGER` env variable, as well as support for the `NOPAGER` env var... A pull request to `pager` with a way to override the default (but still allowing env vars) would probably be nice.

Furthermore, it doesn't seem to be supported on Windows:

https://gitlab.com/imp/pager-rs/issues/5

So we should probably exclude the pager functionality when building for Windows.

I'll test the PR on macOS next week.

Regarding defaults, I'm not sure yet whether paging should be the default behavior or not. Probably not, but we can add it to the configuration file once #43 has landed.

jdvr:
I will keep the -p and them we can use the configuration file and implement the use of PAGER env var on our own.

I mean, we can follow a priority to check if the output has to be paged:
1. -p (page always)
2. Enabled in configuration file

Once we know we have to page, check if any pager is configure (PAGER env var or config file), otherwise use `less -r`.

dbrgn:
I tested it on macOS. Seems to work generally, but without colors:

screen shot 2018-09-04 at 09 35 29

Reason is that `more` is an alias for `less`, which requires the `-r` option to render escape codes.

screen shot 2018-09-04 at 09 35 55

All the more reason to use `less -r` by default :)

jdvr:
yep, totally agree, let's change by `less -r` because it also will work well in most Linux distribution

dbrgn:
I opened an issue: https://gitlab.com/imp/pager-rs/issues/10

dbrgn:
Released in pager 0.15! This should do:

    Pager::with_default_pager("less -R").setup();

Also, in the meantime configfile support was merged :slightly_smiling_face:

jdvr:
Thank you @dbrgn, I will update my PR today. I will also include an option for the configuration file as we commented previously on this thread.

jdvr:
Totally forgot about this 🙏. I will try to finish the task.

jdvr:
@dbrgm I have already tested this on Linux with several commands. I think is ready to be merge

src/config.rs
```diff
@@ -59,6 +64,13 @@ struct RawStyle {
     pub bold: bool,
 }
 
+
+#[derive(Copy, Clone, Default, Debug, Serialize, Deserialize, PartialEq, Eq)]
+struct RawDisplay {
+    #[serde(default = "default_pager")]
+    pub underline: bool,
```
dbrgn: That should probably not be called `underline` :slightly_smiling_face: Also, is the struct used anywhere?

src/config.rs
```diff
@@ -108,9 +120,16 @@ struct RawStyleConfig {
     pub example_variable: RawStyle,
 }
 
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+struct RawDisplayConfig {
+    #[serde(default)]
+    pub pager: bool,
```
dbrgn: I would rename this to `use_pager`.

```suggestion
    pub use_pager: bool,
```

src/config.rs
```diff
@@ -137,9 +157,15 @@ pub struct StyleConfig {
     pub example_variable: Style,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct DisplayConfig {
+    pub pager: bool,
```
dbrgn:
```suggestion
    pub use_pager: bool,
```

src/main.rs
```diff
@@ -56,6 +58,7 @@ Options:
     -o --os <type>      Override the operating system [linux, osx, sunos]
     -u --update         Update the local cache
     -c --clear-cache    Clear the local cache
+    -p --pager          User `less -r` command to page output
```
dbrgn:
```suggestion
    -p --pager          Use a pager to page output
```

src/main.rs
```diff
@@ -75,11 +78,13 @@ To render a local file (for testing):
     $ tldr --render /path/to/file.md
 ";
 const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/master.tar.gz";
+const PAGER_COMMAND: &str = "less -r";
```
dbrgn: I'd switch to `-R` by default, so that only color escape codes are interpreted by default. It's safer that way.

```suggestion
const PAGER_COMMAND: &str = "less -R";
```

src/main.rs
```diff
@@ -118,6 +123,30 @@ fn print_page(path: &Path) -> Result<(), String> {
...
+                eprintln!("Could not load config: {}", msg);
+                process::exit(1);
+            }
+            Err(e) => {
+                eprintln!("Could not load config: {}", e);
+                process::exit(1);
+            }
+        };
+
+        if (config.display.pager) {
```
dbrgn: No need for parentheses:

```suggestion
        if config.display.pager {
```

tests/config.toml
```diff
@@ -15,3 +15,6 @@ underline = false
 [style.example_variable]
 underline = true
 bold = false
+
+[display]
+pager = false
```
dbrgn:
```suggestion
use_pager = false
```

jdvr:
Tanks for the feedback @dbrgn 

I have updated the code.

dbrgn:
Thanks @jdvr!

For some reason the PR deletes the `Cargo.lock` file, but I can fix that in a manual merge.

dbrgn:
Huh, what's wrong with Travis... The tests passed but the state wasn't updated.

jdvr:
I am happy to see this merged after a lot of time 😁

mystal:
Oh no, this broke Windows support :( pager-rs doesn't support Windows (yet): https://gitlab.com/imp/pager-rs/issues/5

dbrgn:
Oh, too bad :face_with_head_bandage: We need Windows CI, otherwise this might happen again in the future.

Maybe we could simply disable pager support on Windows?

mystal:
Yeah, that's a good idea. I'll try to take a look next weekend when I should hopefully have some time.

Aside from that, pager support on Unix sounds great!

PR #47, opened 2018-09-16
Bump curl version

Voultapher:
Fix building on systems with openssl 1.1.1.
See https://github.com/sfackler/rust-openssl/issues/994 for more info.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: rust:1.19
+      - image: rust:1.28
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


Voultapher:
It looks like ci fails because of older rust version. This project is intended as binary and not library, I don't see a reason to support older toolchains. It build and passes tests with a recent rust version locally. I suggest updating the ci.

dbrgn:
Yeah, I think it makes sense to start committing to reverse compatibility once the 2018 edition is here.

dbrgn:
Huh, for some reason when I tried to rebase this PR, the commit disappeared...

Anyways, I'll merge it manually.

Issue #46, opened 2018-09-14
supress 'page not found in cache' message?


jedahan:
Maybe with a `-q, --quiet` flag?

https://github.com/dbrgn/tealdeer/blob/master/src/main.rs#L246

dbrgn:
Yeah, why not? A pull request would be welcome.

Maybe the `-q` flag could also silence the "cache old" warning: https://github.com/dbrgn/tealdeer/blob/master/src/main.rs#L144-L150

jedahan:
mind assigning me and I'll take a crack at it this week?

PR #48, opened 2018-09-17
add -q, --quiet flag to surpress non-error output

jedahan:
First pass at #46, tests pass and manual testing looked good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Possible log types:
 
 ### Unreleased
 
-- [added] New --quiet / -q option to surpress most non-error messages
+- [added] New --quiet / -q option to suppress most non-error messages
 
 ### v1.0.0 (2018-02-11)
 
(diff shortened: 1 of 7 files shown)


dbrgn:
Thanks, that was fast :) PR looks good.

The README is missing an update with the new command. Furthermore, I currently prefer if the version bump is only done on release. But I can fix these two issues myself.

dbrgn:
One more thing: Could you add a few small tests for this? You can take a look at `tests/lib.rs` for reference, for example `test_missing_cache`. Thanks to [assert_cli](https://docs.rs/assert_cli/0.5.6/assert_cli/) it's very simple to call the binary and to verify the output.

jedahan:
K, added tests. A bit hacky but seem to work.

dbrgn:
Nice, thanks @jedahan!

dbrgn:
Merged in 8784b55.

PR #49, opened 2018-10-09
Style fix

equal-l2:
- Applied rustfmt
- Comply with clippy


dbrgn:
Thanks! I don't agree with all those automatially applied changes (for example I think it makes `tests/lib.rs` worse), but I'll apply most of them.

dbrgn:
Merged manually in efde7430d15b5d222db5e48fdb708f25e941d9af.

PR #52, opened 2018-10-23
CI: Add Windows to TravisCI config

dbrgn:
Replaces #41, refs #39.


dbrgn:
Codebase needs changes because Windows does not conform to the XDG specification.

dbrgn:
I'll try GitHub Actions instead.

PR #55, opened 2018-10-23
add Nix install instructions

das-g:
(tested on NixOS. Not yet tested on other Nix-enabled Linux or on Nix-enabled macOS, but should work there, too, I hope.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,18 @@ To enable the log output, set the `RUST_LOG` env variable:
 
     $ export RUST_LOG=tldr=debug
 
+### Through Nix (NixOS, Linux with Nix or macOS with Nix)
+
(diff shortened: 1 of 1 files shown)


dbrgn:
Thanks! :tada: Can you move the section below macOS? I would estimate that more people use Arch or macOS than a Nix based system :slightly_smiling_face:

das-g:
> Can you move the section below macOS?

Done
> I would estimate that more people use Arch or macOS than a Nix based system :slightly_smiling_face:

True, but certainly more people use an OS where Nix is available (=could be installed, if wanted) than one where homebrew is available (=could be installed, if wanted). :wink:

(OK, alright, not quite true if we count [Linuxbrew](https://linuxbrew.sh) as a homebrew variant.)

dbrgn:
Thanks :slightly_smiling_face:

PR #61, opened 2018-11-18
move from curl to reqwest

### LINKED ISSUE #60: Support platform native TLS implementation
Currently this depends on curl for fetching the pages, and the curl crate depends on openssl-sys. Switching to a more portable high level http library like https://crates.io/crates/reqwest should help with that, and is probably also a requirement for #53.

jcgruenhage:
Fixes #60

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 #[cfg(unix)]
 use std::os::unix::fs::MetadataExt;
 
-use curl::easy::Easy as CurlEasy;
+use reqwest;
 use flate2::read::GzDecoder;
 use tar::Archive;
 use time;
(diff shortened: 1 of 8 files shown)


jcgruenhage:
Apparently this breaks rust 1.28 support. How important is that?

jcgruenhage:
Taking 1.31 once that is released sounds very good to me. `reqwest` isn't that old yet, yes, but really nice to use.

src/cache.rs
```diff
@@ -61,19 +61,9 @@ impl Cache {
...
-            let mut transfer = handle.transfer();
-            try!(transfer.write_function(|data| {
-                buf.extend_from_slice(data);
-                Ok(data.len())
-            }));
-            try!(transfer.perform());
-        }
+        let mut resp = reqwest::get(&self.url)?;
+        let mut buf: Vec<u8> = vec![];
+        resp.copy_to(&mut buf)?;
```
dbrgn: The `unused_results` linter complains here. To avoid, use `let _bytes = resp.copy_to(&mut buf)?;`.

jcgruenhage: What do you think about logging it instead? It's not quite the same, but it is getting rid of the linter warning and shouldn't hurt anyway.

jcgruenhage:
I don't understand the CI failure here.. The log line shouldn't change anything about the behavior when updating the cache, and the previous commit passed those tests.

src/cache.rs
dbrgn: Ah, yeah, why not. Something like "Downloaded {} bytes".

dbrgn:
@jcgruenhage since you run the command inside the `debug!()` macro, it is not executed when debug logs are not enabled :smiley: 

Solution: Assign the number of bytes to a variable and log that.

src/cache.rs
jcgruenhage: I worded it slightly differently, "{} bytes downloaded", but the result is the same ^^

jcgruenhage:
Oh, okay. Stupid me.

jcgruenhage:
Interesting, github now shows force pushes. Neat!

dbrgn:
> Interesting, github now shows force pushes. Neat!

Yeah, I noticed that too yesterday. Very nice.

Looks good now, thanks!

dbrgn:
Merged in d9507007469c9be6c811ce820982c714bc2d2f26

PR #63, opened 2018-11-20
Mention Void Linux Package in the README

jcgruenhage:
My PR for packaging tealdeer on void was merged, so you can now easily install tealdeer on void: https://github.com/void-linux/void-packages/pull/4908


dbrgn:
Thanks!

PR #68, opened 2018-11-22
Add (back) support for proxies

aldanor:
@dbrgn @jcgruenhage ^


aldanor:
Actually, hold on; do we want to bail if creating `Proxy()` from env fails? I would be inclined not to.

dbrgn:
>  I would be inclined not to.

Yes, I agree :slightly_smiling_face: 

Thank you!

PR #72, opened 2018-12-21
Make network support optional

### LINKED ISSUE #64: Allow --update from a local tarball; allow disabling update warnings
Two simple changes would make tealdeer more offline-use friendly. In particular, when packaging it for use in an offline environment, one could include tldr's tarball (master.tar.gz) in the package itself, and make it run `tldr --update <tarball>` when installing.

Related: it would be nice to be able to disable "Cache wasn't updated for more than 30 days" warning (maybe at build time) -- if you're in an offline environment, you can't really update it so it's just an annoyance...

dbrgn:
Work in progress.

Fixes #64.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -1,15 +1,19 @@
+#[cfg(feature = "networking")]
+use std::borrow::Cow;
 use std::env;
 use std::fs;
 use std::io::Read;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
-#[cfg(unix)]
+#[cfg(all(unix, feature = "networking"))]
 use std::os::unix::fs::MetadataExt;
 
+#[cfg(feature = "networking")]
 use reqwest::{Client, Proxy};
 use flate2::read::GzDecoder;
 use log::debug;
 use tar::Archive;
+#[cfg(feature = "networking")]
 use time;
 use walkdir::{DirEntry, WalkDir};
 use xdg::BaseDirectories;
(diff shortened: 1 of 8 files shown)


dbrgn:
Currently blocked on https://github.com/assert-rs/assert_cmd/issues/64

src/cache.rs
```diff
@@ -74,22 +94,38 @@ impl Cache {
...
         let mut buf: Vec<u8> = vec![];
         let bytes_downloaded = resp.copy_to(&mut buf)?;
         debug!("{} bytes downloaded", bytes_downloaded);
         Ok(buf)
     }
 
+    /// Download the archive from the network or load it from a file.
+    #[cfg(feature = "networking")]
+    fn load(&self) -> Result<Vec<u8>, TealdeerError> {
+        if self.source.starts_with("http") {
```
aldanor: Nitpicking, maybe something less stringly-typed? 😄 

```rust
pub enum Source {
    File(PathBuf),
    #[cfg(feature = "networking")]
    URL(String),
}
```

dbrgn: Hm, good point. Then we can move the parsing up in the hierarchy and convert the input string right after argument parsing :slightly_smiling_face:

dbrgn:
I think this should be done (except the edge cases commented above). Thorough testing would be welcome :slightly_smiling_face:

dbrgn:
@aldanor Would be great if you could quickly glance over the diff and test the branch!

dbrgn:
Closing this, since added complexity without a real need is not worth it.

dbrgn:
@yorkerhazan the repository should be cloned into `~/.cache/tealdeer/tldr-pages/`. If everything is set up correctly, then the file at `~/.cache/tealdeer/tldr-pages/index.json` and the directory at `~/.cache/tealdeer/tldr-pages/pages/` should exist.

PR #73, opened 2019-01-07
Add --auto-update flag to allow automatically updating the cache

rnestler:
When not updating the cache manually one often gets the following
message:

    Cache wasn't updated in 30 days.

The auto-update flag allows to just alias tldr to tldr -a which will
automatically update the cache if it is older than the maximum age.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -126,13 +126,19 @@ fn check_cache(args: &Args, cache: &Cache) {
         match cache.last_update() {
             Some(ago) if ago > MAX_CACHE_AGE => {
                 if args.flag_auto_update {
-                    cache.update().unwrap_or_else(|e| {
-                        match e {
-                            CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
-                                eprintln!("Could not update cache: {}", msg)
+                    cache.update()
+                        .map(|()| {
+                            if !args.flag_quiet {
+                                println!("Successfully updated cache.");
                             }
-                        };
-                    });
+                        })
+                        .unwrap_or_else(|e| {
+                            match e {
+                                CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
+                                    eprintln!("Could not update cache: {}", msg)
+                                }
+                            };
+                        });
                     return;
                 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,6 +55,7 @@ Options:
     -f --render <file>  Render a specific markdown file
     -o --os <type>      Override the operating system [linux, osx, sunos]
     -u --update         Update the local cache
+    -a --auto-update    Update the local cache automatically if out of date
     -c --clear-cache    Clear the local cache
     -q --quiet          Suppress informational messages
     --config-path       Show config file path
@@ -86,6 +87,7 @@ struct Args {
     flag_render: Option<String>,
     flag_os: Option<OsType>,
     flag_update: bool,
+    flag_auto_update: bool,
     flag_clear_cache: bool,
     flag_quiet: bool,
     flag_config_path: bool,
@@ -123,6 +125,17 @@ fn check_cache(args: &Args, cache: &Cache) {
     if !args.flag_update {
         match cache.last_update() {
             Some(ago) if ago > MAX_CACHE_AGE => {
+                if args.flag_auto_update {
+                    cache.update().unwrap_or_else(|e| {
+                        match e {
+                            CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
+                                eprintln!("Could not update cache: {}", msg)
+                            }
+                        };
+                    });
+                    return;
+                }
+
                 if args.flag_quiet {
                     return;
                 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -127,18 +127,14 @@ fn check_cache(args: &Args, cache: &Cache) {
             Some(ago) if ago > MAX_CACHE_AGE => {
                 if args.flag_auto_update {
                     match cache.update() {
-                        Ok(()) => {
-                            if !args.flag_quiet {
-                                println!("Successfully updated cache.");
+                        Ok(()) => if !args.flag_quiet {
+                            println!("Successfully updated cache.");
+                        },
+                        Err(err) => match err {
+                            CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -55,6 +55,7 @@ Options:
     -f --render <file>  Render a specific markdown file
     -o --os <type>      Override the operating system [linux, osx, sunos]
     -u --update         Update the local cache
+    -a --auto-update    Update the local cache automatically if out of date
```
dbrgn: Hmm, "if out of date" might indicate (wrongly) that the cache is updated if the upstream changes.

Maybe a wording like this would be better?

```suggestion
    -a --auto-update    Update the local cache automatically if older than 30 days
```

The `30 days` part might be wrong if we change `MAX_CACHE_AGE` but I'd prefer to keep the usage string static.

src/main.rs
```diff
@@ -123,6 +125,17 @@ fn check_cache(args: &Args, cache: &Cache) {
...
+                if args.flag_auto_update {
+                    cache.update().unwrap_or_else(|e| {
+                        match e {
+                            CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
+                                eprintln!("Could not update cache: {}", msg)
+                            }
+                        };
+                    });
+                    return;
+                }
```
dbrgn: I think it would be good to print an information if the cache was successfully updated:

```rust
if !args.flag_quiet {
    println!("Successfully updated cache.");
}
```

src/main.rs
rnestler: > The 30 days part might be wrong if we change MAX_CACHE_AGE but I'd prefer to keep the usage string static.

I choose my wording exactly because of that :stuck_out_tongue_closed_eyes: but yes, for the user this message is probably less confusing.

src/main.rs
rnestler: Yes, but since I don't abort with the '-a' flag when updating fails I'll need to do it with a `.map` call.

dbrgn: Yes, or match :)

rnestler: I did it with map, didn't see your comment.

dbrgn: To be honest `match` would be much more readable in this case. Don't you think?

rnestler: Well it looks almost exactly the same, except that it uses on more line:

```diff
diff --git a/src/main.rs b/src/main.rs
index 616829a..d21da36 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -126,19 +126,20 @@ fn check_cache(args: &Args, cache: &Cache) {
         match cache.last_update() {
             Some(ago) if ago > MAX_CACHE_AGE => {
                 if args.flag_auto_update {
-                    cache.update()
-                        .map(|()| {
+                    match cache.update() {
+                        Ok(()) => {
                             if !args.flag_quiet {
                                 println!("Successfully updated cache.");
                             }
-                        })
-                        .unwrap_or_else(|e| {
-                            match e {
+                        }
+                        Err(err) => {
+                            match err {
                                 CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
                                     eprintln!("Could not update cache: {}", msg)
                                 }
                             };
-                        });
+                        }
+                    }
                     return;
                 }
```

dbrgn: What about this?

```rust
match cache.update() {
    Ok(_) if !args.flag_quiet => println!("Successfully updated cache."),
    Ok(_) => {},
    Err(e) => match e {
        CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
            eprintln!("Could not update cache: {}", msg)
        }
    },
}
```

(The error match pattern could also be collapsed, but then it's a bit repetitive. There's an RFC to allow things like `Err(CacheError(msg) | ConfigError(msg) | UpdateError(msg))` in the future, but it hasn't landed yet AFAIK.)

dbrgn: Well, you argued with length, and this version is shorter. And I *love* my bikeshed in darker hues of blue! Light blue is just abhorrent...

dbrgn: And if you prefer it, you can also use this:

```rust
match cache.update() {
    Ok(_) => if !args.flag_quiet => {
        println!("Successfully updated cache.");
    },
    Err(e) => match e {
        CacheError(msg) | ConfigError(msg) | UpdateError(msg) => {
            eprintln!("Could not update cache: {}", msg)
        }
    },
}
```

I think that's actually the nicest version :slightly_smiling_face:

rnestler: @dbrgn Ping. I applied your favorite shade of medium blue :wink:

dbrgn: Sorry, I've been sick for the last three days. Will take a look in a few days.

dbrgn:
Thanks for the update.

I thought about this PR a bit. I'm still not sure if the command line flag is the right place for such a setting. It's not something you would want for some command invocations but not for others, so the config file would probably make more sense, right?

This would require a slight reorganization of the `main` function, but much less than I initially thought:

- Handle `args.flag_config_path` and `args.flag_seed_config` right after handling `args.flag_version` (simply move up the code blocks)
- Move the config loading code in `print_page` into the main function, right after the `args.flag_seed_config` handling block
- In `src/config.rs`, add the new config elements to the `RawConfig` and `Config` structs and add the defaults to `RawConfig::new()`.

The default configuration can be generated with `tldr --seed-config`. My suggestion would be:

```
[updates]
auto_update = false
```

(The `auto_update_interval_hours` could then be set to `MAX_CACHE_AGE * 24` by default.)

What do you think? Sorry for the change request, but every new flag is a maintenance burden because removing it later is a breaking change, so I want to think twice before adding such new flags...

dbrgn:
> But for the auto update the config file probably makes most sense since it is a rather static option.

:+1:

devnoname120:
Any updates on this?

dbrgn:
> Any updates on this?

The changes outlined in https://github.com/dbrgn/tealdeer/pull/73#issuecomment-453783054 would need to be implemented.

@rnestler are you still interested in implementing those changes, or should we close the PR for now so somebody else can take over?

rnestler:
> @rnestler are you still interested in implementing those changes, or should we close the PR for now so somebody else can take over?

This PR is probably the wrong place to implement this anyways. So let's close it and me or somebody else may open a new one.

dbrgn:
Opened an issue at #90.

PR #86, opened 2019-08-26
Add zsh autocompletion

ma-renaud:
It's my first attempt at a zsh completion file. Let me know if some changes are needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ set the `PAGER` environment variable.
 
 - *Bash*: copy `bash_tealdeer` to `/usr/share/bash-completion/completions/tldr`
 - *Fish*: copy `fish_tealdeer` to `~/.config/fish/completions/tldr.fish`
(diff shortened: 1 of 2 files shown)


dbrgn:
I'm not a zsh user myself  Can you find someone that can test this PR?

Also, I'm on my mobile right now and can't check, but I think there is a lot of duplication with the bash completion, right? Maybe the configs could be shared somehow? Potentially tealdeer could even generate the completion file(s) on demand.

dbrgn:
> Concerning the duplication, the structure of the file is similar, but the mechanics are not the same between bash and zsh.

Ok, then let's keep it for now. In the long run I'd like to generate those completion files.

For docopt (which we use right now) there doesn't seem to be a nice solution, but I'd like to switch to [clap](https://docs.rs/clap/) anyways. And clap supports [generating completions for various shells](https://docs.rs/clap/2.33.0/clap/struct.App.html#method.gen_completions)!

@Jarhmander thanks for testing!

PR #87, opened 2019-08-27
Use `string replace` instead of `sed`

james2doyle:
It seems like this was broken on my machine (OSX, 10.14.6) when using the completions. It would strip the `, ` but it would not add a proper newline. Instead it added nothing. So it just become a giant single line. Basically, it didn’t work.

With this change, the string gets split properly. Since this is fish shell, the `string replace` command is guaranteed to be there.


dbrgn:
Thanks! Not a fish user, but I'll just believe you that this works :slightly_smiling_face:

PR #89, opened 2019-10-03
Add compact output option

### LINKED ISSUE #88: Compact output
This is a bug I've meant to keep track of for a long time. I love tealdeer but the output drives me crazy because it's not compact enough compared to other tldr clients.

Official client:
image

Tealdeer:
image

As you can see in the screenshot it takes twice the space.

If I remove the excessive newlines, here is what it looks like (better imho):
image

I'd suggest to match the official client's output by default. I will give it a shot when I get the chance to, but feel free to fix it.

korrat:
This adds a config option to suppress empty lines in the output.

Closes #88.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,14 +180,25 @@ Example customization:
 
 ### Display
 
-Using `display` config you can customize whether the pager should be use by
-default or not (default `false`).
+In the `display` section you can configure the output format.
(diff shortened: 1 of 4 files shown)


dbrgn:
Thank you!

PR #95, opened 2019-12-17
Add -m / --markdown option for raw rendering

Plommonsorbet:
This is useful for when you want to open it up in a pager which doesn't
support ansi escape codes by default. In my use case it allows me to
open it inside vim without having to use an addon to convert ansi escape
codes to colours.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ These are the clients I tried but failed to compile or run:
         -u --update          Update the local cache
         -c --clear-cache     Clear the local cache
         -p --pager           Use a pager to page output
+        -m --markdown        Display the raw markdown instead of rendering it
         -q --quiet           Suppress informational messages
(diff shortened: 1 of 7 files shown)


dbrgn:
Thanks!

Hmm, that would make `tealdeer` a glorified `cat`, right? :smile:

Plommonsorbet:
Hi,

Yes essentially in that case it's only for finding the right file and displaying it. But I think that has it's value too, in my use case of tldr I use it normally in the terminal but if I want to copy something from the output that becomes quite awkward, or if I already have vim up then it would be nice to be able to just render it in a vim buffer.


It's a pretty minor change, but at least for me it allows me to not have to use more than one tool to view the tldr-pages. I'm sure other people could find it useful as well. :)

dbrgn:
Thanks!

PR #99, opened 2020-01-22
CI: Switch to GitHub Actions

dbrgn:


dbrgn:
Note to self: Update build badges in README

dbrgn:
Very nice!

This also fixes Windows compatibility that has been broken in #44 :slightly_smiling_face:

PR #101, opened 2020-02-03
Fix clippy warnings

jesdazrez:
Hi!

I noticed some clippy warnings while the AUR package was getting installed on my computer, so I fixed them.

Let me know if you want some adjustments :).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -37,14 +37,14 @@ pub enum RawColor {
 impl From<RawColor> for Color {
     fn from(raw_color: RawColor) -> Self {
         match raw_color {
-            RawColor::Black => Self::Black,
-            RawColor::Red => Self::Red,
-            RawColor::Green => Self::Green,
-            RawColor::Yellow => Self::Yellow,
-            RawColor::Blue => Self::Blue,
-            RawColor::Purple => Self::Purple,
-            RawColor::Cyan => Self::Cyan,
-            RawColor::White => Self::White,
+            RawColor::Black => Color::Black,
+            RawColor::Red => Color::Red,
+            RawColor::Green => Color::Green,
+            RawColor::Yellow => Color::Yellow,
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -35,16 +35,17 @@ pub enum RawColor {
 }
 
 impl From<RawColor> for Color {
+    #[must_use]
```
dbrgn: Hm, I don't see a big reason for these must use annotations. Also, the docs say:

> Not bad at all, this lint just shows places where you could add the attribute.

I'd remove all `#[must_use]` annotations and add `#[allow(clippy::must_use_candidate)]` to `src/main.rs` instead. (There are some more cases further down.)

jesdazrez: Ok, makes sense for me. In fact, after updating to 1.41, clippy no longer lints those.

jesdazrez:
Hi, sorry for the delay. I just updated the PR.

Thanks for the time you took to review it!

dbrgn:
Ugh, for some reason GitHub CI doesn't seem to run for forked pull requests. I fixed it in master and will push a merge commit to your branch.

dbrgn:
Ok, that's what I feared. Some of the features you used are not yet available on Rust 1.36. Unfortunately I did not notice those issues initially, because CI did not run.

I try to support the lowest stable Rust version that still works, because some distros package a Rust version that's not the latest.

Could you "backtrack" the changes until the CI passes again? I think the main issue is using `Self::` in enum matches.

(If you don't have time to do so, you can also let me know and I'll handle the necessary changes.)

jesdazrez:
Oops, my mistake for not noticing the MSRV in the README.

Let me try with 1.36. I will update the PR as soon as I can.

jesdazrez:
Done :)

dbrgn:
Great, thanks!

PR #108, opened 2020-03-07
Argument parsing: Switch from docopt to clap

### LINKED ISSUE #106: Switch command-line parsing to clap
It's faster and we might be able to auto-generate the autocompletion files (see comment in #86).

(Generating autocompletion doesn't need to be part of this issue.)

Potentially related to #79.

dbrgn:
Fixes #106.


src/main.rs
```diff
@@ -414,31 +403,9 @@ fn main() {
     }
 
     // Some flags can be run without a command.
-    if !(args.flag_update || args.flag_clear_cache || args.flag_config_path) {
-        eprintln!("{}", USAGE);
+    if !(args.update || args.clear_cache || args.config_path) {
```
dbrgn: @pksunkara how would I implement this logic in `clap_derive`?

In structopt, it seems that for custom help printing I could simply use [`args.clap().print_help()`](https://docs.rs/structopt/0.3.11/structopt/trait.StructOpt.html?search=#tymethod.clap) but that doesn't seem to be available in clap itself.

pksunkara: I don't get this part. Are you trying to print help if no other arguments are given?

dbrgn: Some flags (e.g. `--update`, `--clear-cache` or `--config-path`) can be used without a command. Other flags (e.g. `--pager` or `--os`) can only be used if a command is provided.

The command is a TLDR page in this case, so we can't create a finite list of clap subcommands.

pksunkara:
```
#[derive(Clap, Debug)]
#[clap(about = "A fast TLDR client", author, version)]
struct Args {
    /// The command to show (e.g. `tar` or `git log`)
    command: Vec<String>,

    /// Use a pager to page output
    #[clap(short = "p", long = "pager", requires = "command")]
    pager: bool,

    /// Show config file path
    #[clap(long = "config-path")]
    config_path: bool,
}
```

Clap v2 docs should mostly still be correct regarding function names. https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.requires

Also you don't need the verison option.

dbrgn: > Clap v2 docs should mostly still be correct regarding function names.

Ah, I see! Will try to implement those dependencies.

> Also you don't need the verison option.

I need it if I want the short argument to be `-v` instead of `-V`, right?

pksunkara: True yeah. I didn't realize you had it short.

dbrgn: @pksunkara: I tried this:

```rust
#[derive(Clap, Debug)]
#[clap(about = "A fast TLDR client", author, version)]
struct Args {
    /// The command to show (e.g. `tar` or `git log`)
    command: Vec<String>,
...
    /// Display the raw markdown instead of rendering it
    #[clap(short = "m", long = "markdown", requires = "command")]
    markdown: bool,
...
```

However, that `requires` doesn't seem to work as intended because `command` is a Vec, so an empty command list is still an empty command list. I'd need something like "requires that `command` is not empty".

pksunkara: `Option<Vec<String>>` should work IIRC.

pksunkara: If not, try `Vec<string>` with `min_values = 1`

Docs regarding those types, https://docs.rs/structopt/0.3.11/structopt/#type-magic. Mostly still be same to the clap_derive.

dbrgn: Clap doesn't like `Option<Vec<_>>`, since it doesn't know whether a missing argument should be `None` or `Some(vec![])`. However, `min_values = 1` seems to work beautifully :slightly_smiling_face:

pksunkara: Please do keep me up to date on the perf/size/valgrind improvements you notice.

pksunkara:
You should also be able to generate completions files using https://github.com/clap-rs/clap/blob/master/clap_generate/src/lib.rs

src/main.rs
dbrgn: Ok, I pushed an update. Two issues I noticed:

- Even though I set `min_values = 1` on `command`, when I run the binary without any arguments, it still runs but doesn't print any output. In that case, args are: `Args { command: [], list: false, render: None, os: None, update: false, clear_cache: false, pager: false, quiet: false, markdown: false, config_path: false, seed_config: false, version: false }` Can I make the command required somehow unless a flag that doesn't require it is used?

- I added a conflict between `command` and `--render`:

    ```diff
         /// Render a specific markdown file
    -    #[clap(short = "f", long = "render")]
    +    #[clap(short = "f", long = "render", conflicts_with = "command")]
         render: Option<String>,
    ```

  However, when padding in both a command and the `--render` flag, I get this error message:

   ```
     [clap]danilo@x2000:tealdeer$ cargo run -- --render tar.md gz
       Finished dev [unoptimized + debuginfo] target(s) in 0.05s
          Running `target/debug/tldr --render tar.md gz`
    error: The argument '--render <render>' cannot be used with '--render <render>'
    
    USAGE:
        tldr [FLAGS] [OPTIONS] [command]...
    
    For more information try --help
    ```

    Bug?

pksunkara: https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.required

Can you move `command` to the end of the struct? It might fix the bug. `command` here is a positional argument here which means most people generally put it at the end. Either way, please create an issue on clap repo.

pksunkara: My bad, wrong link.  Check https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.required_unless and other functions.

dbrgn: Ah, the bug is https://github.com/clap-rs/clap/issues/1622! Reordering the struct does not help.

pksunkara: So, that bug is being fixed today.

dbrgn: Just saw it, awesome :slightly_smiling_face:

dbrgn:
I'll work on this again when clap v3 is released. We cannot release to crates.io if we depend on unreleased crates. And it seems that clap v3 still has quite some work ahead: https://github.com/clap-rs/clap/milestone/76

pksunkara:
We released a beta on crates.io. But yes, we still have some work to finish.

dbrgn:
I wonder if standalone flags (like `--show-paths`) should be changed to subcommands instead... Might be cleaner, but might be a breaking change for some of them.

niklasmohrin:
@dbrgn I don't think we should do that. If I recall correctly the spec requires `--update` to be a flag and the subcommands could clash with actual pages (maybe someone writes a program called `show-paths` that gets its own page)

dbrgn:
Hah, that's a good point. I'm convinced 🙂

dbrgn:
Still open:

- Spec compliance
- Auto-generate completion files

However, these things can be handled in separate issues. I think it would be good if we could merge this soon, the current version should have feature-parity with master/docopt.

Current status regarding binary size:

- `main` at 3f4910030770e298d0187b601226e28bf1c07450: 8483 KiB (unstripped), 5355 KiB (stripped)
- `clap`: 7581 KiB (unstripped), 4519 KiB (stripped)

...so we shaved off almost a megabyte.

@niklasmohrin could you help testing this, to see if there's any regression? I hope I didn't overlook anything 🙂

dbrgn:
Dependency count increased (153 vs previously 140), but that isn't necessarily bad. Compilation time for a clean release build went down slightly on my 12-core machine (48.9s now vs 53.1s previously).

dbrgn:
I also took a look in callgrind (see https://github.com/dbrgn/tealdeer/issues/106#issuecomment-596026250 for comparison).

Assembly instruction count for `tldr::main` went down from 9'214'656 to 515'050 😄 Nice!

Hyperfine benchmark comparison on my machine:

```
# docopt
Benchmark #1: target/release/tldr tar
  Time (mean ± σ):      15.9 ms ±   0.7 ms    [User: 1.8 ms, System: 2.7 ms]
  Range (min … max):    13.2 ms …  17.7 ms    150 runs
```

```
clap
Benchmark #1: target/release/tldr tar
  Time (mean ± σ):      12.4 ms ±   0.5 ms    [User: 1.0 ms, System: 2.2 ms]
  Range (min … max):    10.2 ms …  13.5 ms    150 runs
```

Hah, that's an awesome improvement! 🎉 With this, we clearly beat fast-tldr-gh and are now the second fastest implementation in the benchmark. Only the Zig based implementation is faster, but it doesn't support user configuratoin at all, so it's an unfair comparison.

dbrgn:
@niklasmohrin rebased against current master and squashed multiple commits.

dbrgn:
I also used this occasion to rename `-m / --markdown` to `-r / --raw` and upgraded to Clap v3 beta 5 (which was released an hour ago).

niklasmohrin:
On first look, I didn't like the colors. I think they **do** improve readability though. I did a quick check and tools like `bat` also have colored help, although `bat` supports different color schemes for its main output.

I think we can keep the colors :)

niklasmohrin:
I just noticed that the examples are not in the `--help` anymore, is this planned?

dbrgn:
Regarding the colors: I'm not really sure if it's an improvement in readability, or just a distraction. I myself am not a big fan of them for that reason. I asked two people what they think about it, and both said "hm, whether or not that looks good depends on the color scheme of the terminal", and both said that they don't think colors are that important for help texts. (For the error messages, I think the colors are useful!)

There's another reason why colored help may be a bad idea: We offer controlling the page output colors through our config file. Some users might want to make help text colors configurable as well. I don't want to implement such a feature, I think it's unneeded. And if someone would want to turn off colors for the help text by setting `NO_COLOR=1`, that would disable tldr colors as well (which is probably not intended).

I think I'll try to disable help colors again.

Regarding examples, you are right, those should be added back.

niklasmohrin:
:man_shrugging: Sure, colors are not a must-have.

dbrgn:
> I just noticed that the examples are not in the `--help` anymore, is this planned?

When researching on how to implement this, I stumbled upon my own question 😂 https://github.com/clap-rs/clap/discussions/2477

dbrgn:
Yeah, I'm a bit unsure about the examples section. In essence, this is a tldr page, but worse.

Maybe we could simply redirect users to `tldr tldr` instead`?

dbrgn:
Here's an implementation of that suggestion:

image

@niklasmohrin do you like it? Or do you think the examples should still be provided? The tldr page only contains the shared features according to the client spec of course, so it will always be a subset of the functionality.

Another approach would be linking to https://dbrgn.github.io/tealdeer/. That might actually be the best solution, because we have much more possibilities to structure the content to make it easier to read / understand.

I think docs are valuable, but I'm not sure it should be part of the `--help` text.

niklasmohrin:
Yep, that seems appropriate. We need to add these examples explicitly in the docs though

dbrgn:
🎉 finally!

PR #109, opened 2020-03-08
Windows: Only warn about missing pager if flag set

rgwood:
Hi, thanks for writing tealdeer! It's by far my favourite tldr client.

### PR

I recently installed tealdeer on my Windows machine, and I noticed that the following message is printed on _every_ invocation:

> Warning: -p / --pager flag not available on Windows!

This is a bit annoying to see every time, so I implemented a change to only show the message if the -p/--pager flag is true.

### Testing performed

All tests performed on Windows 10 (1909). Ran `cargo test` successfully before+after the change. Ran `tldr`, `tldr ls`, `tldr -p ls`, and `tldr --pager ls` and confirmed expected output.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -168,8 +168,10 @@ fn configure_pager(args: &Args, enable_styles: bool) {
 }
 
 #[cfg(target_os = "windows")]
-fn configure_pager(_args: &Args, _enable_styles: bool) {
-    eprintln!("Warning: -p / --pager flag not available on Windows!");
+fn configure_pager(args: &Args, _enable_styles: bool) {
+    if args.flag_pager {
+        eprintln!("Warning: -p / --pager flag not available on Windows!");
+    }
 }
 
 /// Check the cache for freshness


dbrgn:
Hi, thanks! Sorry for not replying earlier, the notification slipped through :slightly_smiling_face: Good catch with the warning!

There's one issue with the fix: The `display.use_pager` config variable defaults to `false`, but if the user explicitly enables paging in the config, the warning should probably appear. With your fix, the warning only appears when using the command line flag.

The problem is [already fixed](https://github.com/dbrgn/tealdeer/pull/115/files#diff-639fbc4ef05b315af92b4d836c31b023R320-R322) in #115 so you don't need to update your commit, but if you wanted you could add an integration test to ensure that a regular call does not result in this warning on Windows. For inspiration, see `test_update_cache` in `tests.lib.rs`. The test could be gated with `#[cfg(target_os = "windows")]`.

If you don't have time or motivation, just let me know, then I can add the test myself!

rgwood:
Thanks for the detailed response! I probably won't have time to add that test but I appreciate the info.

dbrgn:
Ok, no problem! Reopening to remind myself to add the test before the next release.

PR #115, opened 2020-04-30
Implement automatic cache updates

### LINKED ISSUE #78: Avoid loading config twice
Right now, due to #44, the config is loaded twice:

```
$ RUST_LOG=tldr=debug cargo run --features logging -- tar
DEBUG 2019-03-11T10:16:47Z: tldr::config: Loading config
Cache wasn't updated for more than 30 days.
DEBUG 2019-03-11T10:16:47Z: tldr::config: Loading config
DEBUG 2019-03-11T10:16:47Z: tldr::formatter: Ignoring title
DEBUG 2019-03-11T10:16:47Z: tldr::formatter: Detected command name: tar

  Archiving utility.
  Often combined with a compression method, such as gzip or bzip.
```

This should be avoided for performance reasons.

### LINKED ISSUE #90: Automatic cache updates
As discussed in #73, there could be a config option where - if it's set - the cache automatically refreshes itself when it's outdated.

Implementation notes (copied from [here](https://github.com/dbrgn/tealdeer/pull/73#issuecomment-453783054)):

This would require a slight reorganization of the `main` function:

- Handle `args.flag_config_path` and `args.flag_seed_config` right after handling `args.flag_version` (simply move up the code blocks)
- Move the config loading code in `print_page` into the main function, right after the `args.flag_seed_config` handling block
- In `src/config.rs`, add the new config elements to the `RawConfig` and `Config` structs and add the defaults to `RawConfig::new()`.

The default configuration can be generated with `tldr --seed-config`. My suggestion would be:

```
[updates]
auto_update = false
```

(The `auto_update_interval_hours` could then be set to `MAX_CACHE_AGE * 24` by default.)

ilai-deutel:
This PR implements automatic updates of the cache.

This behavior is disabled by default and can be enabled in `config.toml`:

```toml
[updates]
auto_update = true
```

Fixes #90
Fixes #78

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,7 @@ use crate::error::TealdeerError::{self, ConfigError};
 
 pub const CONFIG_FILE_NAME: &str = "config.toml";
 pub const MAX_CACHE_AGE: Duration = Duration::from_secs(2_592_000); // 30 days
+const DEFAULT_UPDATE_INTERVAL_HOURS: u64 = MAX_CACHE_AGE.as_secs() / 3600; // 30 days
 
 fn default_underline() -> bool {
     false
@@ -118,12 +119,28 @@ struct RawDisplayConfig {
     pub use_pager: bool,
 }
 
-#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+/// Serde doesn't support default values yet (tracking issue:
+/// https://github.com/serde-rs/serde/issues/368), so we need to wrap DEFAULT_UPDATE_INTERVAL_HOURS
+/// in a function to be able to use #[serde(default = ...)]
+const fn default_auto_update_interval_hours() -> u64 {
+    DEFAULT_UPDATE_INTERVAL_HOURS
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 struct RawUpdatesConfig {
     #[serde(default)]
     pub auto_update: bool,
-    #[serde(default)]
+    #[serde(default = "default_auto_update_interval_hours")]
     pub auto_update_interval_hours: u64,
+}
+
+impl Default for RawUpdatesConfig {
+    fn default() -> Self {
+        Self {
+            auto_update: false,
+            auto_update_interval_hours: DEFAULT_UPDATE_INTERVAL_HOURS,
+        }
+    }
 }
 
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
@@ -146,8 +163,6 @@ impl RawConfig {
         raw_config.style.example_code.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.underline = true;
-        raw_config.display.use_pager = false;
-        raw_config.updates.auto_update_interval_hours = MAX_CACHE_AGE.as_secs() / 3600;
 
         raw_config
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@ use std::env;
 use std::fs;
 use std::io::{Error as IoError, Read, Write};
 use std::path::PathBuf;
+use std::time::Duration;
 
 use ansi_term::{Color, Style};
 use app_dirs::{get_app_root, AppDataType};
@@ -12,6 +13,7 @@ use toml;
 use crate::error::TealdeerError::{self, ConfigError};
 
 pub const CONFIG_FILE_NAME: &str = "config.toml";
+pub const MAX_CACHE_AGE: Duration = Duration::from_secs(2_592_000); // 30 days
 
 fn default_underline() -> bool {
     false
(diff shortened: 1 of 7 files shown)


tests/lib.rs
```diff
@@ -352,3 +352,61 @@ fn test_list_flag_rendering() {
...
+
+    // The cache is not updated with a subsequent call
+    check_cache_updated(false);
+
+    let (_, orig_mtime) = utime::get_file_times(&cache_file_path).unwrap();
+
+    // We update the modification and access times such that they are about 23 hours from now.
+    // auto-update interval is 24 hours, the cache should not be updated
+    let new_mtime = orig_mtime - 82_800;
+    utime::set_file_times(&cache_file_path, new_mtime, new_mtime).unwrap();
```
ilai-deutel, line 401: Since the `utime` library expects a `u64`, I had to set the mtime relative to `orig_mtime`, as opposed to something more explicit like `let new_mtime = std::time::SystemTime::now() - Duration::from_secs(82_800)`.

The [`filetime`](https://crates.io/crates/filetime) crate has a [`FileTime::from_system_time()`](https://docs.rs/filetime/0.2.9/filetime/struct.FileTime.html#method.from_system_time) function and could be used to set the atime and mtime. The crate is actively maintained, unlike `utime` (last updated in October 2018). This is not critical since `utime` works fine here.

dbrgn:
Cooo, thanks for the PR! I hope to review it sometime next week, I'm quite busy at the moment. If someone sees this and wants to speed up the process, reviews by other people certainly help :slightly_smiling_face: 

@ilai-deutel windows CI currently fails, seems to be a permission issue. Maybe you could take a look at that...

ilai-deutel:
> @ilai-deutel windows CI currently fails, seems to be a permission issue. Maybe you could take a look at that...

It looks like both the `utime` and `filetime` libraries have issues with retrieving/setting modified times of Windows folders. I created a PR to fix this, alexcrichton/filetime#54

ilai-deutel:
@dbrgn Fixed!

src/config.rs
```diff
@@ -116,12 +118,22 @@ struct RawDisplayConfig {
     pub use_pager: bool,
 }
 
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+struct RawUpdatesConfig {
+    #[serde(default)]
+    pub auto_update: bool,
+    #[serde(default)]
+    pub auto_update_interval_hours: u64,
```
dbrgn, line 126: With `#[serde(default)]` this will default to 0, causing every invocation to auto update if `auto_update` is set to true. We should probably set an explicit serde default instead. (The default set in `RawConfig::new` only applies if no config file exists, that could be removed if we have an explicit default.)

I think `#[serde(default = ...)]` should work, but haven't tested it.

tests/lib.rs
```diff
@@ -352,3 +357,59 @@ fn test_list_flag_rendering() {
...
+    check_cache_updated(false);
+
+    // We update the modification and access times such that they are about 25 hours from now.
+    // auto-update interval is 24 hours, the cache should be updated
+    let new_mtime = SystemTime::now() - Duration::from_secs(90_000);
+    filetime::set_file_mtime(&cache_file_path, new_mtime.into()).unwrap();
+    check_cache_updated(true);
+
+    // The cache is not updated with a subsequent call
+    check_cache_updated(false);
```
dbrgn, line 414: Nice and clean test! :+1:

src/config.rs
```diff
@@ -118,14 +119,30 @@ struct RawDisplayConfig {
     pub use_pager: bool,
 }
 
-#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+/// Serde doesn't support default values yet (tracking issue:
+/// https://github.com/serde-rs/serde/issues/368), so we need to wrap DEFAULT_UPDATE_INTERVAL_HOURS
+/// in a function to be able to use #[serde(default = ...)]
```
dbrgn, line 124: Ah, wasn't aware that this is a function, not a literal.

src/config.rs
```diff
@@ -118,14 +119,30 @@ struct RawDisplayConfig {
...
 }
 
+impl Default for RawUpdatesConfig {
+    fn default() -> Self {
+        Self {
+            auto_update: false,
+            auto_update_interval_hours: DEFAULT_UPDATE_INTERVAL_HOURS,
+        }
+    }
+}
```
dbrgn, line 144: If we use `#[serde(default...)]`, why do we need to manually impl `Default`? Wouldn't serde handle that for us, and delegate to the field default values when getting the default value for a `#[derive(Default)]` struct?

ilai-deutel: I thought so too, but it doesn't appear that it's the case:

- Output of `println!("{:?}", RawUpdatesConfig::default())` with the current implementation:

```rust
RawUpdatesConfig { auto_update: false, auto_update_interval_hours: 720 }
```

- Output of `println!("{:?}", RawUpdatesConfig::default())` with `derive(Default)`:

```rust
RawUpdatesConfig { auto_update: false, auto_update_interval_hours: 0 }
```

dbrgn: Ah I see. The `#[serde(default = ...)]` attribute only changes deserialization, but does not influence the `Default` implementation (which is used to seed a config).

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=387ce3b5e2c29cc869ca648cf0ee8dba

We could use a newtype with a different `Default`, then you don't need the custom serde function anymore. However, I'm not sure yet whether that's cleaner or not. What do you think?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f03be819dc29496526d3d93ec6d51187

ilai-deutel: I think both options makes sense, I don't have a strong preference. Creating a new type here doesn't add too much complexity, but it could become messy if there are more fields like this in the future.

A third option I can think of is getting rid of the `Default` implementations altogether and initializing raw configs with `toml::from_str("")` instead of `RawConfig::new()`. This allows to unify the defaults (always use the serde defaults), but the performance may not be as good.

dbrgn: Ok, let's stick to your approach for now. Should be simple enough to understand.

ilai-deutel:
Sorry it took so long, I updated the README.

dbrgn:
Thank you @ilai-deutel!

PR #124, opened 2020-08-02
Switch to maintained app_dirs

kornelski:


dbrgn:
Thanks! References to the crate need to be updated though:

```
26 | use app_dirs::AppInfo;
   |     ^^^^^^^^ use of undeclared type or module `app_dirs`
```

dbrgn:
The failing test is probably due to #126.

@niklasmohrin in case you'd like to cherry-pick your test fix into a separate pull request, that would be welcome!

niklasmohrin:
Sure thing, see #128

dbrgn:
Thanks! #128 is merged, so this branch can be rebased.

PR #125, opened 2020-08-10
Support multiple languages

### LINKED ISSUE #96: Support multiple languages
`tealdeer` currently fetches all languages from the tldr.sh API.  However, it only ever displays English documentation.  It seems that `tealdeer` should offer a flag/config option that would display documentation from other languages, when available.  (Or maybe it should read the `$LANG` environmental variable?)

Alternatively, if `tealdeer` is not going to display non-English documentation, it probably should not save the non-English docs in the user's cache (since they take up space without providing any benefit).

In an ideal world, `tealdeer` would offer a user the option of selecting their language(s) before downloading the cache, but that seems like a lower priority.

### LINKED ISSUE #126: Outdated test data
Hey, I was working on #96 and noticed that the file `tar-markdown.expected` in the `tests` directory was outdated. I replaced the content with the current version from the `tldr` repo in #125, but I still wanted to document it here. Keep up the good work!

niklasmohrin:
Fixes #96, fixes #126 

Hey everybody! This is the first time I am contributing to a rust project, so please don't refrain from commenting.

I only started working on this and its not ready yet, but I wanted to open the draft to open the comment thread. I think the basic control flow is pretty close to how it will be implemented in the final version. I have not read about the standards for the `LANG` and `LANGUAGE` environment variables yet. I saw that there is the [env-lang](https://crates.io/crates/env-lang) crate to parse the `LANG` variable, so I might use this.

Should I `cargo fmt` the files that I am editing in a final commit after all the others?

Let me know what you think

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::Read;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use app_dirs::{get_app_root, AppDataType};
 use flate2::read::GzDecoder;
@@ -142,41 +142,53 @@ impl Cache {
         }
     }
 
+    /// Check for pages for a given platform in one of the given languages.
+    fn find_page_for_platform(
+        name: &str,
+        cache_dir: &Path,
+        platform: &str,
+        language_dirs: &[String],
+    ) -> Option<PathBuf> {
+        language_dirs
+            .iter()
+            .map(|lang_dir| cache_dir.join(lang_dir).join(platform).join(name))
+            .find(|path| path.exists() && path.is_file())
+    }
+
     /// Search for a page and return the path to it.
-    pub fn find_page(&self, name: &str) -> Option<PathBuf> {
-        // Build page file name
+    pub fn find_page(&self, name: &str, languages: &[String]) -> Option<PathBuf> {
         let page_filename = format!("{}.md", name);
 
         // Get platform dir
-        let platforms_dir = match Self::get_cache_dir() {
-            Ok(cache_dir) => cache_dir.join("tldr-master").join("pages"),
+        let cache_dir = match Self::get_cache_dir() {
+            Ok(cache_dir) => cache_dir.join("tldr-master"),
             Err(e) => {
                 log::error!("Could not get cache directory: {}", e);
                 return None;
             }
         };
 
-        // Determine platform
-        let platform = self.get_platform_dir();
+        let lang_dirs: Vec<String> = languages
+            .iter()
+            .map(|lang| {
+                if lang == "en" {
+                    String::from("pages")
+                } else {
+                    format!("pages.{}", lang)
+                }
+            })
+            .collect();
 
-        // Search for the page in the platform specific directory
-        if let Some(pf) = platform {
-            let path = platforms_dir.join(&pf).join(&page_filename);
-            if path.exists() && path.is_file() {
-                return Some(path);
+        // Try to find a platform specific path first.
+        if let Some(pf) = self.get_platform_dir() {
+            let pf_path = Self::find_page_for_platform(&page_filename, &cache_dir, pf, &lang_dirs);
+            if pf_path.is_some() {
+                return pf_path;
             }
         }
 
-        // If platform is not supported or if platform specific page does not exist,
-        // look up the page in the "common" directory.
-        let path = platforms_dir.join("common").join(&page_filename);
-
-        // Return it if it exists, otherwise give up and return `None`
-        if path.exists() && path.is_file() {
-            Some(path)
-        } else {
-            None
-        }
+        // Did not find platform specific results, fall back to "common"
+        Self::find_page_for_platform(&page_filename, &cache_dir, "common", &lang_dirs)
     }
 
     /// Return the available pages.
--- a/src/main.rs
+++ b/src/main.rs
@@ -494,7 +494,7 @@ fn main() {
 
 #[cfg(test)]
 mod test {
-    use crate::{Args, OsType, USAGE};
+    use crate::{get_languages, Args, OsType, USAGE};
     use docopt::{Docopt, Error};
 
     fn test_helper(argv: &[&str]) -> Result<Args, Error> {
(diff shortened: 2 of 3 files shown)


niklasmohrin:
Okay, so I did some research and I think querying the `tldr` pages should now work like expected. I don't know whether languages should be considered when listing the pages, what do you guys think? Do you want me to write some tests for the new behavior?

While developing, I recognized, that some test data was outdated, so I fixed that in one commit (I also opened #126 for tracking).

Greetings

dbrgn:
Great, thanks @niklasmohrin for your contribution!

Expectation management: I hope to review this soon, but can't promise any date since there are already a few other open PRs on my TODO list. Also, currently I have to take advantage of the nice summer weather to hike and paraglide during my free days :smile: 

In any case, if I don't review this in the next 2-3 weeks, feel free to ping me! Also, reviews by other people are always welcome and helpful.

src/cache.rs
```diff
@@ -142,41 +142,54 @@ impl Cache {
         }
     }
 
+    /// Check for pages for a given platform in one of the given languages.
+    fn find_page_for_platform(
+        &self,
+        name: &str,
+        cache_dir: &PathBuf,
```
dbrgn, line 149: Passing in a `&PathBuf` is a bit strange, like passing in a `&String`. Instead, use a `&Path`.

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
...
     }
 
+    // Language list according to
+    // https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language
+    let languages = if let Some(ref lang) = args.flag_language {
+        // Language overwritten by console argument
+        vec![lang.clone()]
+    } else if let Ok(lang) = std::env::var("LANG") {
+        let locales = std::env::var("LANGUAGE").unwrap_or_default();
+        let mut locales: Vec<&str> = locales.split(':').collect();
```
dbrgn, line 419: I'd rename the first variable for clarity:

```suggestion
        let language = std::env::var("LANGUAGE").unwrap_or_default();
        let mut locales: Vec<&str> = language.split(':').collect();
```

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
...
+        // According to the client specification, these should be appended
+        locales.push(&lang[..]);
+        locales.push("en");
+
+        let mut lang_list = Vec::new();
+        for locale in locales.iter() {
```
dbrgn, line 425: I think you can leave away the `.iter()` part.

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
...
+            if locale.len() >= 5 && locale.chars().nth(2) == Some('_') {
+                // Language with county code
```
dbrgn, line 427: Typo

```suggestion
                // Language with country code
```

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
...
+                lang_list.push(&locale[..5]);
+            }
+            if locale.len() >= 2 {
+                // Language code
+                lang_list.push(&locale[..2]);
+            }
```
dbrgn, line 433: Note: According to the spec, the `LANGUAGE` var may contain the values `POSIX` or `C`, which should be ignored. In this case, they would be converted to `PO` and `C`, respectively.

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
...
+        }
+
+        lang_list.dedup();
```
dbrgn, line 436: Note: This will only deduplicate consecutive repeated values, so `LANG=es LANGUAGE=fr,es,de` would result in `fr,es,de,es`.

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
         process::exit(0);
     }
 
+    // Language list according to
+    // https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language
+    let languages = if let Some(ref lang) = args.flag_language {
+        // Language overwritten by console argument
+        vec![lang.clone()]
+    } else if let Ok(lang) = std::env::var("LANG") {
```
dbrgn, line 417: According to the spec, if `LANG` contains `C` or `POSIX`, it should be ignored.

src/main.rs
```diff
@@ -407,6 +409,38 @@ fn main() {
         process::exit(0);
     }
 
+    // Language list according to
```
dbrgn, line 412: This code would probably benefit from being moved into a dedicated function (`get_language_list`). That would also be a nice target for a few unit tests (containing at least the examples from the spec).

dbrgn (changes requested):
Finally got around to review this! Here's the first round of feedback :slightly_smiling_face:

src/main.rs
niklasmohrin: I think I will just move this method chaining into one assignment for clarity (I don't think having so many variables named "lang", "languages", "locale" and "language" is good either, but good point, this is also weird :D)

src/main.rs
niklasmohrin: I added some tests in `main.rs`. Since I didn't find a sane way to mock environment variables, I constructed the `get_languages` function around the environments variables and kept the logic around the `--language` flag out of it. Instead, this is handled in `main`.

Let me know what you think and what other tests you want to see

src/main.rs
```diff
@@ -290,6 +293,46 @@ fn get_os() -> OsType {
     OsType::Other
 }
 
+fn get_languages(
+    env_lang: Result<String, std::env::VarError>,
+    env_language: Result<String, std::env::VarError>,
```
niklasmohrin, line 298: Note to self: Use `Option` instead

dbrgn: Good point, an `Option<String>` would be better!

dbrgn (approved):
Hi @niklasmohrin, sorry for the long delay (again)! There are a few optimizations I'd like to see (I think some of the code has potential for reducing the amount of string creation / heap allocations), but the code seems to be correct, easy to read and well tested.

I'll merge it for now and will do adjustments myself, to avoid another long ping-pong with me not being as responsive as I'd like to 🙂

Thanks for your contribution!

dbrgn:
Rebased and merged in 1db4b409bc74b8db47cf2e9d6b0db4a845390f79!

PR #133, opened 2020-08-30
Run cargo update

dbrgn:
This includes a security fix of the bumpalo crate, as reported by `cargo audit`:

https://rustsec.org/advisories/RUSTSEC-2020-0006


dbrgn:
Ugh, *of course* running cargo update breaks builds on older Rust versions :roll_eyes: Trying to stick to a fixed minimal supported Rust version is a futile attempt in any real-world project.

dbrgn:
Just a single dev dependency, fortunately.

Issue #98, opened 2020-01-07
Allow adding custom entries to pages


dbrgn:
Sometimes I want to add custom entries to pages that already contain 8 entries. The tldr guidelines only allow up to 8 entries per page.

Maybe we could set up a way to register "patchfiles" with tealdeer (e.g. by placing them somewhere in the `.config/tealdeer/` directory). Those patchfiles would be appended to the corresponding pages when updating the cache.

jasonkena:
Adding custom pages would also be nice

dmaahs2017:
Implementing this in #142

PR #142, opened 2020-09-22
Support custom pages and patches

dmaahs2017:
Feature implementation for issue #98
This is a feature that lets you create custom tldr pages locally. 
You can set a directory where tealdeer will search for these local pages with the env variable `TEALDEER_CUSTOM_PAGES_DIR`. If this is not set then it will use the `tldr-master/pages.custom`

If you give a custom tldr page the same name as an existing one, then your custom one will be appended to it. 

<details>
  <summary>Example with existing tldr page:</summary>
  <image src="https://user-images.githubusercontent.com/18738072/93942448-07215200-fcf6-11ea-9cfd-f54518f1c3a3.png"></image>
</details>

<details>
  <summary>Example with no existing tldr page:</summary>
  <image src="https://user-images.githubusercontent.com/18738072/93942633-67b08f00-fcf6-11ea-8427-adc70c2a17c8.png"></image>
</details>

I find the value added here is that overtime a person can build their own collection of common use-cases to a particular command that they use semi-frequently. I created this because I would create my own repository of command specific use-cases, and to document personalized scripts for myself.

Side note: This is my first real open-source contribution, and I'm not super familiar w/ the ins and outs of rust. Please leave suggestions for improvements :smile: . 

# Update, Dec 18. Implemented suggested changes from @dbrgn 
* Implemented .page and .patch differences. {}.page files overwrite existing tldr entries. {}.patch appends to existing entries.
* Now using the config file for custom_pages_dir, default directory is `CONFIGDIR/pages`
* ~Also added 2 flags, `--patch` and `--overwrite`. These commands create and or open the .page or .patch files in your `$EDITOR`. I added this because I found it was annoying to constantly switch the the custom page directory. Also this enables people to very quickly make changes to their custom patches/pages~

<details>
  <summary>Example of a .patch file in work</summary>
  <image src="https://user-images.githubusercontent.com/18738072/102593233-34558480-40da-11eb-9fcc-9f9522eaa956.png"></image>
</details>
<details>
  <summary>Example of a new tldr page with a .page file</summary>
  <image src="https://user-images.githubusercontent.com/18738072/102593289-4800eb00-40da-11eb-98ac-0c173eb25751.png"></image>
</details>
<details>
  <summary>Example of overwriting an existing tldr page using a .page file</summary>
  <image src="https://user-images.githubusercontent.com/18738072/102593355-5f3fd880-40da-11eb-8f09-a91455e4ccf3.png"></image>
</details>
<details>
  <summary>How the config looks now</summary>
  <image src="https://user-images.githubusercontent.com/18738072/102660551-d019db80-4140-11eb-8a93-0de89d9c7270.png"></image>
</details>


src/cache.rs
```diff
@@ -160,22 +160,50 @@ impl Cache {
...
             }
-        }
+        } else {
+            None
+        };
 
         // If platform is not supported or if platform specific page does not exist,
         // look up the page in the "common" directory.
-        let path = platforms_dir.join("common").join(&page_filename);
+        let path = if path.is_none() {
```
dmaahs2017, line 176: Don't accept this yet, just noticed a bug here, will fix tomorrow. Path is overriding to none when the previous had Some. This is incorrect.

dmaahs2017: Fixed

dvergeylen:
Hey,

This looks very promising, it there any plan to merge it anytime soon? 😇

Keep up the great work! 💯

dmaahs2017:
> Hey,
> 
> This looks very promising, it there any plan to merge it anytime soon? 😇
> 
> Keep up the great work! 💯

Great to hear back from you! 😊

I was just waiting on a code review, since I'm fairly new to Rust.

If you don't have any suggestions then it should be good to merge

src/cache.rs
```diff
@@ -160,22 +160,49 @@ impl Cache {
...
-        let path = platforms_dir.join("common").join(&page_filename);
+        let common_path = platforms_dir.join("common").join(&page_filename);
+        let common_path = if common_path.exists() && common_path.is_file() {
+            Some(common_path)
+        } else {
+            None
+        };
+
+        // Get custom pages directory
+        let custom_env = if let Ok(value) = env::var("TEALDEER_CUSTOM_PAGES_DIR") {
```
dbrgn, line 184: I would prefer if this would be a configuration file option, and not an environment variable.

Maybe like this?

```toml
[directories]
custom_pages_dir = "/abs/path/to/dir"
```

We could also use a default value in the config directory as fallback, e.g. `CONFIGDIR/pages/`, if not set. That way, custom pages could be added without having to set up a config variable.

src/cache.rs
```diff
@@ -160,22 +160,49 @@ impl Cache {
...
+        };
+
+        // Return pages if they exists, otherwise give up and return `None`
+        match (pf_path, common_path, custom_path) {
+            (Some(pfp), _, Some(cup)) => Some(vec![pfp, cup]),
+            (Some(pfp), _, None) => Some(vec![pfp]),
+            (None, Some(cop), Some(cup)) => Some(vec![cop, cup]),
+            (None, Some(cop), None) => Some(vec![cop]),
+            (None, None, Some(cup)) => Some(vec![cup]),
+            (None, None, None) => None,
```
dbrgn, line 205: The way this is implemented right now means that if a platform specific page exists, instead of a single path lookup we now need to do three (because the early return was removed). Since I/O can be slow, this should be avoided if possible. For example, if a platform-specific page exists, the common directory lookup doesn't need to be done.

Your current approach always appends the custom pages to the existing pages. However, in some cases people might want to replace entire pages, or add a page if it does not exist yet. So here's an idea:

- If a page filename ends with `.page`, it should replace the entire page, no matter whether it already exists or not
- If a page filename ends with `.patch`, it should be appended to an existing page, if it exists.

The logic could be implemented as follows:

1. Look up custom page (`<name>.page`). If it exists, return it directly,
2. Look up custom patch (`<name>.patch`). If it exists, store it in a variable.
3. Look up platform-specific page. If it exists, return page plus the patch above (if set).
4. Look up page from the "common" directory. If it exists, return page plus the patch above (if set).
5. As a fallback, return an empty vector.

What do you think?

dbrgn (changes requested):
Hi @dmaahs2017, sorry for the long delay, especially since this is your first open source contribution. I'm currently quite busy at work and with other projects, so that leaves less time for maintaining tealdeer :slightly_smiling_face: 

I left two comments!

src/cache.rs
dmaahs2017: Sounds great I'll make those changes after Thanksgiving, thanks for the feedback!

dmaahs2017: Changes made, let me know what you think 😄

jasonkena:
Any updates?

dmaahs2017:
> Any updates?

Just waiting on @dbrgn to review again.

But you can build it from my fork until he gets the chance to review again.

dbrgn:
Yeah, sorry for the review delays. I'm a bit swamped with work at the moment.

In any case, if other people want to help, doing code review is always useful!

src/cache.rs
```diff
@@ -143,40 +144,65 @@ impl Cache {
     }
 
     /// Search for a page and return the path to it.
-    pub fn find_page(&self, name: &str) -> Option<PathBuf> {
-        // Build page file name
-        let page_filename = format!("{}.md", name);
-
+    pub fn find_pages(&self, name: &str, config: &Config) -> Vec<PathBuf> {
```
dbrgn, line 147: For performance and readability reasons, I think it would be good if this wouldn't return a `Vec`, but instead two separate values. Additionally, it can be made easier to read by creating a custom struct:

```rust
struct PageLookupResult {
    page_path: PathBuf,
    patch_path: Option<PathBuf>,
}
```

This way, the name of the function can be kept as `find_page`. After all, it finds a page and corresponding patch files, not multiple pages.

```rust
pub fn find_page(&self, name: &str, config: &Config) -> Option<PageLookupResult> {
```

The calling code can still use a `for path in` loop, if you implement a method that returns an iterator:

```rust
impl PageLookupResult {
    fn files(&self) -> impl Iterator<Item = &Path> {
        iter::once(self.page_path.as_path())
            .chain(self.patch_path.as_deref().into_iter())
    }
}
```

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6131f92216308d4866fecbe09c8736ef for an interactive version that includes tests.

src/cache.rs
```diff
@@ -143,40 +144,65 @@ impl Cache {
...
+        let custom_pages_directory = &config.directories.custom_pages_dir;
 
-        // Search for the page in the platform specific directory
-        if let Some(pf) = platform {
-            let path = platforms_dir.join(&pf).join(&page_filename);
-            if path.exists() && path.is_file() {
-                return Some(path);
-            }
+        // Look up custom page (<name>.page). If it exists, return it directly,
+        let custom_page = format!("{}.page", name);
```
dbrgn, line 161: Can you inline this variable? I prefer to avoid shadowing variables with different types 🙂  Simply use `.join(&format!("{}.page", name))`.

dbrgn (changes requested):
Sorry again for the long delay. I hope the next one will be quicker 🙂 I left a few comments.

Note: Some of the files were changed in a PR I merged recently, so you need to merge master into your branch and fix the conflicts.

src/config.rs
```diff
@@ -147,6 +147,31 @@ impl Default for RawUpdatesConfig {
...
+
+        Self {
+            // default to `CONFIGDIR/pages`
+            // OR `~/.local/share/tealdeer_custom_pages`
+            // OR `/.local/share/tealdeer_custom_pages`
+            custom_pages_dir: match get_config_dir() {
+                Ok(config_dir) => config_dir.join("pages"),
+                Err(_) => PathBuf::from(&home_or_root).join(".local/share/tealdeer_custom_pages"),
+            },
+        }
```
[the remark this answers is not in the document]
dbrgn: On second thought, the config dir is probably not the correct location for this. Instead, the `UserData` directory should probably be used (see https://docs.rs/app_dirs/1.2.1/app_dirs/enum.AppDataType.html).

This can be retrieved using `get_app_root(AppDataType::UserData, &crate::APP_INFO)`. On Linux, this returns `/home/user/.local/share/tealdeer`.

(Of course, you'd still need `.join("pages")`.)

src/usage.docopt
```diff
@@ -18,6 +18,8 @@ Options:
     --config-path       Show config file path
     --seed-config       Create a basic config
     --color <when>      Control when to use color [always, auto, never] [default: auto]
+    --patch             Create a .patch file that appends to a certain page.
+    --overwrite         Create a .page file that overwrites a certain page. Can be used to create new pages.
```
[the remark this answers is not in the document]
dmaahs2017: That makes a lot of sense. Best to follow KIS(S), Thanks for the feedback

src/cache.rs
```diff
@@ -143,40 +144,65 @@ impl Cache {
...
+        let custom_patch = platforms_dir
+            .join(&custom_pages_directory)
+            .join(&custom_patch);
+        let custom_patch = if custom_patch.exists() && custom_patch.is_file() {
+            Some(custom_patch)
         } else {
             None
+        };
+        dbg!(&custom_pages_directory);
+        dbg!(&custom_patch);
```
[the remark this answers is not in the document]
dmaahs2017: Ope :facepalm:

dmaahs2017:
@dbrgn I've got a couple questions now that I'm coming back to this:

**Should we allow Overwrites to have patches?**

Ex: `my_custom_page.page` and `my_custom_page.patch` If these both exist should we append the patch to the custom page? Thinking as a user, if I were to do this (even though it seems redundant) I would expect it to work. 

The downside is if we did that we would always have to check for patches first. We would then check in this order 
```
patches -> custom pages (return early) -> platform pages (return early) -> common pages (return early) -> None Found
``` 
As opposed to 
```
custom pages (return early) -> patches -> platform pages (return early) -> common pages (return early) -> None Found
``` 
I find the former to be a bit more intuitive, but I'll leave it the latter way for now.

**Secondly, I'm not sure how to handle this? Is this non-recoverable should it fail?**
```rust
    fn default() -> Self {
        Self {
            custom_pages_dir: match get_app_root(AppDataType::UserData, &crate::APP_INFO) {
                Ok(data_path) => data_path.join("pages"), ‣data_path: PathBuf
                Err(err) => todo!("What to do on fail here?"), ‣err: AppDirsError
            }
        }
    }
}
```

src/cache.rs
```diff
@@ -179,16 +231,36 @@ impl Cache {
...
+                &lang_dirs,
+            );
             if pf_path.is_some() {
                 return pf_path;
             }
         }
 
         // Did not find platform specific results, fall back to "common"
-        Self::find_page_for_platform(&page_filename, &cache_dir, "common", &lang_dirs)
+        Self::find_page_for_platform(
```
niklasmohrin, line 256: As a follow up to my last comment, I now see that the patch lookup should really not be included in that function, but rather be called directly in this function here.

Looking up patches really has nothing to do with the platform or language, so it should not be handled together. This makes it harder to see, what the changing factors are in the different calls to `Self::find_page_for_platform` (for me at least)

src/config.rs
```diff
@@ -147,6 +147,23 @@ impl Default for RawUpdatesConfig {
...
+    #[serde(default)]
+    pub custom_pages_dir: PathBuf,
+}
+
+impl Default for RawDirectoriesConfig {
+    fn default() -> Self {
+        Self {
+            custom_pages_dir: match get_app_root(AppDataType::UserData, &crate::APP_INFO) {
+                Ok(data_path) => data_path.join("pages"),
+                Err(err) => todo!("What to do on fail here?"),
```
niklasmohrin, line 161: I think there are two options here. First, you could move this code into a function `RawDirectoriesConfig::new -> Result<Self, ...>` and then have `default` call that and then call `expect` on the result (you have to panic, because you cannot return anything else). Due to the (I assume) low failure rate of `get_app_root`, this would work fine.

Otherwise, you could make `RawConfig` less active and have the field be an option (which would be more close to my understanding of a _raw_ config than the first approach). This would mean that the default  just has `None` and that some code later does the lookup of the app root (when a full-on config is _needed_ and failure should actually be handled).

I am not too sure whether one of these suggestions conflicts with some idiome applied here, but let me know if any of the two seems right to you too

src/main.rs
```diff
@@ -432,13 +432,14 @@ fn main() {
...
 
         // Search for command in cache
-        if let Some(path) = cache.find_page(&command, &languages) {
-            if let Err(msg) = print_page(&path, args.flag_markdown, &config) {
-                eprintln!("{}", msg);
-                process::exit(1);
-            } else {
-                process::exit(0);
+        if let Some(paths) = cache.find_pages(&command, &languages, &config) {
+            for path in paths.iter() {
```
niklasmohrin, line 436: This does not really tell me, that `paths.iter()` actually gives me two connected (and properly ordered) paths to a page and maybe a path, but rather seems as if multiple pages are output. This does however produce the expected output, so it is more about naming.

I suggest you call `print_page` once with the `LookupResult` and adjust this function to print everything found. To print both files in the function, you may refactor everything in it into a closure and call that on both paths contained in the `LookupResult`.

src/cache.rs
```diff
@@ -21,6 +23,30 @@ pub struct Cache {
...
+            patch_path: None,
+        }
+    }
+
+    fn with_optional_patch(mut self, patch_path: Option<PathBuf>) -> Self {
+        self.patch_path = patch_path;
+        self
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &Path> {
```
niklasmohrin, line 45: Although I really love the idea to align this with common naming, I would rather call this something like `paths_to_print`.

niklasmohrin (changes requested):
Looks good overall, thank you! I have some suggestions that mostly involve moving code around and renaming things to better fit (my perception of) what is actually happening. I also suggested something about your last comment, but I am not sure it's the best thing to do and I would want to hear what @dbrgn thinks

niklasmohrin:
Oh, and I think that if a patch and a custom page exist, both should be output, forgot to mention that. I don't think that it would be a great pain, because I would suggest looking for patches once in `find_pages` (or another outer function) and appending it to whatever page is found.

src/main.rs
niklasmohrin: An implication of this is also that I would opt to keep the name `find_page` instead of `find_pages`, because only one page (plus maybe a _patch_) would be found, right?

dbrgn:
> Oh, and I think that if a patch and a custom page exist, both should be output, forgot to mention that.

Hm, I'm not sure if I agree. I can't really think of a use case where a user would add both a custom page file *and* a patch file. I think I'd stick to rendering just the page, unless there's a compelling reason to apply the patch as well. As @dmaahs2017 wrote, the implementation with an early return for custom pages is more efficient, so I'd prefer that.

> Secondly, I'm not sure how to handle this? Is this non-recoverable should it panic?

That's a good question. Maybe the `custom_pages_dir` config could be an `Option`? If it cannot be determined, we could simply skip the "custom patch" logic.

Another option would be to propagate the error, but that's not possible in a `Default` implementation. And a panic is also not the nicest thing from a user point of view (although it might be appropriate if the app directory cannot be determined, this should not happen). @niklasmohrin what do you think?

src/cache.rs
```diff
@@ -21,6 +23,30 @@ pub struct Cache {
     os: OsType,
 }
 
+#[derive(Debug)]
```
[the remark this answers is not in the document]
dbrgn: I agree for public types in libraries, but for internally used types, I usually only derive the types that are actually used, because deriving unused types gives more work to the optimizer (=longer compilation times) and has the potential of making the binary bigger (if the optimizer isn't good enough). So in this case, I'd not add more derives. `Debug` is definitely good though, it's required for some error handling cases!

src/cache.rs
dbrgn: I agree about the more meaningful naming, but I'd stick with the simpler `paths`.

src/cache.rs
```diff
@@ -142,22 +168,45 @@ impl Cache {
         }
     }
 
+    fn find_custom_patch(name: &str, custom_dir: &Path) -> Option<PathBuf> {
+        // Look up custom patch (<name>.patch). If it exists, store it in a variable.
```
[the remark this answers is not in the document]
dbrgn: I'm generally a fan of many comments, because they help summarizing the logic below. But it could be moved to the method docs:

```
/// Look up custom patch (<name>.patch). If it exists in the file system, return the file path.
fn find_custom_patch(name: &str, custom_dir: &Path) -> Option<PathBuf> {
```

src/config.rs
dbrgn: Yeah, I agree. A panic might be fine, since `get_app_root` shouldn't usually fail, but if it doesn't complicate the code too much I think I'd go with an `Option`. If it cannot be determined, we could simply skip the "custom patch" logic.

If that makes the page lookup logic too complicated, use `expect()` / `unwrap()` instead.

dmaahs2017:
@niklasmohrin and @dbrgn love the feedback I'm learning a ton working on this 😊.

Expect me to address the new reviews by Friday evening

src/cache.rs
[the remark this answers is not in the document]
dmaahs2017: So I'm not going to derive the other traits (until needed since this is not a public library to help reduce compilation times.

src/cache.rs
dmaahs2017: Okay, so I extracted the patch lookup to a separate fn. Called that in the `find_page` function and stored it in an `Option<PathBuf>` then mapped onto the result of `find_page_for_platform`

src/cache.rs
```diff
@@ -142,22 +168,45 @@ impl Cache {
...
+        }
+    }
+
     /// Check for pages for a given platform in one of the given languages.
     fn find_page_for_platform(
-        name: &str,
+        page_name: &str,
+        patch_name: &str,
         cache_dir: &Path,
+        custom_dir: &Path,
```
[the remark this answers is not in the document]
dmaahs2017: Yeah I agree makes sense to separate this logic. I moved the call to `find_patch` into the `find_page` function.

src/config.rs
dmaahs2017: I was able to change it to an Option without too much hassle.

src/main.rs
dmaahs2017: Renamed back to `find_page` and made `iter` -> `paths`

dmaahs2017:
Ready for the next round 'o reviews

src/cache.rs
```diff
@@ -144,20 +170,39 @@ impl Cache {
 
     /// Check for pages for a given platform in one of the given languages.
     fn find_page_for_platform(
-        name: &str,
+        page_name: &str,
         cache_dir: &Path,
         platform: &str,
         language_dirs: &[String],
-    ) -> Option<PathBuf> {
+    ) -> Option<PageLookupResult> {
```
niklasmohrin, line 177: In my opinion, this now should not return a `PageLookupResult` anymore, because the result of this function will never "exhaust the full potential of the type" / will always have patch be `None` and therefore makes handling the result more complex than it has to be.

src/cache.rs
```diff
@@ -144,20 +170,39 @@ impl Cache {
...
+
+    /// Look up custom patch (<name>.patch). If it exists, store it in a variable.
+    fn find_patch(patch_name: &str, maybe_custom_dir: Option<&Path>) -> Option<PathBuf> {
+        if let Some(custom_dir) = maybe_custom_dir {
+            let patch_path = custom_dir.join(patch_name);
+            if patch_path.exists() && patch_path.is_file() {
+                return Some(patch_path);
+            }
+        }
+        None
```
niklasmohrin, line 193: This function can be converted with some Rust magic:
```rust
maybe_custom_dir
    .map(|custom_dir| custom_dir.join(patch_name))
    .filter(|path| path.exists() && path.is_file())
```

src/cache.rs
```diff
@@ -144,20 +170,39 @@ impl Cache {
...
     }
 
     /// Search for a page and return the path to it.
-    pub fn find_page(&self, name: &str, languages: &[String]) -> Option<PathBuf> {
+    pub fn find_page(
+        &self,
+        name: &str,
+        languages: &[String],
+        config: &Config,
```
niklasmohrin, line 201: This function doesn't need the full config, just the `custom_pages_dir` field, so in my opinion it should only get passed that to be more specific.

src/cache.rs
```diff
@@ -179,16 +224,29 @@ impl Cache {
...
 
-        // Try to find a platform specific path first.
+        // Get Custom Pages dir
+        let custom_dir = &config.directories.custom_pages_dir;
+
+        // Look up custom page (<name>.page). If it exists, return it directly
+        if let Some(custom_page) = custom_dir.as_ref().map(|dir| dir.join(custom_filename)) {
+            if custom_page.exists() && custom_page.is_file() {
+                return Some(PageLookupResult::with_page(custom_page));
+            }
```
niklasmohrin, line 234: You can get rid of the `as_ref` and `.map` if you first unwrap the directory with `if let Some(dir) = custom_dir { ... }` and then call `join` in the nested block.

src/cache.rs
```diff
@@ -179,16 +224,29 @@ impl Cache {
...
+        let maybe_patch = Self::find_patch(&patch_filename, custom_dir.as_deref());
+
+        // Try to find a platform specific path next, append custom patch to it.
         if let Some(pf) = self.get_platform_dir() {
-            let pf_path = Self::find_page_for_platform(&page_filename, &cache_dir, pf, &lang_dirs);
-            if pf_path.is_some() {
-                return pf_path;
+            let page = Self::find_page_for_platform(&page_filename, &cache_dir, pf, &lang_dirs);
+            if page.is_some() {
+                return page.map(|lookup_result| lookup_result.with_optional_patch(maybe_patch));
```
niklasmohrin, line 243: If the return value of `find_page_for_platform` would be `PathBuf` again, I would discard this shortcut with `is_some` and properly unpack the path and wrap it in `Some` again. This is mainly because I think it's strange to call `map` on an option that is knowingly `Some`. It could roughly look like this
```rust
let page = Self::find_page_for_platform(&page_filename, &cache_dir, pf, &lang_dirs);
if let Some(page_path) = page {
    return Some(LookupResult::with_page(page_path).with_optional_patch(maybe_patch));
}
```

niklasmohrin (changes requested):
Wow, you were a lot quicker than us 😅 I have added some small comments that should give this the final touch. Please also make sure to check CI (red cross next to your commit hash) when you push, clippy usually has some pretty nice suggestions too (it's insane what kind of stuff clippy can detect).

Other than that, great stuff! Keep it up 🚀

dbrgn:
> Please also make sure to check CI (red cross next to your commit hash) when you push, clippy usually has some pretty nice suggestions too (it's insane what kind of stuff clippy can detect).

Note: I just bumped Rust and Clippy to 1.41 on master, so to successfully pass the CI you'll need to rebase your branch (or merge master into it).

You can also run clippy locally: `cargo clean && cargo +1.41.1 clippy`.

src/cache.rs
dmaahs2017: filter on an option. I wouldn't have known that. Super cool.

dmaahs2017:
@dbrgn 
> You can also run clippy locally: `cargo clean && cargo +1.41.1 clippy`.

image
I'm not sure why, but I'm on the stable toolchain (and I merged master)

niklasmohrin (approved):
Sweet!

Yeah, I totally love these shorthand methods on `Option` and `Result` too, I think it absolutely worth it to scroll through the officials docs to see every one of them. I also saw this [reddit post](https://www.reddit.com/r/rust/comments/kdfb9k/rusts_option_in_one_figure/) a while back.

Finally, the `+1.41` syntax for `cargo` means that the `1.41` toolchain from `rustup` should be used instead of the default (probably `stable`). For that to work, you have to have the `1.41` toolchain installed though (`rustup toolchain install 1.41`). This only works if you have installed `cargo` through `rustup` (on Arch, you could install `rustc` and `cargo` alone; then this would not work). As you said, CI passes so that is a fine for this PR :)

niklasmohrin:
(note to self and @dbrgn: don't forget to squash these 16 commits)

niklasmohrin:
Hey @dmaahs2017, casual reminder that we have not forgotten your PR :+1:

@dbrgn Any chance we can merge this some time soon?

dbrgn (changes requested):
Thanks very much for your patience, @dmaahs2017 (and thanks @niklasmohrin for the reviews) 🙂 I really like how this PR turned out, especially the `PageLookupResult`!

There are only three minor nitpicks left, and I'll merge it as soon as those are fixed!

We should also include the custom pages directory in the new `--show-paths` option output, but that's probably best done in a separate PR, in order not to hold up this one.

Another issue for a new PR: Right now the patches need to be formatted like a full page, including a heading. Otherwise, the first line is interpreted as the heading, and skipped when printing (because the page and the patch are passed to separate tokenizer instances). However, I think that too can be fixed in a separate PR, I don't want to hold up this PR too long.

src/cache.rs
```diff
@@ -179,16 +218,29 @@ impl Cache {
             })
             .collect();
 
-        // Try to find a platform specific path first.
+        // Look up custom page (<name>.page). If it exists, return it directly
+        //if let Some(custom_page) = maybe_config_dir.as_ref().map(|dir| dir.join(custom_filename)) {
```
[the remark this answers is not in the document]
dmaahs2017: Removed

src/cache.rs
```diff
@@ -144,20 +169,34 @@ impl Cache {
...
+            .filter(|path| path.exists() && path.is_file())
+    }
+
     /// Search for a page and return the path to it.
-    pub fn find_page(&self, name: &str, languages: &[String]) -> Option<PathBuf> {
+    pub fn find_page(
+        &self,
+        name: &str,
+        languages: &[String],
+        maybe_config_dir: Option<&Path>,
```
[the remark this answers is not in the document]
dmaahs2017: Renamed

src/cache.rs
```diff
@@ -144,20 +169,34 @@ impl Cache {
...
     ) -> Option<PathBuf> {
         language_dirs
             .iter()
-            .map(|lang_dir| cache_dir.join(lang_dir).join(platform).join(name))
+            .map(|lang_dir| cache_dir.join(lang_dir).join(platform).join(page_name))
             .find(|path| path.exists() && path.is_file())
     }
 
+    /// Look up custom patch (<name>.patch). If it exists, store it in a variable.
+    fn find_patch(patch_name: &str, maybe_custom_dir: Option<&Path>) -> Option<PathBuf> {
```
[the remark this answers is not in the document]
dmaahs2017: Renamed

niklasmohrin (approved):
Still great! I only now noticed that there are no tests for the new behavior, but I would honestly rather have this PR merged and have the tests in another PR

dbrgn (approved):

dmaahs2017:
> Still great! I only now noticed that there are no tests for the new behavior, but I would honestly rather have this PR merged and have the tests in another PR

Oh yeah! I'll create myself an issue for that :)

PR #148, opened 2020-10-31
Add support for ANSI code and RGB colors

### LINKED ISSUE #147: Allow 'grey' color
And ideally even more, even ANSI escape codes provide a much bigger palette?

tomasfarias:
Closes #147

Allows defining colors as ANSI codes or RGB in `config.toml` e.g.:

```
[style.command_name]
background = { ansi = 7 }
foreground =  { rgb = { r = 255, g = 51, b = 255 } }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,12 +174,12 @@ Possible styles:
 
 Currently supported attributes:
 
-- `foreground` (color string, see below)
-- `background` (color string, see below)
+- `foreground` (color string, ANSI code, or RGB, see below)
+- `background` (color string, ANSI code, or RGB, see below)
 - `underline` (`true` or `false`)
 - `bold` (`true` or `false`)
 
-The currently supported colors are:
+The currently supported color strings are:
 
 - `black`
 - `red`
@@ -189,6 +189,16 @@ The currently supported colors are:
 - `purple`
 - `cyan`
 - `white`
+
+ANSI color codes are also supported:
+```
+foreground = { ansi = 4 }
+```
+
+Finally, colors can also be configured by their RGB representation:
+```
+background = { rgb = { r = 255, g = 255, b = 255 } }
+```
 
 Example customization:
 
--- a/README.md
+++ b/README.md
@@ -191,14 +191,12 @@ The currently supported color strings are:
 - `white`
 
 ANSI color codes are also supported:
-```
-foreground = { ansi = 4 }
-```
+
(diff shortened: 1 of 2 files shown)


tomasfarias:
FYI, my first attempt implemented `RGB` variant as a `tuple` `struct`:
```
pub enum RawColor {
    ...
    RGB(u8, u8, u8),
}
```

Since I expected the config to use an array, for example `foreground = { rgb = [255, 10, 125] }`. But apparently `toml` deserializing expects a table for tuple structs, not an array, so in order for deserialization to work the config would have had to look like `foreground = { rgb = { 0 = 255,  1 = 10,  2 = 125} }`.

For that reason, I went with a regular struct as I think having r, g and b as the keys is more intuitive than 0, 1 and 2: `foreground = { rgb = { r = 255,  g = 10,  b = 125} }`

dbrgn:
I actually prefer the explicit struct with key names :slightly_smiling_face:

src/config.rs
```diff
@@ -34,6 +34,8 @@ pub enum RawColor {
     Purple,
     Cyan,
     White,
+    ANSI(u8),
```
dbrgn, line 37: Rust style guides would generally suggest `Ansi` here.

```suggestion
    Ansi(u8),
```

dbrgn (changes requested):
Thanks, looks pretty straightforward!

Can you also update the README?

(Btw, README is getting a bit long, we should probably use mdbook or something like that in the future...)

tomasfarias:
@dbrgn Thanks for the review! Let me know if the README update looks fine, or if there's anything more I should add or change.

README.md
```diff
@@ -190,6 +190,16 @@ The currently supported colors are:
...
+ANSI color codes are also supported:
+```
+foreground = { ansi = 4 }
+```
+
+Finally, colors can also be configured by their RGB representation:
+```
+background = { rgb = { r = 255, g = 255, b = 255 } }
+```
+
```
dbrgn, line 202: One last nitpick, for consistency:

```suggestion
ANSI color codes are also supported:

    foreground = { ansi = 4 }

Finally, colors can also be configured by their RGB representation:

    background = { rgb = { r = 255, g = 255, b = 255 } }

```

dbrgn (changes requested):

dbrgn (approved):
Thank you!

PR #159, opened 2021-01-13
tar markdown rendering test fixed

tveness:
Build tests failing because tldr-pages repository changed file
https://github.com/tldr-pages/tldr/blob/master/pages/common/tar.md
and so test was failing, as the file the test compared to was not the version pulled from the archive.
Could attempt to generate this automatically in future?

Not very familiar with pull requests, so please let me know what would be helpful in future.


dbrgn:
Thanks! Yeah, this happens every once in a while. I wonder if we should mock the page, or switch to an alternative page that's updated very rarely. (The goal of an integration test is to test end-to-end, so not using mocks would be better I guess.)

PR #162, opened 2021-01-30
Implement new --show-dirs command

dbrgn:
The command prints out the different directories (config, cache, etc) used by tealdeer, and the reason why this directory path was chosen.

```
$ ./tldr --show-paths
Config dir:  /home/danilo/.config/tealdeer/ (OS convention)
Config path: /home/danilo/.config/tealdeer/config.toml
Cache dir:   /mnt/data/Caches/.cache/tealdeer/ (OS convention)
Pages dir:   /mnt/data/Caches/.cache/tealdeer/tldr-master/
```

```
$ TEALDEER_CONFIG_DIR=/tmp/conf ./tldr --show-paths
Config dir:  /tmp/conf/ (env variable)
Config path: /tmp/conf/config.toml
Cache dir:   /mnt/data/Caches/.cache/tealdeer/ (OS convention)
Pages dir:   /mnt/data/Caches/.cache/tealdeer/tldr-master/
```

CC @niklasmohrin

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [1.40.0, stable]
+        rust: [1.41.1, stable]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v2
@@ -45,7 +45,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-            toolchain: 1.46.0
+            toolchain: 1.41.1
             components: clippy
             override: true
       - uses: actions-rs/clippy-check@v1
--- a/bash_tealdeer
+++ b/bash_tealdeer
@@ -6,7 +6,7 @@ _tealdeer()
 	_init_completion || return
 
 	case $prev in
-		-h|--help|-v|--version|-l|--list|-u|--update|-c|--clear-cache|-p|--pager|-m|--markdown|--config-path|--seed-config|-q|--quiet)
+		-h|--help|-v|--version|-l|--list|-u|--update|-c|--clear-cache|-p|--pager|-m|--markdown|--show-paths|--seed-config|-q|--quiet)
 			return
 			;;
 		-f|--render)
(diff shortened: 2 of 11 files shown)


niklasmohrin:
I think it would be a bit risky to assume anything to be an "OS convention", because some people may have strong opinions on conventions and they might be subject to change (I, for one, always thought that the tealdeer cache is in `$HOME/.cache` aka `$XDG_CACHE_HOME`, but I stand corrected 😄 ). Maybe something more defensive like "default" would be more appropriate?

Apart from that (and the clippy suggestions), this looks good to me 👍

dbrgn:
Hm, what do you mean?The tealdeer cache *is* at `$XDG_CACHE_HOME/tealdeer`, on my computer I actually override the `$XDG_CACHE_HOME` variable. And I'd say that XDG is the de-facto Linux OS convention for directory paths.

niklasmohrin:
Ohhhh, I was just thrown off by the location of the cache in the example from the description and thought that `/mnt/data/Caches` was a convention I hadn't heard about, my bad 😄. In that case, I totally agree with what tealdeer is doing here.

What I wanted to say is, that I would not include anything that someone may disagree with (such as conventions / good practices) if it is not necessary. Also, if (for some reason) someone would use tealdeer on an OS that does not have this convention (I don't know any though), this would be conflicting with the output. This is why I would opt for the label "default" or something that explains what OS convention is meant to not display it as _the_ convention and instead make it clear, maybe "XDG" or so.

This is of course only my opinion as an external viewer and it is your decision after all. 🤷‍♂️

dbrgn:
No worries 🙂 Yeah, I agree with your general statement. I also wondered whether it would be better to write "XDG" instead of "OS convention" on Linux systems. Should be easy (I mainly feared that people might not know what XDG is). By the way, for determining "OS conventions" we use the https://docs.rs/app_dirs/ crate.

À propos "external viewer", are you interested in becoming a collaborator? I'm looking for someone to help maintaining this crate (reviewing contributions and occasionally implementing new features). The workflow would be PR based, just as before (but with someone else helping I could ask you for reviews).

Of course, feel free to decline if you don't have time or motivation 🙂

niklasmohrin:
This would be my first time maintaining software, but I guess I could try it. Since I am still stuying full time, I might not find time for a couple of days, depending on my other responsibilities. If that's okay for you, feel free to request reviews or tag me and I will try my best to help 😄

dbrgn:
Sure, you already know my response times, this is not a commercial project with deadlines 😄 

Great, I'll invite you to this repository! Reviews of pull requests are very welcome. (Branch protection will remain active, so please use pull requests and not pushes to master.)

PR #163, opened 2021-02-04
Hyperfine benchmarks

### LINKED ISSUE #129: Benchmarks with hyperfine
https://github.com/sharkdp/hyperfine

As suggested in https://github.com/dbrgn/tealdeer/issues/38#issuecomment-672343132

niklasmohrin:
Closes #129

I started putting together a Dockerfile that gets all the clients and benches them with hyperfine. I decided to do it all in Docker, because I would not want to have all the clients on my system and to make it accessible for third parties to easily verify the results that our benchmarks produce.

At the time of writing, only tealdeer and the C client are compared, but I will add more clients.

Eventually, we could think about running benchmarks in CI, but this is not the focus of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,38 +31,37 @@ High level project goals:
 
 - [x] Download and cache pages
 - [x] Don't require a network connection for anything besides updating the cache
-- [x] Command line interface similar or equivalent to the [NodeJS client][tldr-node-client]
+- [x] Command line interface similar or equivalent to the [NodeJS client][node-gh]
+- [x] Comply with the [tldr client specification][client-spec]
+- [x] Advanced highlighting and configuration
 - [x] Be fast
 
-A tool like `tldr` should be as frictionless as possible to use. It should be
-easy to invoke (just `tldr tar`, not using another subcommand like `tldr find
-tar`) and it should show the output as fast as possible.
+A tool like `tldr` should be as frictionless as possible to use and show the
+output as fast as possible.
 
-tealdeer reaches these goals. During a (highly non-scientific) test (see
-[#38](https://github.com/dbrgn/tealdeer/issues/38) for details), I tested the
-invocation speed of `tldr <command>` for a few of the existing clients:
+We think that `tealdeer` reaches these goals. We put together a (more or less)
+reproducible benchmark that compiles a handful of clients from source and
+measures the execution times on a cold disk cache. The benchmarking is run in a
+Docker container using sharkdp's [`hyperfine`][hyperfine-gh]
+([Dockerfile][benchmark-dockerfile]).
 
-| Client | Times (ms) | Avg of 5 (ms) |
-| --- | --- | --- |
-| [Tealdeer](https://github.com/dbrgn/tealdeer/) | `15/11/5/5/11` | `9.4` (100%) |
-| [C client](https://github.com/tldr-pages/tldr-cpp-client) | `11/5/12/11/15` | `10.8` (115%) |
-| [Bash client](https://github.com/pepa65/tldr-bash-client) | `15/19/22/25/24` | `21.0` (223%) |
-| [Go client by k3mist](https://github.com/k3mist/tldr/) | `98/96/100/95/101` | `98.8` (1'051%) |
-| [Python client](https://github.com/lord63/tldr.py) | `152/148/151/158/140` | `149.8` (1'594%) |
-| [NodeJS client](https://github.com/tldr-pages/tldr-node-client) | `169/171/170/170/170` | `170.0` (1'809%) |
+| Client (50 runs, 29.07.2021)      | Programming Language | Mean in ms | Deviation in ms | Comments                |
+| :---:                             | :---:                | :---:      | :---:           | :---:                   |
+| `tealdeer`                        | Rust                 | 16.2       | 1.2             |                         |
+| [`tldr-c`][c-gh]                  | C                    | 35.7       | 3.2             |                         |
+| [`tldr-hs`][hs-gh]                | Haskell              | 23.6       | 1.0             | no example highlighting |
+| [`fast-tldr`][fast-tldr-gh]       | Haskell              | 15.2       | 0.7             | no example highlighting |
+| [`tldr-bash`][bash-gh]            | Bash                 | 28.3       | 1.1             |                         |
+| [`tldr-python-client`][python-gh] | Python               | 97.9       | 3.5             |                         |
+| [`outfieldr`][outfieldr-gh]       | Zig                  | 4.7        | 0.5             | no user configuration   |
+| [`tldr-node-client`][node-gh]     | Javascript / Node    | 533.1      | 12.5            |                         |
 
-tealdeer was the winner here, although the C client and the Bash client are in
-the same speed class. Interpreted languages are clearly much slower to invoke,
-a delay of 170 milliseconds is definitely noticeable and increases friction for
-the user.
-
-These are the clients I tried but failed to compile or run:
-[Haskell client](https://github.com/psibi/tldr-hs),
-[Ruby client](https://github.com/YellowApple/tldrb),
-[Perl client](https://github.com/skaji/perl-tldr),
-[Go client by anoopengineer](https://github.com/anoopengineer/tldr/),
-[PHP client](https://github.com/BrainMaestro/tldr-php).
-
+As you can see, `tealdeer` is not the fastest of the tested clients anymore. We
+strive for useful features and code quality over raw performance, even if that
+means that we don't come out on top in this friendly competition. That said, we
+are still optimizing the code, for example when the `outfieldr` developers
+[suggested to switch][outfieldr-comment-tls] to a native TLS implementation
+instead of the native libraries.
 
 ## Development
 
@@ -113,7 +112,18 @@ be dual licensed as above, without any additional terms or conditions.
 Thanks to @severen for coming up with the name "tealdeer"!
 
 
-[tldr-node-client]: https://github.com/tldr-pages/tldr-node-client
+[node-gh]: https://github.com/tldr-pages/tldr-node-client
+[c-gh]: https://github.com/tldr-pages/tldr-c-client
+[hs-gh]: https://github.com/psibi/tldr-hs
+[fast-tldr-gh]: https://github.com/gutjuri/fast-tldr
+[bash-gh]: https://4e4.win/tldr
+[outfieldr-gh]: https://gitlab.com/ve-nt/outfieldr
+[python-gh]: https://github.com/tldr-pages/tldr-python-client
+
+[benchmark-dockerfile]: https://github.com/dbrgn/tealdeer/blob/master/benchmarks/Dockerfile
+[client-spec]: https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md
+[hyperfine-gh]: https://github.com/sharkdp/hyperfine
+[outfieldr-comment-tls]: https://github.com/dbrgn/tealdeer/issues/129#issuecomment-833596765
 
 <!-- Badges -->
 [github-actions]: https://github.com/dbrgn/tealdeer/actions?query=branch%3Amaster
--- a/benchmarks/Dockerfile
+++ b/benchmarks/Dockerfile
@@ -1,0 +1,139 @@
+# Benchmark Dockerfile for tealdeer
+#
+# To run the benchmarks, execute
+#
+#     docker build -t tldr-benchmark .
+#     docker run --privileged --rm -it tldr-benchmark
+#
+# as root in the directory of this Dockerfile. This will build the compared
+# clients and benchmark them with `hyperfine` at the end.
+#
+# The `--privileged` flag is needed to drop the disk caches before every run. If
+# you want to test with hot caches or don't want to use this flag, you will have
+# to remove the `--prepare` line from the `hyperfine` command at the end of this
+# file and rebuild the image.
+
+################################################################################
+
+FROM rust AS tealdeer-builder
+
+WORKDIR /build
+RUN git clone https://github.com/dbrgn/tealdeer.git \
+        && cd tealdeer \
+        && cargo build --release \
+        && mkdir /build-outputs \
+        && cp target/release/tldr /build-outputs/tealdeer
+
+################################################################################
+
+FROM ubuntu:latest AS tldr-c-builder
+
+WORKDIR /build
+RUN apt-get update && apt-get install -y build-essential git && rm -rf /var/lib/apt/lists/*
+RUN git clone https://github.com/tldr-pages/tldr-c-client.git \
+        && cd tldr-c-client \
+        && DEBIAN_FRONTEND=noninteractive ./deps.sh \
+        && make \
+        && mkdir /build-outputs /deps \
+        && cp tldr /build-outputs/tldr-c \
+        && cp deps.sh /deps/tldr-c-deps.sh
+
+################################################################################
+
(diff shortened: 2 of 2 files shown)


niklasmohrin:
I noticed that when using `fast-tldr` (or `tldr-hs`), the output still has the `{{` and `}}` which is not intended by the client specification. I am not sure how to compensate these differences fairly. After all, `fast-tldr` runs considerably faster than `tealdeer` in `hyperfine`.

benchmarks/Dockerfile
```diff
@@ -0,0 +1,66 @@
+FROM ubuntu:latest
+
+RUN apt update && apt install -y wget git
+RUN apt update && apt install -y build-essential
+RUN apt update && apt install -y unzip
+RUN apt update && apt install -y curl
```
dbrgn, line 6: Some notes about this:

- Only one update is necessary
- Group all dependencies into a single command, to prevent creating multiple layers
- Clean caches after installing, to reduce the image size
- Use apt-get instead of apt, it guarantees a stable command-line interface

My suggestion:

```
RUN apt-get update \
        && apt-get install -y wget git build-essential unzip curl \
        && rm -rf /var/lib/apt/lists/*
```

(Of course during development of the Dockerfile copy-paste of these lines may make sense, to re-use the cache.)

dbrgn (commented):
Ah, great idea with Docker! This way the benchmarks are reproducible. And if something (e.g. compile flags) are not fair, people can correct them.

Yeah, the features of the various clients are different. Every new feature costs a little bit performance. Tealdeer without custom formatting and no support for a config file is surely faster than the version we've arrived at by now. fast-tldr states "We strive to be the fastest tldr-client with no compromises regarding features". However, speed and features are always a compromise 🙂

I tried to test this, but right now fast-tldr complains about a missing glibc:

> ./fast-tldr: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./fast-tldr)

PS: Make sure to initialize all caches, and ensure that the clients don't clash (e.g. if they use the same config path or cache path). If possible, use a client-local cache directory, to ensure proper separation.

dbrgn:
Just so I'm up to date: @niklasmohrin this is still WIP, right? (I'll wait with another review until you think it's ready.)

niklasmohrin:
@dbrgn Yeah I am a bit out of the loop myself, I will have to come back to this with the zig build some time soon. I will just request the review once it's ready.

About the results: I think it would be fairest if we included _some_ evaluation with the results, for example that fast-tldr doesn't really parse the files and just outputs variables as they are read, otherwise we really are just breaking our own legs here

We have some performance PRs on the way, I am excited about #187. Is there any news on #108 or a way to get rid of docopt otherwise?

dbrgn:
Yeah, I agree, clients that simply stream a file to stdout aren't really fair comparison.

dbrgn:
I'm working on #108 right now. It would be kind of nice to have benchmarks before and after, but that can also be done after the PR was merged.

niklasmohrin:
Oh, before and after seems like a nice thing to have too, well then I can't rest on that argument anymore :laughing:

dbrgn:
Yeah, but as I mentioned, we can simply backport the hyperfine docker script to an old version of the repository. That's no problem 🙂

niklasmohrin:
@dbrgn I think I am satisfied with the current state of Dockerfile. I found a way to drop the disk caches from within the container, so that we can test with cold caches (which is closer to the actual use case).

Unless you have something else to add, the only thing left is getting outfieldr to build, don't know when dev will respond (only messaged earlier).

With #187 merged, we are also standing pretty strong against fast-tldr again, only falling 5-10% behind. I am wondering if we can reproduce the 30x speedup that outfieldr claim on their README.

Nevertheless, I think having all the results in a table that outlines the tradeoffs is fair to everyone

README.md
```diff
@@ -32,37 +32,38 @@ High level project goals:
...
+([Dockerfile][benchmark-dockerfile]).
+
+| Client (500 runs)                     | Mean in ms | Deviation in ms | Comments                                             |
+| :---:                                 | :---:      | :---:           | :---:                                                |
+| `tealdeer`                            | 17.6       | 1.2             |                                                      |
+| [`tldr-c`][c-gh]                      | 36.6       | 2.3             |                                                      |
+| [`tldr-hs`][hs-gh]                    | 22.7       | 2.2             | outputs raw pages, no syntax highlighting of command |
+| [`fast-tldr (Haskell)`][fast-tldr-gh] | 14.8       | 1.3             | outputs raw pages, no syntax highlighting of command |
+| [`tldr-bash`][bash-gh]                | 27.8       | 1.1             |                                                      |
+| [`outfieldr (Zig)`][outfieldr-gh]     | 7.9        | 1.0             | no syntax highlighting of command                    |
```
ve-nt, line 55: Hi, author of outfieldr here. Just wondering what is meant by "no syntax highlighting of command"?

niklasmohrin: Tealdeer can outtput command names, user variables and the rest of the example code in three completely different styles according to the user config. As far as I could tell, outfieldr only changes the text color of the whole line, so I wanted to note that here (correct me if I am wrong though)

Maybe the wording is a bit misleading, what do you propose as a fairer comment?

---

On another note: I read that zig 0.8 is out now. Should I pin the compiler to 0.8 or do you plan on using the dev version?

ve-nt: Outfieldr renders user arguments in bright/bold red, and the rest of the code (including the command name) in red. So if I understand correctly, it's lacking the feature of stylizing the command name in a different style?

---

I will be pinning Outfieldr to stable releases from now on, it'll be easier to maintain that way.

niklasmohrin: Oh sorry, I didnt see that the variables are any different. Good thing I had you look over 🙃

In that case, is it rather a "no config for syntax highlighting" / "no user config"? Or did I just not see this either? 😂

I wouldn't list the fact about the command name as its only a nice addon but not a necessity imo.

Great, Ill make sure to set the compiler image version

benchmarks/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+
+# Update all the individual caches
+RUN bash -c 'mkdir -p /caches/{tealdeer,tldr-c,tldr-hs,fast-tldr,tldr-bash,tldr-node,outfieldr/.local/share}' \
+        && TEALDEER_CACHE_DIR=/caches/tealdeer ./tealdeer -u \
+        && TLDR_CACHE_DIR=/caches/tldr-c ./tldr-c -u \
+        && XDG_DATA_HOME=/caches/tldr-hs ./tldr-hs -u \
+        && XDG_DATA_HOME=/caches/fast-tldr ./fast-tldr -u \
+        && XDG_DATA_HOME=/caches/tldr-bash ./tldr-bash -u \
+        && HOME=/caches/tldr-node ./tldr-node -u \
+        && HOME=/caches/outfieldr ./outfieldr --fetch
```
ve-nt, line 112: Outfieldr now uses the same flags as defined in the spec, so `--fetch` has been replaced with `-u`.

ve-nt (changes requested):

README.md
ve-nt: You're correct about the lack of configuration, there is none at the moment. Might be a good thing to add in the future :thinking:

niklasmohrin: Great, thank you for your cooperation! 👍

README.md
```diff
@@ -32,37 +32,38 @@ High level project goals:
...
+([Dockerfile][benchmark-dockerfile]).
+
+| Client (500 runs)                     | Mean in ms | Deviation in ms | Comments                |
+| :---:                                 | :---:      | :---:           | :---:                   |
+| `tealdeer`                            | 17.6       | 1.2             |                         |
+| [`tldr-c`][c-gh]                      | 36.6       | 2.3             |                         |
+| [`tldr-hs`][hs-gh]                    | 22.7       | 2.2             | no example highlighting |
+| [`fast-tldr (Haskell)`][fast-tldr-gh] | 14.8       | 1.3             | no example highlighting |
+| [`tldr-bash`][bash-gh]                | 27.8       | 1.1             |                         |
+| [`outfieldr (Zig)`][outfieldr-gh]     | 7.9        | 1.0             | no user configuration   |
```
dbrgn, line 55: As an improvement idea, the programming language could be listed in a separate column.

dbrgn (approved):
Really cool, thanks @niklasmohrin!

I think the Node and Python clients should be listed as well in the comparison, because both are quite commonly being used (the Python client is the one packaged in official Arch repos) and it's important to show the difference in invocation time between interpreted and compiled languages. However, feel free to merge the current state as-is, we can always improve it later on.

I'm wondering whether in the long run the benchmarking should be moved to a separate repository to simplify maintenance (the benchmarks can be interesting as a commonly-maintained reference for all tldr clients), but for now we can keep it inside the tealdeer repository as long as we don't get a "please add/update this implementation in the benchmark" request every week 🙂

Here are my numbers for the benchmarks on a recent Ryzen 5900X system:

tealdeer 14.0 ms ±   0.8 ms
tldr-c 29.5 ms ±   1.4 ms
tldr-hs 19.6 ms ±   0.6 ms
fast-tldr 14.3 ms ±   0.7 ms
tldr-bash 25.8 ms ±   0.6 ms
outfieldr 6.1 ms ±   0.5 ms
tldr-node 403.8 ms ±   9.0 ms

It's interesting that here the tealdeer numbers are better than fast-tldr, I wonder why that's the case? (Feel free to use those numbers if you want to make us "the second-fastest client on my system" 😜)

niklasmohrin:
From this perspective, having node and python in the list makes sense, yeah. I didn't want to run the node app 500 times, but I have two ideas for how to present that:
1. We list node in the table, but with an asterisk to note that the mean is only from 50 runs or
2. There are actually two columns, one for a benchmark with 500 runs and one with a benchmark with only 50 runs. Node just wouldn't appear in the 500 column.

I think I'd prefer the first option, because it's less (redundant) numbers on the screen and still presents the main point that node is an order of magnitude slower. What do you think?

---

Having this code in a separate repo means it is its own thing, whereas here it's just documentation for how numbers in a readme were generated. I would like to keep the impression of the latter for as long as possible. I was wondering if we should set up a gh action to run the benchmarks (I _think_ you can run privileged docker containers inside of the vm), but the ci runners are afaik quite the opposite of a cold machine, so I'm not sure (it's more work too ofc). It would add credibility though, since the numbers don't just come from a single person

dbrgn:
50 runs are perfectly fine. Btw, why don't we simply omit the `--runs 400` argument? Hyperfine will try to determine on its own how many runs are required to achieve statistical significance. The number of runs is not really relevant to the user, as long as the numbers are repeatable.

niklasmohrin:
Without this, it would do too few runs leading to too much variance. I also used `--min-runs` but that would just mean that everyone ran about that number of times and then I came to the conclusion that I could just fix the number of runs :man_shrugging: I'll see how reliable 50 runs are for me and check back in with you

dbrgn:
Doesn't hyperfine take the variance into account? I assumed so, but maybe I'm wrong...

niklasmohrin:
I am not sure, maybe it does to a degree, but with the small number of runs it was still somewhat unstable when I tested

dbrgn:
Ok, then let's not hold up this PR any more, we can do additional improvements later. Feel free to merge!

niklasmohrin:
(I gotta add python first though)

niklasmohrin:
@dbrgn I think everything is ready

dbrgn:
Can you build the Dockerfile? On my machine, `tldr-hs` fails every time (even with `--no-cache`).

```
Step 10/39 : RUN git clone https://github.com/psibi/tldr-hs.git         && cd tldr-hs         && stack build --install-ghc
 ---> Running in ea506c72af75
Cloning into 'tldr-hs'...
Downloading lts-18.0 build plan ...
RedownloadInvalidResponse Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/fpco/lts-haskell/master//lts-18.0.yaml"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 "/root/.stack/build-plan/lts-18.0.yaml" (Response {responseStatus = Status {statusCode = 404, statusMessage = "Not Found"}, responseVersion = HTTP/1.1, responseHeaders = [("Connection","keep-alive"),("Content-Length","14"),("Content-Security-Policy","default-src 'none'; style-src 'unsafe-inline'; sandbox"),("Strict-Transport-Security","max-age=31536000"),("X-Content-Type-Options","nosniff"),("X-Frame-Options","deny"),("X-XSS-Protection","1; mode=block"),("Content-Type","text/plain; charset=utf-8"),("X-GitHub-Request-Id","C986:10058:47268:7522A:6103AE47"),("Accept-Ranges","bytes"),("Date","Fri, 30 Jul 2021 07:46:15 GMT"),("Via","1.1 varnish"),("X-Served-By","cache-fra19120-FRA"),("X-Cache","MISS"),("X-Cache-Hits","0"),("X-Timer","S1627631175.106098,VS0,VE149"),("Vary","Authorization,Accept-Encoding"),("Access-Control-Allow-Origin","*"),("X-Fastly-Request-ID","0468e71bfb3d2f26e693cecec5a72f9a0374bb0b"),("Expires","Fri, 30 Jul 2021 07:51:15 GMT"),("Source-Age","0")], responseBody = (), responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose})
The command '/bin/sh -c git clone https://github.com/psibi/tldr-hs.git         && cd tldr-hs         && stack build --install-ghc' returned a non-zero code: 1
```

niklasmohrin:
Ah, I think it's the `haskell:8.6`, I reverted too far after trying to avoid building half the haskell ecosystem... I tested a no-cache rebuild (when I still had just `haskell`) and it worked there, I am gonna build again now before pushing :D

dbrgn:
That works, although I'd pin the version to `haskell:9.0` instead. Building works for me with that version.

dbrgn:
It's interesting how big the effect of the disk cache is! After running `sync; echo 3 | tee /proc/sys/vm/drop_caches`, a tealdeer run takes around 14 ms. For subsequent calls, it's only 2-4 ms.

Here are my numbers:

2021-07-30-132454_1164x1013_scrot

niklasmohrin:
> That works, although I'd pin the version to `haskell:9.0` instead. Building works for me with that version.

Prior to today, I had it pinned to 8.6, but the build would fail. I think that `--install-ghc` always figures out the needed version, which is why I just had it be the latest

I think running on a cold cache is most realistic, what do you think?

dbrgn:
> which is why I just had it be the latest

Ok, I don't know about that, you might be right. (I personally prefer to pin versions to avoid unexpected incompatibilities when a newer version of the image is pushed.)

> I think running on a cold cache is most realistic, what do you think?

Yes, I agree!

dbrgn (approved):
Don't forget to squash 🙂

niklasmohrin:
whoops, apparently `--autosquash` does nothing without `--interactive` :thinking: 

I think I would separate the docker and readme changes, because we may want to trace the table in the readme separately from the benchmark setup

dbrgn:
> whoops, apparently `--autosquash` does nothing without `--interactive` :thinking: 

Yep, that's the case.

> I think I would separate the docker and readme changes, because we may want to trace the table in the readme separately from the benchmark setup

Yeah, that makes sense. I was referring to the fixup commit 🙂 Feel free to merge now.

PR #164, opened 2021-02-16
Update tar-markdown.expected

niklasmohrin:
Upstream changed the tar file, updating here


dbrgn (approved):
Again? 😄 Maybe we should use a less common command for the tests.

(LGTM. Make sure to squash-merge if there's just a single commit in a PR.)

PR #168, opened 2021-03-19
zsh-completion: fix issue with [.md, add basic -L completion, remove extra _files args

### LINKED ISSUE #166: zsh completion issue due to new '[.md' page
Currently, ZSH tab-completion is erroring with:
```console
❯ tldr <tab>
_values:compvalues:11: invalid value definition: [
```

This may be due to TLDR adding a new page for ['\[.md' recently](https://github.com/tldr-pages/tldr/blob/master/pages/common/%5B.md)

Taking a look at [ZSH Parameter Expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags), one option seems to be to use the `q` option:

> q
>
> Quote characters that are special to the shell in the resulting words with backslashes; unprintable or invalid characters are quoted using the $’\NNN’ form, with separate quotes for each octet.
> 
> If this flag is given twice, the resulting words are quoted in single quotes and if it is given three times, the words are quoted in double quotes; in these forms no special handling of unprintable or invalid characters is attempted. If the flag is given four times, the words are quoted in single quotes preceded by a $. Note that in all three of these forms quoting is done unconditionally, even if this does not change the way the resulting string would be interpreted by the shell.
> 
> If a q- is given (only a single q may appear), a minimal form of single quoting is used that only quotes the string if needed to protect special characters. Typically this form gives the most readable output.
> 
> If a q+ is given, an extended form of minimal quoting is used that causes unprintable characters to be rendered using $’...’. This quoting is similar to that used by the output of values by the typeset family of commands.
> 
>

cho-m:
Changes information:
- Use array `commands=(...)` and `_describe` to deal with '[.md' & empty cache scenario.  Fixes #166
- Hide `tldr --list` stderr (`2>/dev/null`) which breaks completion with empty cache
- Remove `sed` since #112 changed commas to newlines
- Add new `sed`-equivalent replacement (`:` -> `\:`) using native [ZSH `${name//pattern/repl}`](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion) since colon is special character in ZSH completions
- Add basic support for `-L, --language` flag from #125. In future, can consider adding extra completions maybe based on caches `pages.{lang}` folders.
- Remove extraneous completion of file names for positional arguments (i.e. `'*:file:_files'`)

The following are output from ZSH 5.8 with locally built master branch.

---

Some errors shown with original completions:
1. [.md:
    error-1

2. Empty cache stderr:
    error-2

---

Output of ZSH with new completions
1. No issue with [.md:
    output1

2. Empty cache, no completions shown (ignore dimmed font due to zsh-autosuggestions plugin):
    output2

3. Empty cache, flag completions after `-`:
    output3

### DIFF
--- a/zsh_tealdeer
+++ b/zsh_tealdeer
@@ -1,7 +1,9 @@
 #compdef tldr
 
 _applications() {
-  _values 'applications' "${(uonzf)$(tldr --list | sed -e 's/, /\n/g')}"
+    local -a commands
+    commands=(${(uonzf)"$(tldr --list 2>/dev/null)"//:/\\:})
+    _describe -t commands 'command' commands
 }
 
 _tealdeer() {
@@ -18,6 +20,7 @@ _tealdeer() {
             sunos
             windows
         ))'
+        "($I -L --language)"{-L,--language}"[Override the language settings]:lang"
         "($I -u --update)"{-u,--update}"[Update the local cache]"
         "($I -c --clear-cache)"{-c,--clear-cache}"[Clear the local cache]"
         "($I -p --pager)"{-p,--pager}"[Use a pager to page output]"
@@ -31,8 +34,7 @@ _tealdeer() {
             never
         ))"
         '(- *)'{-h,--help}'[Display help]'
-            '(- *)'{-v,--version}'[Show version information]'
-            '*:file:_files'
+        '(- *)'{-v,--version}'[Show version information]'
         '1: :_applications'
     )
 


niklasmohrin (approved):
Thank you very much! 😄 Everything works as expected on my machine.

Having suggestions for the language seems like a good idea. I don't think there is a way to predict how to specify your language right now (should it be "de", or "deutsch" or "german") and completions would help with that. I don't think that it is a must-have though.

dbrgn:
Thanks @cho-m for the fix and @niklasmohrin for the review/test!

Issue #165, opened 2021-02-22
Release Using the GitHub Action


black7375:
These advantages include:
- Secret Token
- MacOS Supported
- It is automatically created at release.

black7375:
It work in process
https://github.com/black7375/tealdeer/blob/master/.github/workflows/release.yml

I'll make PR.

Now, Have upload file issue.
https://github.com/softprops/action-gh-release/issues/82

dbrgn:
Fixed in #240.

PR #170, opened 2021-03-21
Release with github workflow

black7375:
image
Issue: #165

I created a workflow that puts Mac into the release.
You don't have to expose the GitHub key.

Here's how to permanently erase files from the Git History:
```shell
git filter-branch --tree-filter 'rm -rf ./release-build.sh' HEAD
git filter-branch --prune-empty HEAD
git push origin --force master
```


.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: ["bash", "fish", "zsh"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Copy completions
+        run: |
+          cp ${{ matrix.target }}_tealdeer completions_${{ matrix.target }}
+      - uses: softprops/action-gh-release@v1 # actions/create-release, actions/upload-release-asset is deprecated
```
dbrgn, line 17: Huh, it's a bit disappointing that there's no official action for this type of thing 😕

dbrgn (commented):
Thanks @black7375 for this config! Hm, in general this looks solid, and it would be nice if we could build our releases in CI. However, for binary builds that we provide to our users as-is, I'm a bit wary of using third party actions. The thing is that we're a single "docker push" away from our next release being compromised.

@black7375 are you aware of other approaches for uploading releases to GitHub in CI? Maybe there's some kind of bash script that does the same thing, which we could check into this repository after a quick review?

black7375:
I'll check it out in the evening

black7375:
@dbrgn

https://gist.github.com/schell/2fe896953b6728cc3c5d8d5f9f3a17a3
I haven't tested it yet, but it looks good.

.github/workflows/release.yml
black7375: I was also embarrassed because it deprecated suddenly.
It's still an incomprehensible decision.

dbrgn:
That looks pretty nice and maintainable!

dbrgn:
Thanks for the update!

I'm not aware of any changes, but this seems to be an OpenSSL linking issue. OpenSSL will get thrown out by #187, so it's best if we wait until that's merged before finishing this.

dbrgn:
@black7375 PR #187 is now merged, OpenSSL is no longer a dependency. Could you rebase this PR?

black7375:
Successfully executed fetch.

dbrgn:
@black7375 what does that mean? 🙂 you did not push any changes.

black7375:
I ran a fetch updstream on GitHub.
Ah. Did you mean to rebase the commit log?

dbrgn:
Yes, it would be great if you could rebase your master branch against the upstream master branch and push the changes (you'll need `git push --force`). Then, you could push a new tag to your fork, to test whether the builds now succeed. The last few workflows all failed: https://github.com/black7375/tealdeer/actions

If you need help with rebasing, just let me know, happy to help!

black7375:
Yes. I rebase and it all passes.

I don't know why the existing action failed.
I have created a separate release-related action, which shouldn't affect the test.

dbrgn:
Replaced by #240.

Issue #169, opened 2021-03-20
`tldr --update` output is displayed by pager


tranzystorek-io:
When ~~a pager is configured, e.g. `less`,~~ `display.use_pager` is set to `true`, the output of `tldr --update` is displayed in the pager.

Info messages should be displayed as typical output, like STDERR, regardless of pager config.

dmaahs2017:
@tranzystorek-io can you include more information about your environment and the issue. I cannot replicate this by setting `export PAGER="less"`. 

However in the code we are printing the "Successfully updated cache." message to stdout. @niklasmohrin or @dbrgn should we just change this to an `eprintln!()` since it is just a progress message?

tranzystorek-io:
Sorry, i meant configuring `use_pager` in config.toml

tranzystorek-io:
@dmaahs2017  As this concerns the `use_pager` setting, your fix is ineffective - `tldr` still first checks for pager in `main` and creates a pipe to the pager.

niklasmohrin:
@tranzystorek-io You are right, the problem is that we spawn the pager process "too early". I think this part of the code could just be moved further down. Still, what @dmaahs2017 proposes is probably a good thing to do (and also part of making this problem go away).

tranzystorek-io:
I tried moving the `configure_pager` block down, but the problem still remains because the cache update logic doesn't necessarily exit the process afterwards. Maybe the `-u` flag should be treated as a "subcommand"?

niklasmohrin:
Ah, yes. I guess to stay consistent with the current behavior, the pager should be spawned in that case too. Actually, to really stay consistent, the pager should also be configured if the pages are listed. To avoid spawning it twice, maybe this block should be before the three big `if`s at the end of `main`, something like

```rust
if (args.flag_render.is_some() || args.flag_list || args.command.is_some()) && (args.flag_pager || config.display.use_pager) {
    configure_pager();
}
```

Maybe it should even stay where it is right now and just be guarded with the extra condition. What do you think?

tranzystorek-io:
We need a working fix, so either of your ideas is fine. But this shouldn't mask the fact that all the logic in the `main` function feels stitched together. Maybe once `tealdeer` switches over to clap, it could be organized better.

dbrgn:
Yeah, maybe it would make sense to finish #108 first, and then take another look at the pager logic?

I can try to give that PR another try soon.

dbrgn:
[...]

PR #171, opened 2021-03-26
Send all progress logging messages to stderr.

dmaahs2017:
Adresses #169. 

- Send logging messages to stderr instead of stdout so pagers do not intercept them.
- Updated tests to match these changes.
- Changed test_markdown_rendering to no longer rely on the tldr pages
       repo. Instead manually adding a known page to use in this test,
       because it failed due to the tldr pages repo updating this command (so
       it no longer matched our expected output, which shouldn't cause our
       tests to fail).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -175,7 +175,7 @@ fn update_cache(cache: &Cache, quietly: bool) {
         process::exit(1);
     });
     if !quietly {
-        println!("Successfully updated cache.");
+        eprintln!("Successfully updated cache.");
     }
 }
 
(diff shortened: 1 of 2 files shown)


niklasmohrin:
To keep this consistent with the rest of the program, messages like "Successfully deleted cache." should also be printed on stderr. Maybe even the "page not found" output, but I am not sure about that one. (@dbrgn ?)

@dmaahs2017 As you can see, the tests have to be updated accordingly, but that should be a quick fix

dbrgn (changes requested):
This is a good thing to do (no matter how we handle the paging issue). However, the tests need an update (expect the string in stderr, not stdout).

@dmaahs2017 can you update the tests as well?

dmaahs2017:
Ready for re-review. @niklasmohrin @dbrgn

dbrgn (approved):
Thanks, LGTM!

PR #173, opened 2021-04-07
Fix clippy warnings

SimplyDanny:


src/types.rs
```diff
@@ -108,8 +108,6 @@ pub enum PathSource {
     OsConvention,
     /// Env variable (TEALDEER_*)
     EnvVar,
-    /// Config file variable
-    ConfigVar,
```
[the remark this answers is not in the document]
SimplyDanny: I see. Shall we add `#[allow(dead_code)]` to the variant instead?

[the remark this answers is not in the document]
dbrgn: It's probably not necessary, the warnings don't cause CI to fail and we'll hopefully land #141 before the next release.

src/main.rs
```diff
@@ -224,11 +224,13 @@ fn show_paths() {
...
-        .map(|mut path| {
-            path.push(""); // Trailing path separator
-            path.to_str().unwrap_or("[Invalid]").to_string()
-        })
-        .unwrap_or_else(|e| format!("[Error: {}]", e));
+        .map_or_else(
+            |e| format!("[Error: {}]", e),
+            |mut path| {
+                path.push(""); // Trailing path separator
+                path.to_str().unwrap_or("[Invalid]").to_string()
```
niklasmohrin, line 231: I think this should be changed further. The two calls to `map` are a bit weird and there are also two allocations that could be removed (`push` instead of `join` does not throw away the old buffer and converting the `PathBuf` to a `String` by using the inner `OsString`):

```rust
let pages_dir = Cache::get_cache_dir().map_or_else(
    |e| format!("[Error: {}]", e),
    |(mut path, _)| {
        path.push("tldr-master");
        path.push(""); // Trailing path separator
        path.into_os_string()
            .into_string()
            .unwrap_or_else(|_| String::from("[Invalid]"))
    },
);
```

src/main.rs
```diff
@@ -298,7 +300,7 @@ fn get_os() -> OsType {
     OsType::Other
 }
 
-fn get_languages(env_lang: Option<String>, env_language: Option<String>) -> Vec<String> {
```
niklasmohrin, line 301: Is there any chance this could be a `Option<&str>` instead?

src/main.rs
niklasmohrin: Similar changes can be made to the calls above, but I don't think they are mandatory.

niklasmohrin (changes requested):
Gotta love those clippy suggestions

src/main.rs
niklasmohrin: Actually, both can be `Option<&str>`:

```rust
fn get_languages(env_lang: Option<&str>, env_language: Option<&str>) -> Vec<String> {
    // Language list according to
    // https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language

    if env_lang.is_none() {
        return vec!["en".to_string()];
    }
    let env_lang = env_lang.unwrap();

    // Create an iterator that contains $LANGUAGE (':' separated list) followed by $LANG (single language)
    let locales = env_language
        .unwrap_or("")
        .split(':')
        .chain(iter::once(env_lang));

    let mut lang_list = Vec::new();
    for locale in locales {
        // Language plus country code (e.g. `en_US`)
        if locale.len() >= 5 && locale.chars().nth(2) == Some('_') {
            lang_list.push(&locale[..5]);
        }
        // Language code only (e.g. `en`)
        if locale.len() >= 2 && locale != "POSIX" {
            lang_list.push(&locale[..2]);
        }
    }

    lang_list.push("en");
    lang_list.clear_duplicates();

    lang_list.into_iter().map(str::to_string).collect()
}
```

Sorry for dictating so much, I just have to point out that old mistake (that I contributed 😄)

Although the tests still work, I think all of the `.into()` calls can be removed now. Also, there could be a `fn get_languages_from_env() -> Vec<String>` that handles the env var retrieval, so that this logic is moved out of `main` and shortened. What do you think? (feel free to leave that to someone else, if you just want to fix the clippy warnings 😄)

src/main.rs
SimplyDanny: Applied. Thank you for your detailed description.

src/main.rs
SimplyDanny: I changed the function as you suggested and removed all the `.into()`s. Please have a look at the new `get_languages_from_env` function. Does it match to what you imagined?

niklasmohrin: Oh, I didn't see your comment here 😆 Yes, almost. I started a new review below, but it's only small details. Thank you for contributing!

src/main.rs
```diff
@@ -298,42 +301,48 @@ fn get_os() -> OsType {
...
+fn get_languages_from_env(language: Option<String>) -> Vec<String> {
+    language.map_or_else(
+        || {
+            get_languages(
+                std::env::var("LANG").ok().as_deref(),
+                std::env::var("LANGUAGE").ok().as_deref(),
+            )
+        },
+        |lang| vec![lang],
+    )
```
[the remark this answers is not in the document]
SimplyDanny: I actually like the first fix. Separation of env and flags is cleaner and makes total sense. Thanks again for your time!

dbrgn (approved):
LGTM, thanks!

PR #174, opened 2021-04-10
Create cache directory path if it does not exist

### LINKED ISSUE #167: `tldr --clear-cache` with $TEALDEER_CACHE_DIR requires manual mkdir to rebuild cache
`tldr --clear-cache` deletes the `$TEALDEER_CACHE_DIR` folder but `tldr --update` is unable to remake the folder.

```console
❯ tldr --list | head -n 2
2to3
7z

~
❯ tldr --clear-cache
Successfully deleted cache.
Could not update cache: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.

~
❯ tldr --list | head -n 2
Could not update cache: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.

~
❯ tldr --update
Could not update cache: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.

~
❯ mkdir -p "$TEALDEER_CACHE_DIR"

~
❯ tldr --update
Successfully updated cache.
```

Also, error messages in `--show-paths`:
```console
❯ ./target/debug/tldr --show-paths
Config dir:  /Users/cho-m/.config/tealdeer/ (env variable)
Config path: /Users/cho-m/.config/tealdeer/config.toml
Cache dir:   [Error: CacheError: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.]
Pages dir:   [Error: CacheError: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.]
Could not update cache: Path specified by $TEALDEER_CACHE_DIR does not exist or is not a directory.
```

SimplyDanny:
This PR tries to resolve #167. If `TEALDEER_CACHE_DIR` is set but does not exist, the complete path is just created as suggested in the issue. The user will be informed about the creation.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -64,8 +64,10 @@ impl Cache {
         // Allow overriding the cache directory by setting the env variable.
         if let Ok(value) = env::var(CACHE_DIR_ENV_VAR) {
             let path = PathBuf::from(value);
-            let path_exists = path.exists();
-            if path_exists && !path.is_dir() {
+            let (path_exists, path_is_dir) = path
+                .metadata()
+                .map_or((false, false), |md| (true, md.is_dir()));
+            if path_exists && !path_is_dir {
                 return Err(CacheError(format!(
                     "Path specified by ${} is not a directory.",
                     CACHE_DIR_ENV_VAR
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -64,26 +64,26 @@ impl Cache {
         // Allow overriding the cache directory by setting the env variable.
         if let Ok(value) = env::var(CACHE_DIR_ENV_VAR) {
             let path = PathBuf::from(value);
-            if path.is_dir() {
-                return Ok((path, PathSource::EnvVar));
-            }
-            if path.exists() {
+            let path_exists = path.exists();
+            if path_exists && !path.is_dir() {
                 return Err(CacheError(format!(
                     "Path specified by ${} is not a directory.",
                     CACHE_DIR_ENV_VAR
                 )));
             }
-            // Try to create the complete directory path.
-            fs::create_dir_all(&path).map_err(|_| {
-                CacheError(format!(
-                    "Directory path specified by ${} cannot be created.",
-                    CACHE_DIR_ENV_VAR
-                ))
-            })?;
-            eprintln!(
-                "Successfully created cache directory path `{}`.",
-                path.to_str().unwrap()
-            );
+            if !path_exists {
+                // Try to create the complete directory path.
+                fs::create_dir_all(&path).map_err(|_| {
+                    CacheError(format!(
+                        "Directory path specified by ${} cannot be created.",
+                        CACHE_DIR_ENV_VAR
+                    ))
+                })?;
+                eprintln!(
+                    "Successfully created cache directory path `{}`.",
+                    path.to_str().unwrap()
+                );
+            }
             return Ok((path, PathSource::EnvVar));
         };
 
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -16,6 +16,8 @@ use walkdir::{DirEntry, WalkDir};
 use crate::error::TealdeerError::{self, CacheError, UpdateError};
 use crate::types::{OsType, PathSource};
 
+static CACHE_DIR_ENV_VAR: &str = "TEALDEER_CACHE_DIR";
+
 #[derive(Debug)]
 pub struct Cache {
     url: String,
@@ -59,19 +61,30 @@ impl Cache {
 
     /// Return the path to the cache directory.
     pub fn get_cache_dir() -> Result<(PathBuf, PathSource), TealdeerError> {
-        // Allow overriding the cache directory by setting the
-        // $TEALDEER_CACHE_DIR env variable.
-        if let Ok(value) = env::var("TEALDEER_CACHE_DIR") {
+        // Allow overriding the cache directory by setting the env variable.
+        if let Ok(value) = env::var(CACHE_DIR_ENV_VAR) {
             let path = PathBuf::from(value);
-
-            if path.exists() && path.is_dir() {
+            if path.is_dir() {
                 return Ok((path, PathSource::EnvVar));
             }
-            return Err(CacheError(
-                "Path specified by $TEALDEER_CACHE_DIR \
-                     does not exist or is not a directory."
-                    .into(),
-            ));
+            if path.exists() {
+                return Err(CacheError(format!(
+                    "Path specified by ${} is not a directory.",
+                    CACHE_DIR_ENV_VAR
+                )));
+            }
+            // Try to create the complete directory path.
+            fs::create_dir_all(&path).map_err(|_| {
+                CacheError(format!(
+                    "Directory path specified by ${} cannot be created.",
+                    CACHE_DIR_ENV_VAR
+                ))
+            })?;
+            eprintln!(
+                "Successfully created cache directory path `{}`.",
+                path.to_str().unwrap()
+            );
+            return Ok((path, PathSource::EnvVar));
         };
 
         // Otherwise, fall back to user cache directory.
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -10,6 +10,9 @@ use predicates::boolean::PredicateBooleanExt;
 use predicates::prelude::predicate::str::{contains, is_empty, similar};
 use tempfile::{Builder, TempDir};
 
+// TODO: Should be 'cache::CACHE_DIR_ENV_VAR'. This requires to have a library crate for the logic.
+static CACHE_DIR_ENV_VAR: &str = "TEALDEER_CACHE_DIR";
+
 struct TestEnv {
     pub cache_dir: TempDir,
     pub custom_pages_dir: TempDir,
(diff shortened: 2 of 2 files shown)


dmaahs2017:
@SimplyDanny Sorry about the conflicts my PR just went through recently for custom pages and patches🙃

src/cache.rs
```diff
@@ -16,6 +16,8 @@ use walkdir::{DirEntry, WalkDir};
 use crate::error::TealdeerError::{self, CacheError, UpdateError};
 use crate::types::{OsType, PathSource};
 
+static CACHE_DIR: &str = "TEALDEER_CACHE_DIR";
```
niklasmohrin, line 19: I feel like this could be more descriptive, maybe something like `CACHE_DIR_ENV_VAR`?

niklasmohrin: I like that you moved it to constant though! 👍

src/cache.rs
```diff
@@ -61,17 +63,31 @@ impl Cache {
...
+                CacheError(format!(
+                    "Directory path specified by ${} cannot be created.",
+                    CACHE_DIR
+                ))
+            })?;
+            println!(
+                "Successfully created cache directory path `{}`.",
+                path.to_str().unwrap()
+            );
+            return Ok((path, PathSource::EnvVar));
```
niklasmohrin, line 90: This should work as expected, but I think this can be reordered to better fit the control flow: First, we check if the path already exists and is not a directory and if so, we reject it. If not, we know that we want the directory at that path. So now, we check again if it exists and if it doesn't, we create it. Finally, we return the path and path source:

```rust
if path.is_dir() {
  // error
}

if !path.exists() {
  // create dir   <-- your PR contributes this logic
}

return Ok(...);
```

I also saw that [`Path::is_dir`](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_dir) also checks that the path exists, so doing `Path::exists` aswell is redundant.

tests/lib.rs
```diff
@@ -10,6 +10,8 @@ use predicates::boolean::PredicateBooleanExt;
 use predicates::prelude::predicate::str::{contains, is_empty, similar};
 use tempfile::{Builder, TempDir};
 
+static CACHE_DIR: &str = "TEALDEER_CACHE_DIR";
```
niklasmohrin, line 13: Same here for the name

tests/lib.rs
```diff
@@ -188,6 +187,50 @@ fn test_quiet_old_cache() {
...
+        .arg("-u")
+        .assert()
+        .success()
+        .stdout(contains(format!(
+            "Successfully created cache directory path `{}`.",
+            internal_cache_dir.to_str().unwrap()
+        )))
+        .stdout(contains("Successfully updated cache."));
+
+    assert!(internal_cache_dir.exists());
```
niklasmohrin, line 211:
```suggestion
    assert!(internal_cache_dir.is_dir());
```

Get another check for free

niklasmohrin (changes requested):
Thank you for contributing :rocket:! Everything looks correct, I only have some suggestions regarding code structure and naming. That said, this is of cource open for discussion :)

src/cache.rs
SimplyDanny: Renamed.

tests/lib.rs
SimplyDanny: Renamed. Also added a TODO that this can be taken from `cache::CACHE_DIR_ENV_VAR` some day.

tests/lib.rs
SimplyDanny: Done.

src/cache.rs
SimplyDanny: Good to know that `Path::is_dir` has an existence check included. 👍  With that knowledge, I changed it to what you suggested with one exception, though, namely the second if-condition is not negated. Its block returns the error while the code after does the creation stuff.

tests/lib.rs
niklasmohrin: Good point 👍

src/cache.rs
niklasmohrin: Oops, of course we should not error when there is already a directory set up 😆 

I stiill think that the control flow should be changed to only return early on errors and have a single succesful return at the end:

```rust
if path.exists() && !path.is_dir() { return Err(...); }
if !path.exists() { create_and_maybe_error(); }

return Ok(...);
```

I think this is more easy to read, because it better reflects the dialog / contract between us and the users: You (the user) want to have something in place `x`. If place `x` cannot be used, we error. Otherwise, you are good to go. If the directory is not there yet, we set it up for you. Finally we have nothing left to complain and return.

Note how this also moves the directory creation code out of the main code block. Right now, to understand when a directory is created, you would have to understand from context that we would return early in case that it already exists. But in the above structure, the code is right by the condition.

Feel free to argue against me here though, maybe I am missing something 

(and feel free to ask about anything else too, I wrote this kind of quickly because github decided to delete all of  my test when I clicked on the unresolve-button)

SimplyDanny: Control flow might be clearer, but only if you really think about. It requires more complex (negated) conditions, which results in more work for a reader not already familiar with the logic. In the current state, the conditions and the contents of their associated blocks are very simple: 

1. If it's a directory return it.
2. If it exists (but is clearly not a directory) throw an error.
3. Otherwise create and return it.

This is certainly just personal preference and I am open to change it, if I could not convince you. 😉

niklasmohrin: :man_shrugging: To be frank, I am not convinced :laughing: 

@dbrgn Can you play the tiebreaker here?

niklasmohrin:
@dbrgn Any chance you can take a quick look at https://github.com/dbrgn/tealdeer/pull/174#discussion_r619826554 so this can move forward? :^)

src/cache.rs
dbrgn: I tend to agree with @niklasmohrin. When I quickly glanced at the code, I first read the `if path.exists() { return Err( ...` line and was a bit confused by it ("why do we return an error if the path exists?"). Only afterwards I saw that it was preceded by an `is_dir` call.

However, in order to avoid multiple repeated checks that need information from filesystem, you could extract the two checks:

```
let exists = path.exists();
let is_dir = path.is_dir();

if exists && !is_dir { /* error */ }
if !exists { /* create and maybe error */ }
// otherwise, success
```

To me, this seems like a logical order: Ensure that no problems exist, and if everything is validated, return the result.

niklasmohrin: Note: this still does two syscalls, which could just be one with https://doc.rust-lang.org/std/path/struct.Path.html#method.metadata

```rust
let metadata = path.metadata();
let exists = metadata.is_ok();
let is_dir = matches!(metadata, Some(md) if md.is_dir());
// or
let is_dir = exists && metadata.unwrap().is_dir();
```
(this is also how these methods are implemented on `Path`)

dbrgn: Good point! (I'd prefer the version without `matches!`, that one confused me initially 😄 It's clever but not easy to read.)

SimplyDanny: Done.

src/cache.rs
```diff
@@ -61,17 +63,29 @@ impl Cache {
     pub fn get_cache_dir() -> Result<(PathBuf, PathSource), TealdeerError> {
         // Allow overriding the cache directory by setting the
         // $TEALDEER_CACHE_DIR env variable.
```
niklasmohrin, line 65: Since the variable name is now a constant, we should remove other occurences:

```suggestion
        // env variable.
```

I am not sure if it appears in other places too

src/cache.rs
```diff
@@ -61,17 +63,29 @@ impl Cache {
...
-        if let Ok(value) = env::var("TEALDEER_CACHE_DIR") {
+        if let Ok(value) = env::var(CACHE_DIR_ENV_VAR) {
             let path = PathBuf::from(value);
-
-            if path.exists() && path.is_dir() {
-                return Ok((path, PathSource::EnvVar));
+            let path_exists = path.exists();
+            if path_exists && !path.is_dir() {
```
niklasmohrin, line 69: This could still be reduced to one syscall (see the thread from the last review):
```suggestion
            let metadata = path.metadata();
            let path_exists = metadata.is_ok();
            let path_is_dir = path_exists && metadata.unwrap().is_dir();
            if path_exists && !path_is_dir {
```

niklasmohrin (changes requested):
only two nitpicks left from me 👍

src/cache.rs
SimplyDanny: It's only this one occurrence - removed it.

src/cache.rs
SimplyDanny: I did not understand the last comment correctly. Sorry for that! Now it should look like you proposed.

However, the `unwrap()` in line 70 is only justified because of applied short-circuit evaluation. What do you think about being a bit more explicit by writing `let path_is_dir = metadata.map_or(false, |md| md.is_dir());`?

niklasmohrin: No worries! I agree, the `unwrap` stands out a bit, I didn't really like the idea of checking whether the metadata is `Some` explicitly twice. Your proposed solution is probably still better than the `unwrap`. I played around with that though a bit, what do you think of

```rust
let (exists, is_dir) = path.metadata().map_or((false, false), |md| (true, md.is_dir()));
```

Pick whichever of the two solutions you like best. @dbrgn and I will look over it once more and I believe we can then merge this :tada:

SimplyDanny: I like the idea with the tuple. 👍

niklasmohrin (approved):

dbrgn (approved):
Looks great now, thank you @SimplyDanny (and @niklasmohrin for the reviews)!

PR #175, opened 2021-04-10
Remove 'extern crate' declarations

SimplyDanny:
Cargo.toml specifies Rust Edition 2018 which does not require 'extern crate' anymore.


dbrgn:
Thanks, I guess that was an oversight!

PR #176, opened 2021-04-14
Fix #82: Allow SunOS override

### LINKED ISSUE #82: OS Override / SunOS issue
I am running FreeBSD but trying to use `tldr --os sunos truss` to view the page for SunOS's truss which exists in tldr's cache and am unable to. I am getting: 

```bash
Page truss not found in cache
```

Looks like this is due to [this line](https://github.com/dbrgn/tealdeer/blob/74c37193e303fb13302c6829b086cf84e2a9b817/src/cache.rs#L136) in `cache.rs`. I can replace the None with Some('sunos') and it lets me see the page as you would expect. I see you have a comment there but not sure how it is relevant.

SimplyDanny:
The issue #82 is quite old, even though it already mentions the line to fix. Or is there something more to do here that I am missing?

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -131,12 +131,11 @@ impl Cache {
     }
 
     /// Return the platform directory.
-    #[allow(clippy::match_same_arms)]
     fn get_platform_dir(&self) -> Option<&'static str> {
         match self.os {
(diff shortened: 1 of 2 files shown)


dbrgn (approved):
LGTM, thanks!

Issue #182, opened 2021-04-22
Include custom pages and Patches directory in `--show-paths` option


dmaahs2017:
Quote from PR #142
> We should also include the custom pages directory in the new --show-paths option output, but that's probably best done in a separate PR, in order not to hold up this one.

Acceptance Criteria:
* Using `--show-paths` includes the location of the custom_pages directory in the output
* Tests are created / updated for new and changed code

Issue #180, opened 2021-04-22
Create tests for Custom Pages and Patches Functionality


dmaahs2017:
Create Integration and Unit Tests for functiality added in PR #142.

Address issue: #180

dbrgn:
@dmaahs2017 from your comment in #142 it seems you're willing to implement these tests, right? I assigned the issue to you, otherwise just let me know 🙂

dmaahs2017:
> @dmaahs2017 from your comment in #142 it seems you're willing to implement these tests, right? I assigned the issue to you, otherwise just let me know 🙂 

Yep, thanks!

dmaahs2017:
Closing this because PR #183 was merged

dmaahs2017:
@niklasmohrin @dbrgn can we close this issue? or is there a way to mark it as complete as opposed to just closing it?

dbrgn:
No, closing is fine!

niklasmohrin:
@dmaahs2017 A note on that: When your initial PR comment has something like "fixes #xxx", the issue will be closed automatically when the PR is merged

https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

dmaahs2017:
> A note on that: When your initial PR comment has something like "fixes #xxx", the issue will be closed automatically when the PR is merged

Oh that's cool, Today I learned 🙂

PR #183, opened 2021-04-22
Tests for PR #142

dmaahs2017:
PR for Issue: #180 

Adds tests that should have been added in PR #182 

Added
- 3 integration tests for custom pages, including a 2 new markdown files for testing
- renamed my poorly named variable when I came across it again
- Unit tests for PageLookupResult
- Some new helper functions for the TestEnv struct for integration tests

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -224,20 +224,20 @@ impl Cache {
             }
         }
 
-        let maybe_patch = Self::find_patch(&patch_filename, custom_pages_dir.as_deref());
+        let patch_path = Self::find_patch(&patch_filename, custom_pages_dir.as_deref());
 
         // Try to find a platform specific path next, append custom patch to it.
         if let Some(pf) = self.get_platform_dir() {
             if let Some(page) =
                 Self::find_page_for_platform(&page_filename, &cache_dir, pf, &lang_dirs)
             {
-                return Some(PageLookupResult::with_page(page).with_optional_patch(maybe_patch));
+                return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
             }
         }
 
         // Did not find platform specific results, fall back to "common"
         Self::find_page_for_platform(&page_filename, &cache_dir, "common", &lang_dirs)
-            .map(|page| PageLookupResult::with_page(page).with_optional_patch(maybe_patch))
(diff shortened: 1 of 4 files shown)


dmaahs2017:
Tests @dbrgn or @niklasmohrin ready for review :)

threema-danilo (commented):
_(...Edit: I should separate work- and after-work-browsers more clearly... The side effects of home office.)_

dbrgn (commented):
Thanks, looks great!

Can you run rustfmt, so that CI passes?

src/cache.rs
```diff
@@ -314,3 +314,27 @@ impl Cache {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_page_lookup_result_iter_with_patch() {
+        let lookup = PageLookupResult::with_page(PathBuf::from("test.page"))
+            .with_optional_patch(Some( PathBuf::from("test.patch") ));
+        let mut iter = lookup.paths();
+        assert_eq!(iter.next(), Some(PathBuf::from("test.page").as_path()));
+        assert_eq!(iter.next(), Some(PathBuf::from("test.patch").as_path()));
```
niklasmohrin, line 329: can't there just be created as `Path`s directly?

niklasmohrin (commented):
Looks good, I only have a small detail :+1:

src/cache.rs
dmaahs2017: Do you mean like this?:
```rust
        let lookup = PageLookupResult {
            page_path: PathBuf::from("test.page"),
            patch_path: PathBuf::from("test.path"),
        };
```

threema-danilo: I think @niklasmohrin meant this:

    assert_eq!(iter.next(), Some(Path::from("test.page")));

I haven't tested whether this works. But it should probably work, it's a Path reference with 'static lifetime.

dbrgn: I think @niklasmohrin meant this:

```rust
assert_eq!(iter.next(), Some(Path::from("test.page")));
```

I haven't tested whether this works. But it should probably work, it's a Path reference with 'static lifetime.

dmaahs2017: Oh yeah `Path::new(...)` is cleaner :). Changed the test

`Path::from(..)` didn't compile when I first tried it because of something about the `sized` trait.

PR #184, opened 2021-04-22
Include custom pages directory in --show-paths command

dmaahs2017:
Adds custom pages to output of `--show-paths`. See issue: #182

If [directories.custom_page_dir] is not set then it will show the default location as defined by the `DirectoriesConfig` `RawDirectoriesConfig` structs
image


src/config.rs
```diff
@@ -212,7 +212,7 @@ pub struct UpdatesConfig {
     pub auto_update_interval: Duration,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Default, Debug, PartialEq)]
```
dmaahs2017, line 215: @niklasmohrin I read another PR where you mentioned not to derive Default on Config structs, to prefer manual implementations in order to choose meaningful defaults. This was already done for the `RawDirectoriesConfig` so I figured it was fine. Thoughts?

niklasmohrin: In general, I think that the same reasons apply here. Here, the derived value (`None`) makes sense, but I think that it would be better to have a manual implementation here too, in case other fields with non-trivial defaults get added. It's more subtle here though :thinking: 

@dbrgn I'm just gonna pass this one to you

src/main.rs
```diff
@@ -235,10 +235,20 @@ fn show_paths() {
...
     );
-    println!("Config dir:  {}", config_dir);
-    println!("Config path: {}", config_path);
-    println!("Cache dir:   {}", cache_dir);
-    println!("Pages dir:   {}", pages_dir);
+
+    // This first unwrap will never panic because RawDirectoriesConfig implements Default. The only
+    // reason this is Option is if get_app_root() fails.
+    let custom_pages_dir = custom_pages_dir
+        .unwrap()
```
dmaahs2017, line 242: This is guarenteed to be safe. Kinda counter-intuitive. It is only Option because of how Default for the RawDirectoryConfig is implemented with get_app_root. But since there is a default it is safe to unwrap.

tests/lib.rs
```diff
@@ -200,21 +214,55 @@ fn test_setup_seed_config() {
...
+                AppDataType::UserData,
+                &AppInfo {
+                    name: "tealdeer",
+                    author: "tealdeer"
+                }
+            )
+            .map(|path| path.join("pages"))
+            .unwrap()
+            .to_str()
+            .unwrap()
```
dmaahs2017, line 241: Just replicates how the Default is defined for custom_pages_dir in `RawDirectoriesConfig`

src/config.rs
dmaahs2017: @niklasmohrin @dbrgn I changed it back. I was using `DirectoriesConfig` it in a weird way. I reasoned that I can safely unwrap the custom_pages_dir. See https://github.com/dbrgn/tealdeer/pull/184#discussion_r618709078

src/main.rs
```diff
@@ -237,22 +235,15 @@ fn show_paths(custom_pages_dir: Option<impl AsRef<Path>>) {
...
-    };
+
+    // This first unwrap will never panic because RawDirectoriesConfig implements Default. The only
+    // reason this is Option is if get_app_root() fails.
+    let custom_pages_dir = custom_pages_dir
+        .unwrap()
+        .as_ref()
+        .to_str()
+        .map(ToString::to_string)
+        .unwrap_or_else(|| String::from("[Invalid]"));
```
niklasmohrin, line 246: Although I think I may have proposed this some time ago, you should be able to just use a `&str` here. I don't think the `unwrap` is proper here, if this variable would really always be `Some`, the argument shouldn't be an option. Some analysis is needed here

---

I am a bit confused by this function again (also outside of your diff). It feels a bit like a lot of similar pieces are coming together here, but all the information is gathered in different ways. I am not up to speed on how config / argument / env var merging is handled in tealdeer, but I feel like this should really be a method on some big instance of `Configuration` (like a singleton, but not global?) or similar that knows all the final settings. This has nothing to do with your diff though, I'm just spiraling :laughing:

dmaahs2017: It was because I had to use get_app_root in the Default for RawDirectoriesConfig and I had I discussion with @dbrgn about whether or not the RawDirectoriesConfig::default() should panic if we unwrapped get_app_root. 

I'll try to find that discussion

tests/lib.rs
niklasmohrin: Two notes on this:

1. If it is duplicated, maybe it should be a function?
2. Why is there a `unwrap` after a `map`? The `unwrap` could be moved up and then there wouldn't have to be a closure. You could also make it an `expect` with a message that just says that `get_app_root` is broken or so (which shouldn't happen, but would be nicer)

src/main.rs
dmaahs2017: It appears that I had made the decision to use option, I think I'm going to go back and change it now that I realize how unintuitive this is.

niklasmohrin: Well, yes. But the `show_paths` function requires the option to be `Some`, so it could just take the value. The caller would then have the `unwrap` or even better, an `expect`. Am I missing something?

niklasmohrin: No worries :+1:

tests/lib.rs
dmaahs2017: This is just the test which is why

niklasmohrin: Oh, I didn't mean to hardcode it, I meant using `expect` instead of `unwrap` to provide a more useful panic message:

```rust
get_app_root(
    AppDataType::UserData,
    &AppInfo {
        name: "tealdeer",
        author: "tealdeer"
    }
)
.expect("get_app_root failed, this should never happen...")
.join("pages")
.to_str()
.expect("path returned from get_app_root was not valid UTF-8, this shouldd never happen...")
```

About having it in a function: I that there are details (like the arguments of `get_app_root`) which don't really belong into this test case (and would have to be replaced if these would ever change). It's not a must though

src/config.rs
```diff
@@ -151,15 +151,18 @@ impl Default for RawUpdatesConfig {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 struct RawDirectoriesConfig {
     #[serde(default)]
-    pub custom_pages_dir: Option<PathBuf>,
+    pub custom_pages_dir: PathBuf,
```
niklasmohrin, line 154: I am not really sure if my understanding is correct here, but I thought that all the `Raw*Config`s are just "dumb" data containers, that are "materialized" into meaningful values (with smart defaults and all) when converting to their `*Config` counterparts. Then again, this happened here before :man_shrugging: 

@dbrgn why exactly do we have this distinction again?

src/config.rs
```diff
@@ -151,15 +151,18 @@ impl Default for RawUpdatesConfig {
...
 impl Default for RawDirectoriesConfig {
     fn default() -> Self {
         Self {
             custom_pages_dir: get_app_root(AppDataType::UserData, &crate::APP_INFO)
                 .map(|path| path.join("pages"))
-                .ok(),
+                .unwrap_or_else(|_| {
+                    eprintln!("Could not locate the App Root");
+                    std::process::exit(1);
+                }),
```
niklasmohrin, line 165: if this stays in, the unwrap should be moved up:

```rust
custom_pages_dir: get_app_root(AppDataType::UserData, &crate::APP_INFO)
    .unwrap_or_else(|_| {
        eprintln!("Could not locate the App Root");
        std::process::exit(1);
    })
    .join("pages"),
```

niklasmohrin (changes requested):

tests/lib.rs
dmaahs2017: Oh i see, yeah, I kinda did the map then the unwraps.. which is kind of a strange order, and wasn't quite clear what I was unwrapping. Changed it to expect() calls like how you showed.

src/config.rs
dmaahs2017: Yeah I'm not sure either tbh. Kinda reminds me of the DTO pattern, since `RawConfig` is the one that's serializable and `Config` is not.

src/config.rs
dmaahs2017: Fixed

src/cache.rs
```diff
@@ -179,19 +179,22 @@ impl Cache {
...
-    fn find_patch(patch_name: &str, custom_pages_dir: Option<&Path>) -> Option<PathBuf> {
-        custom_pages_dir
-            .map(|custom_dir| custom_dir.join(patch_name))
-            .filter(|path| path.exists() && path.is_file())
+    fn find_patch(patch_name: &str, custom_pages_dir: impl AsRef<Path>) -> Option<PathBuf> {
+        let patch_path = custom_pages_dir.as_ref().join(patch_name);
+        if patch_path.exists() && patch_path.is_file() {
+            return Some(patch_path);
+        }
+        None
```
niklasmohrin, line 187: Since we will allocate a `PathBuf` here in every case with `Path::join`, it would be good practice to just take a `PathBuf` upfront so that if you already have the custom pages directory as a `PathBuf` and don't need it anymore, you can give the memory to the function. I personally liked the `Option::filter` version better, but you can decide which one you like best:`Some(custom_pages_dir.join(patch_name)).filter(|path| path.is_file())` (`is_file` also implies `exists`).

src/cache.rs
```diff
@@ -217,14 +220,12 @@ impl Cache {
...
-            if custom_page.exists() && custom_page.is_file() {
-                return Some(PageLookupResult::with_page(custom_page));
-            }
+        let custom_page = custom_pages_dir.join(custom_filename);
+        if custom_page.exists() && custom_page.is_file() {
+            return Some(PageLookupResult::with_page(custom_page));
         }
 
-        let patch_path = Self::find_patch(&patch_filename, custom_pages_dir.as_deref());
+        let patch_path = Self::find_patch(&patch_filename, custom_pages_dir);
```
niklasmohrin, line 228: The logic that checks for custom pages is on the same abstraction layer as `Self::find_path`, yet one is inline here and the other is it's own function. Maybe `find_patch` is not needed at all anymore if the directory has to be specified anyways. What do you think?

src/cache.rs
```diff
@@ -217,14 +220,12 @@ impl Cache {
             .collect();
 
         // Look up custom page (<name>.page). If it exists, return it directly
-        if let Some(config_dir) = custom_pages_dir {
-            let custom_page = config_dir.join(custom_filename);
-            if custom_page.exists() && custom_page.is_file() {
-                return Some(PageLookupResult::with_page(custom_page));
-            }
+        let custom_page = custom_pages_dir.join(custom_filename);
+        if custom_page.exists() && custom_page.is_file() {
```
niklasmohrin, line 224: `exists` is not needed, `is_file` implies that

src/cache.rs
niklasmohrin: (keeping in mind that we may allow not specifying this dir after all)

src/main.rs
```diff
@@ -406,6 +400,15 @@ fn main() {
...
 
+    // Show config file and path, pass through
+    if args.flag_config_path {
+        eprintln!("Warning: The --config-path flag is deprecated, use --show-paths instead");
+        show_config_path();
+    }
+    if args.flag_show_paths {
+        show_paths(&config.directories.custom_pages_dir);
+    }
+
```
niklasmohrin, line 411: I don't know if I asked this before, but what is the rationale behind this change?

niklasmohrin (changes requested):
Not much I can say about content, just some small oddities I found. I still find it weird that the configuration comes in so many different ways, but that's not your fault. Don't really know what to think about the config stuff until @dbrgn explains [my above question](https://github.com/dbrgn/tealdeer/pull/184/files#r618759706) and we decide how to handle this. To be frank, the config seems more convoluted to me than I imagine it has to be, but we'll have to see. That said, this is nothing that this PR has to care about really, we'll just have to wait and see :man_shrugging:

src/cache.rs
dmaahs2017: Yeah, I think inlining the patch finding would be more elegant now, especially using `Option::filter` to make it a 1 liner

src/main.rs
dmaahs2017: Because now show_paths needs a reference to the custom_pages_dir from the config. So had to move this logic to after `Config::load`

src/cache.rs
dmaahs2017: Yeah I like the 1 liner too. I didn't think of manually wrapping it in Some to get access to the filter method again.

src/config.rs
dbrgn: If `PathBuf` is not an `Option`, then deserializing a TOML config file without a `custom_pages_dir` specified will fail.

(Actually, in this case it will use the default value of `PathBuf` due to the annotation, but I'm not sure what that is, and I don't think it's meaningful.)

@niklasmohrin is right, `Raw*Config` is purely here for deserializing the config files into Rust-y values, and then the non-raw config objects contain validated and potentially more structured values.

dmaahs2017:
@niklasmohrin fixed the merge conflicts. Could we try to push this in sooner than later? 😄 

I don't think there is anything else required here. Unless we want to refactor the Config Structs somehow, in which case I'd be happy to do it in a separate PR just so we can get this functionality in.

src/config.rs
dbrgn: @dmaahs2017 can you revert the removal of the options? otherwise we cannot differentiate between someone not setting the path in the config at all (which is valid of course) and someone setting an empty path (which is invalid).

With your change, if I'm not mistaken, the default value will be used if the custom pages dir is not set by the user (which is an empty path). This empty path will be joined with a filename. This means that a user that does not set the custom pages dir, but has a `tldr.tar` file in the current directory, will be served this local file instead of the actual tar page.

dbrgn (changes requested):
See the change request. We cannot merge this as-is, since the custom pages directory lookup logic would be changed.

dbrgn:
With the goal of getting a new release out soon: Replaced by #236 🙂

PR #185, opened 2021-05-09
Fixed packet/package typo, added brew instructions

cam8001:


docs/src/installing.md
```diff
@@ -2,21 +2,22 @@
...
 
 Tealdeer has been added to a few package managers:
 
 - Arch Linux AUR: [`tealdeer`](https://aur.archlinux.org/packages/tealdeer/),
   [`tealdeer-bin`](https://aur.archlinux.org/packages/tealdeer-bin/) or
   [`tealdeer-git`](https://aur.archlinux.org/packages/tealdeer-git/)
+- Brew [`tealdeer`](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/tealdeer.rb)
 - Fedora: [`tealdeer`](https://src.fedoraproject.org/rpms/rust-tealdeer)
 - FreeBSD: [`sysutils/tealdeer`](https://www.freshports.org/sysutils/tealdeer/)
 - macOS Homebrew: [`tealdeer`](https://formulae.brew.sh/formula/tealdeer)
```
SimplyDanny, line 23: Homebrew is already mentioned here referencing the official formula website.

dbrgn (changes requested):
@SimplyDanny is right. @cam8001 can you remove that line again?

The other changes are good though. And if you think it's an improvement, you could drop the "macOS " prefix for the homebrew entry.

SimplyDanny:
I think it would be an improvement since Homebrew can also be used on Linux.

dbrgn:
> I think it would be an improvement since Homebrew can also be used on Linux.

Ah, I didn't know about that. Yes, then that change would be welcome 🙂

docs/src/installing.md
cam8001: Can't believe I missed that

docs/src/installing.md
```diff
@@ -2,24 +2,24 @@
...
 
-## Packet Managers
+## Package Managers
 
 Tealdeer has been added to a few package managers:
 
 - Arch Linux AUR: [`tealdeer`](https://aur.archlinux.org/packages/tealdeer/),
   [`tealdeer-bin`](https://aur.archlinux.org/packages/tealdeer-bin/) or
   [`tealdeer-git`](https://aur.archlinux.org/packages/tealdeer-git/)
+- Brew [`tealdeer`](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/tealdeer.rb)
```
dbrgn, line 20: Sorry for nitpicking, but I think most people know the project under the name "Homebrew" (even if the package manager itself is called "brew"). At least the project refers to itself as "Homebrew". Maybe renaming it would be better? (Also, if you rename, sort alphabetically again.)

dbrgn:
I changed it for you, hope that's ok for you.

Thanks for the fix, @cam8001!

PR #186, opened 2021-05-09
Update test reference for 'tar' tldr page

SimplyDanny:


dbrgn:
Thanks! We should really switch away from `tar` (which is being updated quite often) to a rarely updated command.

dbrgn:
@SimplyDanny a PR that updates the test files would be welcome! We could also use the page for `which`. It's not simpler than `pygmentize`, but shorter.

SimplyDanny:
`pages/common/which.md` is pretty simple with respect to `{{variables}}`. I thought something more complex is preferred. We can also argue, though, that the integration test shall not test complex logic but just how everything works together. So if you like `which` more and think it is sufficient, I'm happy to open a PR to switch to it.

dbrgn:
Well, as long as it contains examples and variables, I think that's sufficient. Or do you see any complexity in `pygmentize` that's not in `which`?

SimplyDanny:
`pygmentize` has example code with multiple variables in one line with variable names not being just single words, like file names. But yeah, as said above, an integration test may use a simple example if the details are tested in other (unit) tests. So let's take `which`. 😉

niklasmohrin:
We could also use the one for `tldr`, this way are guaranteed to notice any changes to the spec 😂

PR #187, opened 2021-05-09
Use unbuffered I/O for faster printing to stdout, switch to Rustls

sondr3:
- Use unbuffered IO for faster printing, see https://llogiq.github.io/2017/06/01/perf-pitfalls.html for more
- Use statically linked `rustls` over dynamically linked `openssl`.

Compared to the benchmark in #129 from the first post (ran `hyperfine --prepare 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' './target/release/tldr tar'`). I don't know the exact command that the author of `outfieldr` ran, but you can also see a comparison I posted there. 

```
Benchmark #1: ./target/release/tldr tar
  Time (mean ± σ):      29.3 ms ±   6.2 ms    [User: 1.7 ms, System: 4.3 ms]
  Range (min … max):    24.3 ms …  50.1 ms    22 runs
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ docopt = "1"
 env_logger = { version = "0.7", optional = true }
 flate2 = "1"
 log = "0.4"
-reqwest = { version = "0.10.3", features = ["blocking", "native-tls"], default-features = false }
+reqwest = { version = "0.10.3", features = ["blocking", "rustls-tls"], default-features = false }
 serde = "1.0.21"
 serde_derive = "1.0.21"
 tar = "0.4.14"
--- a/src/formatter.rs
+++ b/src/formatter.rs
@@ -6,6 +6,7 @@ use ansi_term::{ANSIString, ANSIStrings};
 use log::debug;
 
 use crate::config::Config;
+use crate::error::TealdeerError::{self, WriteError};
 use crate::tokenizer::Tokenizer;
 use crate::types::LineType;
 
@@ -63,17 +64,21 @@ fn format_code(command: &str, text: &str, config: &Config) -> String {
 }
 
 /// Print a token stream to an ANSI terminal.
-pub fn print_lines<R, T>(writer: &mut T, tokenizer: &mut Tokenizer<R>, config: &Config)
+pub fn print_lines<T, R>(
+    writer: &mut T,
+    tokenizer: &mut Tokenizer<R>,
+    config: &Config,
+) -> Result<(), TealdeerError>
 where
-    R: BufRead,
     T: Write,
+    R: BufRead,
 {
     let mut command = String::new();
     while let Some(token) = tokenizer.next_token() {
         match token {
             LineType::Empty => {
                 if !config.display.compact {
-                    writeln!(writer).unwrap();
+                    writeln!(writer).map_err(|e| WriteError(e.to_string()))?;
                 }
             }
             LineType::Title(title) => {
@@ -85,16 +90,19 @@ where
                 debug!("Detected command name: {}", &command);
             }
             LineType::Description(text) => {
-                writeln!(writer, "  {}", config.style.description.paint(text)).unwrap();
+                writeln!(writer, "  {}", config.style.description.paint(text))
+                    .map_err(|e| WriteError(e.to_string()))?;
             }
             LineType::ExampleText(text) => {
-                writeln!(writer, "  {}", config.style.example_text.paint(text)).unwrap();
+                writeln!(writer, "  {}", config.style.example_text.paint(text))
+                    .map_err(|e| WriteError(e.to_string()))?;
             }
             LineType::ExampleCode(text) => {
-                writeln!(writer, "      {}", &format_code(&command, &text, &config)).unwrap();
+                writeln!(writer, "      {}", &format_code(&command, &text, &config))
+                    .map_err(|e| WriteError(e.to_string()))?;
             }
             LineType::Other(text) => debug!("Unknown line type: {:?}", text),
         }
     }
-    writeln!(writer).unwrap();
+    writeln!(writer).map_err(|e| WriteError(e.to_string()))
 }
--- a/src/formatter.rs
+++ b/src/formatter.rs
@@ -1,6 +1,6 @@
 //! Functions related to formatting and printing lines from a `Tokenizer`.
 
-use std::io::BufRead;
+use std::io::{BufRead, Write};
 
 use ansi_term::{ANSIString, ANSIStrings};
 use log::debug;
@@ -63,16 +63,17 @@ fn format_code(command: &str, text: &str, config: &Config) -> String {
 }
 
 /// Print a token stream to an ANSI terminal.
-pub fn print_lines<R>(tokenizer: &mut Tokenizer<R>, config: &Config)
+pub fn print_lines<R, T>(writer: &mut T, tokenizer: &mut Tokenizer<R>, config: &Config)
 where
     R: BufRead,
+    T: Write,
 {
     let mut command = String::new();
     while let Some(token) = tokenizer.next_token() {
         match token {
             LineType::Empty => {
                 if !config.display.compact {
-                    println!()
+                    writeln!(writer).unwrap();
                 }
             }
             LineType::Title(title) => {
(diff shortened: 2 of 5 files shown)


tests/tar-markdown.expected
```diff
@@ -4,34 +4,34 @@
 > Often combined with a compression method, such as gzip or bzip2.
 > More information: <https://www.gnu.org/software/tar>.
 
-- [c]reate an archive from [f]iles:
+- [c]reate an archive and write it to a [f]ile:
```
sondr3, line 7: Had to update this test file, let me know if I did this correctly.

src/formatter.rs
```diff
@@ -68,6 +68,8 @@ where
     R: BufRead,
 {
     let mut command = String::new();
+    let mut stdout = std::io::stdout();
+    stdout.lock();
     while let Some(token) = tokenizer.next_token() {
```
niklasmohrin, line 73: I think the lock should be passed via an argument

src/formatter.rs
```diff
@@ -68,6 +68,8 @@ where
...
         match token {
             LineType::Empty => {
```
niklasmohrin, line 75: This branch is still using `println`

niklasmohrin (changes requested):
Love it, great idea

Here is what I measured on my machine

image

Cargo.toml
```diff
@@ -23,7 +23,7 @@ docopt = "1"
 env_logger = { version = "0.7", optional = true }
 flate2 = "1"
 log = "0.4"
-reqwest = { version = "0.10.3", features = ["blocking", "native-tls"], default-features = false }
+reqwest = { version = "0.10.3", features = ["blocking", "rustls-tls"], default-features = false }
```
dbrgn, line 26: Yeah, rustls ALL THE THINGS! 🙂

src/formatter.rs
sondr3: Good catch!

src/formatter.rs
niklasmohrin: Yes, maybe even a generic type that implements `io::Read`

It's just something to try out for now, maybe we don't like that version then, but I would like to see it 😅

sondr3: Alright, I'll fix these comments later today :smiley:

sondr3:
Updated according to the reviews, I'm on a different computer so can't run the benchmark again but there shouldn't be a difference.

src/formatter.rs
```diff
@@ -63,16 +63,17 @@ fn format_code(command: &str, text: &str, config: &Config) -> String {
 }
 
 /// Print a token stream to an ANSI terminal.
-pub fn print_lines<R>(tokenizer: &mut Tokenizer<R>, config: &Config)
```
niklasmohrin, line 66: I think it's convention that the types are listed in the same order they are used, but I am not sure :thinking:

src/main.rs
```diff
@@ -90,14 +90,23 @@ fn print_page(
         let reader = BufReader::new(file);
 
         if enable_markdown {
+            let mut stdout = std::io::stdout();
+            stdout.lock();
```
niklasmohrin, line 94: I am surprised the compiler / linter didn't catch this: This returns a lock that can be written to, it doesn't actually lock stdout for our variable `stdout`. So this should roughly be `let stdout = io::stdout().lock()`

You also do this in both branches of this conditional, so it can be moved out :)

src/main.rs
```diff
@@ -90,14 +90,23 @@ fn print_page(
...
+                .flush()
+                .map_err(|_| "Could not flush stdout".to_string())?;
         } else {
             // Create tokenizer and print output
             let mut tokenizer = Tokenizer::new(reader);
-            print_lines(&mut tokenizer, &config);
+            let mut stdout = std::io::stdout();
+            stdout.lock();
+            print_lines(&mut stdout, &mut tokenizer, &config);
+            stdout.flush().map_err(|err| err.to_string())?;
```
niklasmohrin, line 109: This can also be done after this conditional block, why exactly is a flush error reported differently depending on `enable_markdown`?

niklasmohrin (changes requested):
Having so many `unwrap` calls looks a bit wrong, but it is equivalent to what we had before, so... :)

src/formatter.rs
sondr3: I’ve never thought about it, happy to fix..

src/main.rs
sondr3: Just didn’t cross my mind 🤷‍♂️

sondr3:
I actually started converting the `unwrap`s to proper errors, but figured it was out of the scope of this PR. Happy to fix though, it may take a few days as we are celebrating our Independence Day tomorrow 😄

sondr3:
I think I managed to fix the requested changes, this is ready for another round of reviews.

src/main.rs
```diff
@@ -88,17 +88,25 @@ fn print_page(
     for path in page.paths() {
```
niklasmohrin, line 88: While you are here, could you remove this comment? :sweat_smile:

src/main.rs
```diff
@@ -88,17 +88,25 @@ fn print_page(
...
-                println!("{}", line.unwrap());
+                writeln!(handle, "{}", line.unwrap())
+                    .map_err(|_| "Could not write to stdout".to_string())?;
             }
         } else {
             // Create tokenizer and print output
             let mut tokenizer = Tokenizer::new(reader);
-            print_lines(&mut tokenizer, &config);
+            print_lines(&mut handle, &mut tokenizer, &config)
+                .map_err(|_| "Could not write to stdout".to_string())?;
```
niklasmohrin, line 104: I think that this could use the specific error:

```
print_lines(&mut handle, &mut tokenizer, &config)
                .map_err(|e| format!("Could not write to stdout: {}", e)?;
```

This would mean that there should be an `impl` of the `Display` trait for `TealdeerError`, but that would just use the `.msg()` method you just wrote. (you could also use `e.msg()` here)

niklasmohrin (approved):
Great! Only thing I can really point out is that you wrote the helpful error message for writing to stdout, but discard it at the caller. I think it's better to propagate this information to the user (in the unlikely case that this goes wrong)

Other than that, thank you for your contribution :+1:

sondr3:
Done :smiley:

niklasmohrin (approved):

dbrgn:
Thanks for the first round of reviews, @niklasmohrin. I'll try to take a look this weekend!

src/formatter.rs
dbrgn: I'm not aware of that convention, I usually try to put them in a logical order. Sometimetimes no logical order is obvious though 🙂

dbrgn (changes requested):
Very nice, thanks @sondr3!

The switch to RusTLS causes an increase in size of roughly 1.8 MiB (pre-strip). After stripping, the size increase is 868 KiB, which I think is fine given that OpenSSL is gone (which makes cross-compilation and static builds so much simpler).

Regarding the performance, I did a hyperfine comparison. Plain `tldr` which just prints the help page:

```
Before
  Time (mean ± σ):      12.1 ms ±   0.8 ms    [User: 1.6 ms, System: 2.7 ms]
  Range (min … max):    10.4 ms …  14.1 ms    45 runs
After
  Time (mean ± σ):      11.0 ms ±   0.7 ms    [User: 1.7 ms, System: 2.0 ms]
  Range (min … max):    10.1 ms …  13.2 ms    33 runs
```

And with `tldr tar`:

```
Before
  Time (mean ± σ):      14.0 ms ±   1.1 ms    [User: 1.5 ms, System: 3.2 ms]
  Range (min … max):    12.2 ms …  17.3 ms    32 runs
After
  Time (mean ± σ):      13.8 ms ±   0.8 ms    [User: 1.5 ms, System: 2.6 ms]
  Range (min … max):    12.5 ms …  15.5 ms    39 runs
```

Looks like a small but clear improvement. Depending on the printed page and I/O speed, it might make a bigger difference.

src/main.rs
```diff
@@ -84,21 +84,28 @@ fn print_page(
     enable_markdown: bool,
     config: &Config,
 ) -> Result<(), String> {
-    // Open file
     for path in page.paths() {
         let file = File::open(path).map_err(|msg| format!("Could not open file: {}", msg))?;
         let reader = BufReader::new(file);
+        let stdout = std::io::stdout();
+        let mut handle = stdout.lock();
```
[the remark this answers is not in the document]
niklasmohrin: Good catch. On that note: The flush could then be moved out of the loop too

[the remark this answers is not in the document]
sondr3: Wow, good catch, can't believe I overlooked that :sweat_smile: Fixed both your comments.

dbrgn (approved):
Thank you @sondr3!

PR #188, opened 2021-05-10
Fix clippy warnings

dmaahs2017:
- allow struct excessive bools for Args struct.
  - change tests to use write_all instead of write
  - allow dead_code for the ConvigVar variant of PathSource

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -108,6 +108,8 @@ pub enum PathSource {
     OsConvention,
     /// Env variable (TEALDEER_*)
     EnvVar,
+
+    #[allow(dead_code)]
     /// Config file variable
     ConfigVar,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -57,6 +57,7 @@ const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/master.tar
 #[cfg(not(target_os = "windows"))]
 const PAGER_COMMAND: &str = "less -R";
 
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Debug, Deserialize)]
 struct Args {
     arg_command: Option<Vec<String>>,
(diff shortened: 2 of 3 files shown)


dmaahs2017:
@niklasmohrin review please :) 

this just fixes `clippy --all-targets` warnings. They were bothering me. 

I wasn't sure if there was a reason for the ConfgVar variant in the `PathSource` enum. Presumably we will use it in the future. So for now I `#allow(dead_code)` for it. Otherwise I can just as easily delete it.

src/types.rs
```diff
@@ -108,6 +108,8 @@ pub enum PathSource {
     OsConvention,
     /// Env variable (TEALDEER_*)
     EnvVar,
+
+    #[allow(dead_code)]
```
dbrgn, line 112: Maybe add a comment here?

    #[allow(dead_code)] // Waiting for #141

dbrgn (changes requested):
@dmaahs2017 the problem with the "excessive bools" lint is that the version in CI does not know that lint yet. Right now we're testing with 1.51.0. Feel free to update that version in the CI config for the clippy job to the latest stable (1.52.0)!

niklasmohrin (commented):
Can only agree with @dbrgn, comment seems like a good idea and feel free to bump the rust version in `.github/workflows/ci.yaml`. Other than that, there's nothing to complain about. Good catch with `write_all`!

@dbrgn Why do we have MSRV anyway? It's not like we're a library that people depend on, so does it even matter? We could just say "builds on stable, everything else no guarantees"

dmaahs2017:
> Can only agree with @dbrgn, comment seems like a good idea and feel free to bump the rust version in `.github/workflows/ci.yaml`. Other than that, there's nothing to complain about. Good catch with `write_all`!
> 
> @dbrgn Why do we have MSRV anyway? It's not like we're a library that people depend on, so does it even matter? We could just say "builds on stable, everything else no guarantees"

I'm curious why we run our tests on 1.41.1 as well as on stable

niklasmohrin:
@dmaahs2017 FYI, further steps with MSRV are currently discussed in #190

dbrgn (approved):

PR #189, opened 2021-05-10
Refactor formatting code

niklasmohrin:
See [this discussion](https://github.com/dbrgn/tealdeer/pull/178#discussion_r615385557).

This aims to introduce some abstractions where needed and refactor code that has accumulated complexity over time


niklasmohrin:
@dbrgn I bumped MSRV to 1.52 here because I wanted to use `str::split_once`. Like I said, I don't really see why we would have to support older versions of the compiler. Are you okay with that or do you see any problems?

niklasmohrin:
Tests are complaining about highlighting, but I think it looks fine. I checked the binary output and it seems that the current implementation outputs more color modifiers than what I refactored it to (left / "stable.out" is current impl, right / "master.out" is this PR)

image

```console
$ du *.out --bytes
1471    master.out
1542    stable.out
``` 

Can someone take a look that knows exactly what is important for the coloring? It looks just fine to me :smile: @SimplyDanny

src/formatter.rs
```diff
@@ -5,60 +5,66 @@ use std::io::BufRead;
...
-                .nth(code_part_end_pos + command_start - 1);
-            if char_before_command.filter(|c| c.is_whitespace()).is_some() {
-                // ... or when preceded by a whitespace character.
-                parts.push(config.style.command_name.paint(command));
-            } else {
-                parts.push(config.style.example_code.paint(command));
-            }
+            let split_point = part
+                .char_indices()
+                .nth(part.chars().count() - command_name.chars().count() - 1)
```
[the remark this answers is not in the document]
SimplyDanny: I see. We should create a test for non-ASCII input.

src/formatter.rs
```diff
@@ -2,77 +2,111 @@
...
+        .char_indices()
+        .take_while(|&(i, _)| i < start)
+        .last()
+        .filter(|(_, prev_char)| !prev_char.is_whitespace())
+        .is_none()
+        && surrouding[end..]
+            .chars()
+            .next()
+            .filter(|next_char| !next_char.is_whitespace())
+            .is_none()
```
SimplyDanny, line 46: For more clarity maybe assign both conditions to (named) variables, e.g. `char_before_start_is_whitespace` and `char_after_end_is_whitespace`.

Also, the doubled negation can be avoided with `....filter(|c| c.is_whitespace()).is_some()`.

src/formatter.rs
```diff
@@ -83,13 +117,119 @@ where
...
             }
+
             LineType::Other(text) => debug!("Unknown line type: {:?}", text),
         }
     }
-    println!();
+    yield_snippet(HighlightingSnippet::Linebreak);
+}
+
+fn print_snippet<'a>(snip: HighlightingSnippet<'a>, style: &StyleConfig) {
```
SimplyDanny, line 134: In object-oriented fashion this function could be added as a method to `HighlightingSnippet`.

src/formatter.rs
```diff
@@ -83,13 +117,119 @@ where
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use HighlightingSnippet::*;
+
+    #[test]
+    fn test_is_freestanding_substring() {
+        let t = |s: &str, words: &[(usize, usize)], expected: bool| {
```
SimplyDanny, line 154: This local function is only helpful for line 160 in that it iterates over the array of sub-strings. In test code it is sometimes better to be a bit more verbose instead of hiding functionality somewhere. As a reader I would have liked

    t("I love tldr", (0, 1), true);
    t("I love tldr", (2, 6), true);
    t("I love tldr", (7, 11), true);

more which would have avoided the `&[...]`s in addition.

SimplyDanny (commented):
Just some more ideas ...

src/formatter.rs
niklasmohrin: Actually, if there is no char before the substring, this would be `None` too, which is why I wrote it like this. It's the best way I found for actually doing the "doesn't exist or is whitespace" check in one go.

I added the names, maybe it's clearer with that?

src/formatter.rs
niklasmohrin: Yeah, that thought crossed my mind too, but in the end I decided that I would like `HighlighttingSnippet` to be just a "stupid" data type rather than an active player. I found it weird that it would "know" the `StyleConfig` type somehow and I didn't want to encode how to use the type in it's definition but have `print_lines` tie the knots between this whole module and the tealdeer cli.

Is there another (OO) intuition that I don't see here and fits this whole module better?

src/formatter.rs
niklasmohrin: Good point :+1:

niklasmohrin: Actually, I could remove `t` altogether :thinking:

src/formatter.rs
SimplyDanny: I got your point with the `None` case. The variables make things clearer, indeed. As both checks do the same thing we could also introduce another method implementing what the comment explains to get rid of it (some people don't like comments :smile:), like:
```rust
fn is_whitespace_or_does_not_exist(char: Option<char>) -> bool {
    char.filter(|c| !c.is_whitespace()).is_none()
}

fn is_freestanding_substring(surrouding: &str, substring: (usize, usize)) -> bool {
    let (start, end) = substring;
    let char_before = surrouding[..start].chars().last();
    let char_after = surrouding[end..].chars().next();
    is_whitespace_or_does_not_exist(char_before) && is_whitespace_or_does_not_exist(char_after)
}
```

What do you think?

src/formatter.rs
SimplyDanny: To get this really clean either `print_lines` and `print_snippet` should be moved to `main.rs` so that the formatting module does not even need to know `StyleConfig` (and the `Tokenizer`) or `StyleConfig` should be defined in `formatter.rs` itself so it can be used freely inside this module (thus `print_snippet` can be a method of `HighlightingSnippet`). Otherwise, the formatting code will not really be separated from the rest.

From a testing perspective, dealing with `StyleConfig` is more work, however, so getting rid of it wherever possible might be preferred.

src/formatter.rs
niklasmohrin: hm, I don't know, seems kind of extreme to me :man_shrugging: I would find this a bit weird as its own function

src/formatter.rs
niklasmohrin: @SimplyDanny I managed to extract the printing from `main.rs` and `formatter.rs` (for details see commit messages)

niklasmohrin:
@dbrgn @SimplyDanny I think I managed to include everything I had in mind, your turn :)

src/formatter.rs
SimplyDanny: Fine for me.

niklasmohrin:
@dbrgn In case you feel like merging this PR, please make sure **not** to squash the history, so that we can keep the documentation I put in the commits

src/line_iterator.rs
```diff
@@ -41,38 +41,40 @@ where
...
-
                 if self.first_line {
-                    // Clear `first_line` flag
-                    self.first_line = false;
-
-                    // It's the old format.
-                    self.format = TldrFormat::V1;
+                    if self.current_line.starts_with('#') {
+                        // It's the old format.
+                        self.format = TldrFormat::V1;
```
dbrgn, line 63: Sidenote: I wonder whether we still need to support the old format. The pages are all converted to the new format, right? (If we want to be super-cautious we could emit a deprecation warning when we detect a V1 page, and ask the user to refresh the cache.)

src/formatter.rs
```diff
@@ -93,23 +89,16 @@ fn highlight_code<'a, E>(
...
+pub fn highlight_lines<L, F, E>(
+    lines: L,
+    yield_snippet: &mut F,
     keep_empty_lines: bool,
-) -> Result<(), TealdeerError>
+) -> Result<(), E>
 where
-    T: Write,
+    L: Iterator<Item = LineType>,
+    F: for<'snip> FnMut(HighlightingSnippet<'snip>) -> Result<(), E>,
```
dbrgn, line 38: What kind of syntax is `for<'snip>`? 😄 I'm not familiar with that.

src/formatter.rs
```diff
@@ -1,108 +1,230 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Represents a snippet from a page from a specific highlighting class.
+pub enum HighlightingSnippet<'a> {
+    CommandName(&'a str),
+    Variable(&'a str),
+    NormalCode(&'a str),
+    Description(&'a str),
+    Text(&'a str),
+    Linebreak,
 }
```
dbrgn, line 17: I think the name of this enum is not ideal, since the type will be used even when disabling any syntax highlighting.

Maybe call it `OutputSnippet` instead? (Unless you have a better idea. Not a big fan of that one either, but I couldn't come up with something better.)

src/formatter.rs
```diff
@@ -1,108 +1,230 @@
...
-/// Print a token stream to an ANSI terminal.
-pub fn print_lines<T, R>(
-    writer: &mut T,
-    tokenizer: &mut Tokenizer<R>,
-    config: &Config,
-) -> Result<(), TealdeerError>
+/// Parse the content of each line yielded by `lines` and yield `HighLightingSnippet`s accordingly.
+pub fn highlight_lines<L, F, E>(
+    lines: L,
+    yield_snippet: &mut F,
```
dbrgn, line 33: The `yield` terminology confused me a bit, I know it mostly from the context of iterators/generators (it's fine for `lines`, which _is_ an iterator).

What do you think about renaming the closure (and related code / comments) to `process_snippet` or `emit_snippet`?

dbrgn (changes requested):
Thanks, this is a clear improvement.

However, with all the ad-hoc parsing in multiple passes (first we detect line types in the tokenizer, then we detect variables in `highlihgt_code`, then we detect the code name in `highlight_code_segment` I wonder if in the long term it would be better if the tokenizer (now line iterator) were more powerful and could emit more fine-grained tokens on the first pass. We would need up to `len(command_name)` lookahead, but that's not an issue at all. This way, we would not need to differentiate between `LineType` tokens and `HighlightingSnippet` anymore (they are kind of similar if you give up the line based approach that I used initially for simplicity).

The main reason for splitting tokenization and output formatting in parsers is that the processing of the input data is separated from the formatting of the output. Right now we process input data both in the tokenizer and in the formatter, which isn't really the cleanest approach.

(However, that was just an observation and does not affect this PR.)

> @dbrgn In case you feel like merging this PR, please make sure not to squash the history, so that we can keep the documentation I put in the commits

Sure, which such clean commit descriptions there's no need for squashing :) If you want, you could squash the commit "Remove unneeded imports in main.rs" into the first commit ("Add FindFrom extension trait to str to start searching at a byte offset") which added those imports in the first place.

src/formatter.rs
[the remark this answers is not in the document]
dbrgn: Fortunately Rust makes it quite hard to mis-use unicode. For example, strings do not allow indexing (to avoid the chance of indexing on a non-boundary) 🙂 So in practice, as long as you're not unwrapping fallible functions, I've never seen a Rust codebase with any of the "simple" unicode bugs.

src/formatter.rs
niklasmohrin: It means that we want `F` to be generic over the lifetime `'snip`. We actually need it to be able to handle multiple disjoint lifetimes (each item in line iterator clears the buffer, so the references from the previous iteration become invalid)

src/line_iterator.rs
niklasmohrin: Are they? As far as I can tell, none of them are

```
.cache/tealdeer/tldr-master via  v16.3.0 took 5s
❯ fd --type f '.*' pages* | xargs -n1 head -1q |  wc -l
4408
.cache/tealdeer/tldr-master via  v16.3.0 took 6s
❯ fd --type f '.*' pages* | xargs -n1 head -1q | rg '^#' | wc -l
4408
```

or am I misunderstanding something?

src/formatter.rs
niklasmohrin: I feel like `OutputSnippet` implies too much what it is supposed to be used for, not what it actually contains. What about something like `PageSnippet`?

src/formatter.rs
niklasmohrin: Ah that's unlucky, I thought it would make it easier to understand if these function actually followed iterator naming conventions, because they really are an iterator adapter or rather a stream adapter. I would really like to model this out as a big stream of `*Snippet`, but it will probably get difficult with lifetimes. The current implementation requires callers to give the generic handler `yield_snippet` to immediately dispose of any emitted snippets instead of accumulating them somewhere (like a more abstract stream model would).

I can see that the responsibilities taken by the different modules are a bit weird, maybe an approach that parses code lines before even yielding the line would be cleaner, we could explore that route

src/formatter.rs
```diff
@@ -1,84 +1,122 @@
...
     R: BufRead,
 {
     let mut command = String::new();
+    let mut yield_snippet = |snip: HighlightingSnippet<'_>| {
+        if snip.is_empty() {
+            Ok(())
+        } else {
+            print_snippet(writer, snip, style).map_err(|e| WriteError(e.to_string()))
+        }
+    };
```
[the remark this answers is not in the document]
niklasmohrin: I would be okay with that, see the other comment below

src/line_iterator.rs
dbrgn: Ah, you are right! https://github.com/tldr-pages/tldr/pull/958 Then we still need it.

src/formatter.rs
dbrgn: Cool!

src/formatter.rs
dbrgn: Yep, that sounds good!

src/formatter.rs
dbrgn: I see. With the current code, I still think it's better to use `process_snippet` or `emit_snippet`. Pick whichever you prefer (I think I'd prefer "process").

dbrgn:
I think only the renaming and a rebase against `master` are left until we can merge this!

dbrgn (approved):
Cool, thank you!

PR #191, opened 2021-05-12
Update MSRV policy

### LINKED ISSUE #190: RFC: Tealdeer MSRV Policy
Right now Tealdeer has a "bump it once we need it" MSRV (Minimal Supported Rust Version) policy. However, the question arose whether this makes sense or not.

As an application (as opposed to a library), there is less of a problem if we always follow stable Rust. The main concern would be distro packaging. If we always jump right on the latest version, we'll make the life harder for packagers.

Right now, according to https://pkgs.org/search/?q=rustc, the following rustc versions are distributed:

- Debian 10 (released): 1.41
- Debian 11 (testing): 1.48
- Ubuntu 21.04 (latest): 1.50
- Ubuntu 20.04 (LTS): 1.41
- FreeBSD 13: 1.51
- OpenSUSE Leap 15.2: 1.43.1

I'm not sure how much effort it is for distro packagers to use a newer rustc when building packages. Right now tealdeer is packaged in the following non-rolling-release distros:

- Fedora
- FreeBSD
- NetBSD
- OpenSUSE
- Solus

---

I see a few options:

### 1. Always track current stable

  - This is most convenient for tealdeer developers
 
### 2. Support the lowest version that we can with our current set of dependencies

  - Dependencies often upgrade the MSRV even in patch releases. This is something I hate, but it's the reality. It's a lot of work to guarantee a fixed MSRV.
  - Bumping the MSRV only if a dependency forces us to, or if there's a feature that we _really_ want, does not cause us developers a lot of trouble, but may help packagers. However, there can still be cases where a distro-shipped rustc is too old to build tealdeer.

### 3. Support "stable + n" versions

For example, if we pick "stable + 4", we'd be able to support building on any system that has a rustc which is not older than half a year. With "stable + 2", we'd be able to support building on any system that has a rustc which is not older than 3 months.

- This gives distributions some time to catch up with Rust packaging
- For the developers it means that all new features can be used, just with a bit of a delay

### 4. Support the version that latest mainstream distros (e.g. Debian, Ubuntu, Red Hat) support

  - Most convenient for packagers
  - This might mean that we'd need to actively pin dependencies to guarantee the MSRV, it's quite an effort

---

Since tealdeer isn't exactly a critical project, I'd be against 4, unless there are compelling reasons for it.

Always tracking stable (1) is convenient, but usually it's very little effort for the developers to support a few more versions of rustc, and it can avoid a lot of pain for packagers.

Right now we're roughly following 2. I'd tend towards picking 2 or 3 for our future MSRV policy.

What are your thoughts, @niklasmohrin @dmaahs2017?

dbrgn:
Here's the updated policy, as discussed in #190.

Closes #190.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [1.41.1, stable]
+        rust: [1.52, stable]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v2
@@ -45,9 +45,9 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-            toolchain: 1.52.0
-            components: clippy
-            override: true
+          toolchain: 1.52
+          components: clippy
+          override: true
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
@@ -60,8 +60,8 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions-rs/toolchain@v1
       with:
-          toolchain: 1.52.0
-          override: true
+        toolchain: 1.52
+        override: true
     - run: rustup component add rustfmt
     - uses: actions-rs/cargo@v1
       with:
--- a/README.md
+++ b/README.md
@@ -90,12 +90,8 @@ To run lints:
 
 ## MSRV (Minimally Supported Rust Version)
 
-Tealdeer will not bump the MSRV requirement in patch versions, but it may
-increase it in minor versions. The reason is that many important libraries
-(e.g. the Tokio ecosystem, which is a dependency of reqwest, which is used for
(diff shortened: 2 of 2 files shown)


dbrgn:
I also bumped the Rust version in CI to 1.52. I still think there's value in having a lower bound in the CI config, but we can now bump it more freely if building fails.

niklasmohrin (approved):
That seems like a good rule 👍

I would like to point out two things:

1. I think we shouldn't pin the patch version of MSRV, we can just say 1.52
2. we could have the msrv as a constant at the top to remove duplication (see https://www.github.com/sharkdp/bat/tree/master/.github%2Fworkflows%2FCICD.yml)

dbrgn:
> I think we shouldn't pin the patch version of MSRV, we can just say 1.52

Good point!

> we could have the msrv as a constant at the top to remove duplication

Again, good point. I'll update my second commit.

dbrgn:
(Also, we should drop the version number from the test name, otherwise we'll have to update the branch protection rules every time we bump the MSRV.)

dbrgn:
Unfortunately using a variable [doesn't work](https://github.com/actions/runner/issues/480), so I'll have to revert that.

PR #192, opened 2021-05-13
Use `which` instead of `tar` as reference tldr page in tests

SimplyDanny:
The page for `tar` changes pretty frequently. `which` was last updated in 2018 and is a rather simple command overall. So it is less likely that the reference in the repository has to be updated as often as the one for `tar`.

Also see conversation in #186.


dbrgn (approved):
Thank you!

PR #193, opened 2021-06-21
Update @SShrike → @severen

bl-ue:
:)


severen:
GitHub actually pinged me for this. I can indeed confirm my identity here!

bl-ue:
Yeah, I didn't actually intend for that, but it worked out that way 🙂

PR #197, opened 2021-07-29
Implement support for italics

### LINKED ISSUE #146: Allow italics styling
I like how https://github.com/raylee/tldr-sh-client shows example arguments in italics, and would like to configure tealdeer similarly, but apparently that's not supported yet :/

WesleyKlop:
Adds a configuration option for displaying information with the italics style. 

Fixes #146 

Before: 
image

After (configured example_variable to use italic instead of underline):
image

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -24,6 +24,10 @@ fn default_bold() -> bool {
     false
 }
 
+fn default_italic() -> bool {
+    false
+}
+
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 #[serde(rename_all = "lowercase")]
 pub enum RawColor {
@@ -64,6 +68,8 @@ struct RawStyle {
     pub underline: bool,
     #[serde(default = "default_bold")]
     pub bold: bool,
+    #[serde(default = "default_italic")]
+    pub italic: bool,
 }
 
 impl Default for RawStyle {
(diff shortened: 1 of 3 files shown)


niklasmohrin:
Thank you for contributing, this looks good! :rocket:

Could you add an integration test that checks that italic styling works (or enhance one of the styling tests to style something as italic too)? It should be as easy as picking something in the test config.toml to be rendered as italic and updating the expected results

WesleyKlop:
Sorry for the late reply, will add this when I have some time :)

WesleyKlop:
Well that was easy :P

niklasmohrin (approved):
Very nicely done. Thanks a lot!

src/config.rs
```diff
@@ -24,6 +24,10 @@ fn default_bold() -> bool {
     false
 }
 
+fn default_italic() -> bool {
+    false
+}
+
```
dmaahs2017, line 30: What is the purpose of this function? Could we just use a constant instead?

dmaahs2017, line 30: What is the purpose of this function? 

It may be better to just use a `const ITALICS_DEFAULT: bool = false`

Or just inline in the `RawStyle` annotation. `#[serde(default = false)]`

WesleyKlop: I looked at how other styles were implemented (like bold and underline) and those were done the same way, I was not looking to make changes to unrelated stuff

WesleyKlop: Also the project does not build when inlining the default value, see: https://serde.rs/field-attrs.html#default--path

niklasmohrin:
looks good to me @dbrgn

dbrgn (approved):
Thank you @WesleyKlop!

PR #200, opened 2021-08-09
Run clippy on stable with msrv setting instead of on msrv

niklasmohrin:
I noticed that this setting exists the other day.

Using this, we should always get the newest (and fixed) lints from stable without comprimising msrv. Note that we are still building and testinig on msrv, so we still catch all these errors. :^)


dbrgn:
Oh, nice. Does this mean that Clippy will show the same results, no matter what local Rust version is installed?

dbrgn (approved):

niklasmohrin:
No, this changes nothing on how rustup determines which version to run. It just tells the version that is used to not lint things that are not possible in 1.52. For example `bool::then` is enforced by clippy over `if { Some(...) } else { None }`, but if your msrv is set lower than the version that introduced `bool::then` it won't complain. This means that we can use recent versions of clippy without it contradicting to msrv build

dbrgn:
Ah, that's even better 🙂

PR #201, opened 2021-08-28
Add various small refactorings

tranzystorek-io:
Highlights:

- Remove redundant `.iter()` / `.into_iter()` calls
- Use arrays and their new IntoIterator impl (instead of vecs/slices)
- Remove redundant `.as_bytes()` calls


dbrgn (approved):
Nice, thank you! Looks good from my side, can be merged as soon as tests pass.

tranzystorek-io:
Ah, the IntoIterator impl for arrays is available from Rust 1.53, and I see 1.52 in CI 😕

dbrgn:
As per #191, feel free to bump the Rust version in CI! (It would be great if you could also reorder the commits so that the version bump commit comes first, in case you know how to do that.)

tranzystorek-io:
Done 😀

dbrgn (changes requested):
Ah, one last thing: Can you also update the version in `clippy.toml` (ideally in the same commit as the CI version bump)?

dbrgn (approved):

Issue #181, opened 2021-04-22
Custom patches should not need to be formatted like a full page


dmaahs2017:
Quote from PR #142 discussion:
> Another issue for a new PR: Right now the patches need to be formatted like a full page, including a heading. Otherwise, the first line is interpreted as the heading, and skipped when printing (because the page and the patch are passed to separate tokenizer instances). However, I think that too can be fixed in a separate PR, I don't want to hold up this PR too long.

Pre-requisites:
* Issue #180

Acceptance Criteria: 
* Format of .patch files should be expected to not have a heading
* Update tests to reflect this change

dbrgn:
Thanks for creating this issue, I almost forgot about it 🙂

@dmaahs2017 are you interested in doing this change?

I see two ways to solve this.

Right now the tokenizer expects a file to start with a heading. Maybe we should parametrize the tokenizer to tell it whether to expect a page or a patch.

(We could also just tokenize whatever we get thrown at us, including - for example - headings in the middle of the page. However, slightly stricter parsing is probably a good idea.)

Another approach would be to concatenate the input data _before_ tokenizing it. Right now we just "print" the page and the patch in sequence, which results in a new tokenization / formatting step for both. If we'd concatenate the source, we could parse it as if it were a single page.

@niklasmohrin maybe you also have some good input here.

dmaahs2017:
I'm leaning towards concatenating the text together then running it through the tokenizer, because when `--markdown` flag is enabled the raw markdown is output as if they are a single page, so the concatenation method is consistent with that. 

This would also be simpler to implement, and require no re-working of the Tokenizer. The downside that I see is that we'll have to Read all the pages into 1 String before sending that to the tokenizer.

niklasmohrin:
I am against collecting all the pages into a string up front. I think that the `Tokenizer` (which just transforms an iterator of lines into an iterator of tokens). I would really like to have this as a `struct Tokens` that impls `Iterator` but also only takes an `Iterator<Item=&str>` instead of a `BufRead`. This is hard (impossible) though, because the this iterator would have to yield string slices for a set lifetime `'a` in which all references are valid. But the lifetimes of these references are _designed_ to be disjoint in order to buffer the lines. So that won't happen. (I think it can't hurt to implement `Iterator` because it provides all of the helper methods for no additional cost)

There is `.chain` on readers that chains two readers of any kind. The resulting struct also implements `BufRead`, so `Tokenizer` wouldn't have to be changed.

[...]

PR #202, opened 2021-09-01
spike/seq file reader. 

dmaahs2017:
This is a proposed solution to #181 

- spike SeqFileReader to abstract away reading multiple files into the tokenizer
- clippy + formatting


dbrgn:
Replaced in favor of #238.

PR #204, opened 2021-09-12
Use tree-style imports

dbrgn:
Initially I was sceptical, but by now I really like the tree-style imports.


niklasmohrin (approved):
I don't really have a hard preference for either style, I am more used to what we have right now though. We can also come back to the topic if we notice that it is harder to read or we don't like it anymore later :shrug:

PR #206, opened 2021-09-23
chore: download main instead of master

invakid404:
In relation to [this tldr release](https://github.com/tldr-pages/tldr/releases/tag/v1.5b), tldr clients should use the main branch and not the master branch. The master branch is simply a mirror of main right now and will be removed in the future, so I thought I might as well fix this real quick.


niklasmohrin (changes requested):
Thank you! We may have missed this change otherwise :)

Edit: Ah, I see you have already started replacing the other occurrences, thank you. To get CI to pass, you will have to pass rustfmt and clippy too. Just run `cargo fmt` in the project directory and make sure that `cargo clippy` doesn't add any new warnings

There are more places where we have the magic word "master": 
```console
$ rg master -g '*.rs'
src/main.rs
53:const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/master.tar.gz";
195:            path.push("tldr-master");
273:    // https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language

src/cache.rs
166:            if let Ok(metadata) = fs::metadata(cache_dir.join("tldr-master")) {
220:            Ok((cache_dir, _)) => cache_dir.join("tldr-master"),
266:        let platforms_dir = cache_dir.join("tldr-master").join("pages");

tests/lib.rs
63:            .join("tldr-master")
205:        testenv.cache_dir.path().join("tldr-master"),
312:                .join("tldr-master")
537:    let cache_file_path = testenv.cache_dir.path().join("tldr-master");
```

It would be great if you could extract this name into its own constant :+1: If you are not confident in Rust, that's not a problem either :)

invakid404:
> Thank you! We may have missed this change otherwise :)
> 
> There are more places where we have the magic word "master":
> 
> ```
> $ rg master -g '*.rs'
> src/main.rs
> 53:const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/master.tar.gz";
> 195:            path.push("tldr-master");
> 273:    // https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#language
> 
> src/cache.rs
> 166:            if let Ok(metadata) = fs::metadata(cache_dir.join("tldr-master")) {
> 220:            Ok((cache_dir, _)) => cache_dir.join("tldr-master"),
> 266:        let platforms_dir = cache_dir.join("tldr-master").join("pages");
> 
> tests/lib.rs
> 63:            .join("tldr-master")
> 205:        testenv.cache_dir.path().join("tldr-master"),
> 312:                .join("tldr-master")
> 537:    let cache_file_path = testenv.cache_dir.path().join("tldr-master");
> ```
> 
> It would be great if you could extract this name into its own constant  If you are not confident in Rust, that's not a problem either :)

I'll extract it to a constant, no problem. I'll update the PR shortly.

src/cache.rs
```diff
@@ -217,7 +217,7 @@ impl Cache {
 
         // Get cache dir
         let cache_dir = match Self::get_cache_dir() {
-            Ok((cache_dir, _)) => cache_dir.join("tldr-master"),
+            Ok((cache_dir, _)) => cache_dir.join("tldr-main"),
```
threema-danilo, line 220: This will result in a "cache directory missing" warning, and two caches side by side.

@niklasmohrin what do you think, should we simply keep the name "tldr-master" so that the cache is re-used? Should we rename the old directory if present?

dbrgn, line 220: This will result in a "cache directory missing" warning, and two caches side by side.

@niklasmohrin what do you think, should we simply keep the name "tldr-master" so that the cache is re-used? Should we rename the old directory if present?

niklasmohrin: I think we should remove the name "tldr-master" too, otherwise the upstream effort would be a pointless. The easiest change would probably be to just remove all the directories with previous names too in `Cache::clear`.

The cleanest approach would be to give the directory a custom name like "pages-cache" or so. This way, there would be less coupling between the downloading and lookup. We could keep the migration code for a year or so and remove it again later

In any case, the migration code should be another PR (tealdeer still works with just this PR)

invakid404: @niklasmohrin is it okay if I leave the PR as is then? I'm having a hiccup with my rust setup lol

niklasmohrin: If CI passes, I am fine with merging :)

dbrgn: @niklasmohrin Good point. The migration logic could work like this:

- If the cache is missing (in the `None` branch of `check_cache` in `main.rs`), before showing the warning, first check whether the old cache directory exists, but not the new one. If that is the case, rename the old directory to the new name and log the migration on stderr.
- When explicitly updating the cache, check whether the old directory exists. If yes, simply remove it. This will also handle the cases where both the old and the new cache exist.

Then proceed as before. This way, the check will only happen if the cache is missing, so we don't have a performance penalty.

This can be done in a new PR though.

invakid404: I ran `cargo fmt`, clippy had nothing to say, so things should be fine.

niklasmohrin (approved):

invakid404:
I've put together a patch for the cache migration to experiment with in Funtoo. I'll append it to this comment. I'll update the PR if you find it okay-ish.

I've basically added a check to see if the cache directory is empty. If not, I move whatever is in it to "tldr-main". The second point, which is clearing the old directory on update, appears to not be needed, since the entire cache dir is being wiped on update right now.

```patch
From dead15f547d17858958c1ad61c629013d50f009d Mon Sep 17 00:00:00 2001
From: invakid404 <invakid404@riseup.net>
Date: Thu, 23 Sep 2021 13:33:54 +0300
Subject: [PATCH] chore: download main instead of master

---
 src/cache.rs |  7 ++++---
 src/main.rs  | 22 ++++++++++++++++------
 tests/lib.rs |  9 +++++----
 3 files changed, 25 insertions(+), 13 deletions(-)

diff --git a/src/cache.rs b/src/cache.rs
index 0e2ea2c..096cccd 100644
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -21,6 +21,7 @@ use crate::{
 };
 
 static CACHE_DIR_ENV_VAR: &str = "TEALDEER_CACHE_DIR";
+pub static TLDR_PAGES_DIR: &str = "tldr-main";
 
 #[derive(Debug)]
 pub struct Cache {
@@ -163,7 +164,7 @@ impl Cache {
     /// Return the duration since the cache directory was last modified.
     pub fn last_update() -> Option<Duration> {
         if let Ok((cache_dir, _)) = Self::get_cache_dir() {
-            if let Ok(metadata) = fs::metadata(cache_dir.join("tldr-master")) {
+            if let Ok(metadata) = fs::metadata(cache_dir.join(TLDR_PAGES_DIR)) {
                 if let Ok(mtime) = metadata.modified() {
                     let now = SystemTime::now();
                     return now.duration_since(mtime).ok();
@@ -217,7 +218,7 @@ impl Cache {
 
         // Get cache dir
         let cache_dir = match Self::get_cache_dir() {
-            Ok((cache_dir, _)) => cache_dir.join("tldr-master"),
+            Ok((cache_dir, _)) => cache_dir.join(TLDR_PAGES_DIR),
             Err(e) => {
                 log::error!("Could not get cache directory: {}", e);
                 return None;
@@ -263,7 +264,7 @@ impl Cache {
     pub fn list_pages(&self) -> Result<Vec<String>, TealdeerError> {
         // Determine platforms directory and platform
         let (cache_dir, _) = Self::get_cache_dir()?;
-        let platforms_dir = cache_dir.join("tldr-master").join("pages");
+        let platforms_dir = cache_dir.join(TLDR_PAGES_DIR).join("pages");
         let platform_dir = self.get_platform_dir();
 
         // Closure that allows the WalkDir instance to traverse platform
diff --git a/src/main.rs b/src/main.rs
index 59986a1..d200b92 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,7 +15,7 @@
 #![allow(clippy::similar_names)]
 #![allow(clippy::too_many_lines)]
 
-use std::{env, path::PathBuf, process};
+use std::{env, fs, path::PathBuf, process};
 
 use ansi_term::{Color, Style};
 use app_dirs::AppInfo;
@@ -35,7 +35,7 @@ mod output;
 mod types;
 
 use crate::{
-    cache::{Cache, PageLookupResult},
+    cache::{Cache, PageLookupResult, TLDR_PAGES_DIR},
     config::{get_config_dir, get_config_path, make_default_config, Config, MAX_CACHE_AGE},
     error::TealdeerError::ConfigError,
     extensions::Dedup,
@@ -50,7 +50,7 @@ const APP_INFO: AppInfo = AppInfo {
 };
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const USAGE: &str = include_str!("usage.docopt");
-const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/master.tar.gz";
+const ARCHIVE_URL: &str = "https://github.com/tldr-pages/tldr/archive/main.tar.gz";
 #[cfg(not(target_os = "windows"))]
 const PAGER_COMMAND: &str = "less -R";
 
@@ -118,8 +118,18 @@ fn check_cache(args: &Args, enable_styles: bool) {
         }
         Some(_) => {}
         None => {
-            eprintln!("Cache not found. Please run `tldr --update`.");
-            process::exit(1);
+            if let Ok((cache_dir, _)) = Cache::get_cache_dir() {
+                if let Some(old_pages_dir) = cache_dir.read_dir().into_iter().flatten().next() {
+                    let old_pages_dir = old_pages_dir.unwrap().path();
+                    let new_pages_dir = old_pages_dir.clone().with_file_name(TLDR_PAGES_DIR);
+
+                    fs::rename(old_pages_dir, new_pages_dir)
+                        .expect("failed to move existing cache dir");
+                } else {
+                    eprintln!("Cache not found. Please run `tldr --update`.");
+                    process::exit(1);
+                }
+            }
         }
     };
 }
@@ -192,7 +202,7 @@ fn show_paths() {
     let pages_dir = Cache::get_cache_dir().map_or_else(
         |e| format!("[Error: {}]", e),
         |(mut path, _)| {
-            path.push("tldr-master");
+            path.push(TLDR_PAGES_DIR);
             path.push(""); // Trailing path separator
             path.into_os_string()
                 .into_string()
diff --git a/tests/lib.rs b/tests/lib.rs
index 2926c75..cbe3186 100644
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -16,6 +16,7 @@ use tempfile::{Builder, TempDir};
 
 // TODO: Should be 'cache::CACHE_DIR_ENV_VAR'. This requires to have a library crate for the logic.
 static CACHE_DIR_ENV_VAR: &str = "TEALDEER_CACHE_DIR";
+static TLDR_PAGES_DIR: &str = "tldr-main";
 
 struct TestEnv {
     pub cache_dir: TempDir,
@@ -60,7 +61,7 @@ impl TestEnv {
         let dir = self
             .cache_dir
             .path()
-            .join("tldr-master")
+            .join(TLDR_PAGES_DIR)
             .join("pages")
             .join(os);
         create_dir_all(&dir).unwrap();
@@ -202,7 +203,7 @@ fn test_quiet_old_cache() {
         .stdout(is_empty());
 
     filetime::set_file_mtime(
-        testenv.cache_dir.path().join("tldr-master"),
+        testenv.cache_dir.path().join(TLDR_PAGES_DIR),
         filetime::FileTime::from_unix_time(1, 0),
     )
     .unwrap();
@@ -309,7 +310,7 @@ fn test_show_paths() {
             testenv
                 .cache_dir
                 .path()
-                .join("tldr-master")
+                .join(TLDR_PAGES_DIR)
                 .to_str()
                 .unwrap(),
         )));
@@ -534,7 +535,7 @@ fn test_autoupdate_cache() {
         .stderr(contains("Cache not found. Please run `tldr --update`."));
 
     let config_file_path = testenv.config_dir.path().join("config.toml");
-    let cache_file_path = testenv.cache_dir.path().join("tldr-master");
+    let cache_file_path = testenv.cache_dir.path().join(TLDR_PAGES_DIR);
 
     // Activate automatic updates, set the auto-update interval to 24 hours
     let mut config_file = File::create(&config_file_path).unwrap();
-- 
2.33.0
```

invakid404:
@niklasmohrin @dbrgn please let me know if you're willing to move forward with the patch from the comment above^

dbrgn:
Thanks for the update, @invakid404!

If you want to do a migration in the scope of this pull request, it's easier if you simply add and push another commit, because then we can comment the changes directly inline.

> I've basically added a check to see if the cache directory is empty. If not, I move whatever is in it to "tldr-main".

I feel that "renaming the first directory inside the cache directory to tldr-main" seems a bit brittle/dangerous. I think I'd prefer it if you'd explicitly check for the existence of a `tldr-master` directory within the cache directory (you can hardcode the name for the migration). What do you think, @niklasmohrin?

Additionally, what happens if the cache directory - for some reason - contains both a `tldr-main` and a `tldr-master` directory? With your patch this will panic, right? (This is not something I expect to happen in practice, but the code should still be robust, especially since this is an automatic migration that will run for all users.) I think in that case the old directory could be removed.

There's another issue with the patch: Right now the `check_cache` function will simply return with success if the cache directory was renamed, but we don't know whether the cache is up to date or not (because the `Cache::last_update()` function is not called again). I _think_ renaming a directory will not influence its mtime, so calling `check_cache` again recursively should work. In case I'm wrong and the mtime _is_ changed by the rename, then this will always assume an up-to-date cache even if it's outdated. We could avoid this by checking the mtime of `tldr-main/pages/` instead of checking the root directory directly.

> The second point, which is clearing the old directory on update, appears to not be needed, since the entire cache dir is being wiped on update right now.

Ah, true, good point. I think in that case we don't need any explicit cleanup.

invakid404:
> Thanks for the update, @invakid404!
> 
> If you want to do a migration in the scope of this pull request, it's easier if you simply add and push another commit, because then we can comment the changes directly inline.
> 
> > I've basically added a check to see if the cache directory is empty. If not, I move whatever is in it to "tldr-main".
> 
> I feel that "renaming the first directory inside the cache directory to tldr-main" seems a bit brittle/dangerous. I think I'd prefer it if you'd explicitly check for the existence of a `tldr-master` directory within the cache directory (you can hardcode the name for the migration). What do you think, @niklasmohrin?
> 
> Additionally, what happens if the cache directory - for some reason - contains both a `tldr-main` and a `tldr-master` directory? With your patch this will panic, right? (This is not something I expect to happen in practice, but the code should still be robust, especially since this is an automatic migration that will run for all users.) I think in that case the old directory could be removed.
> 
> There's another issue with the patch: Right now the `check_cache` function will simply return with success if the cache directory was renamed, but we don't know whether the cache is up to date or not (because the `Cache::last_update()` function is not called again). I _think_ renaming a directory will not influence its mtime, so calling `check_cache` again recursively should work. In case I'm wrong and the mtime _is_ changed by the rename, then this will always assume an up-to-date cache even if it's outdated. We could avoid this by checking the mtime of `tldr-main/pages/` instead of checking the root directory directly.
> 
> > The second point, which is clearing the old directory on update, appears to not be needed, since the entire cache dir is being wiped on update right now.
> 
> Ah, true, good point. I think in that case we don't need any explicit cleanup.

I'll update the PR, I just wasn't sure whether it's okay to do it in this PR. As to whether it's safe to move the first file to tldr-main, assuming things are in a consistent state, it should be fine. At first I thought about doing some sort of glob along the lines of `tldr-*` and performing some sort of merge on those, but I thought it might be overkill. I'll call check_cache recursively though, makes sense.

invakid404:
I think we might have a problem though, as I think moving a directory tampers with its date modified, at least on Linux. Might have to do things in a different order.

invakid404:
I've just updated the branch, I'll try to come up with a solution for the date modified stuff.

invakid404:
I've committed a fix for the date modified stuff. I now get the date modified before doing any moving, and I pass it as an argument when calling check_cache recursively. Appears to work on my machine. Simulating a case where I'd need migration and my cache is stale with `mv tldr-main tldr-master && touch -d "2 months ago" tldr-master` and running `tealdeer --list` performs the migration and still tells me my cache is stale as it should be.

niklasmohrin (changes requested):
Revoking approval until discussion is finished

niklasmohrin:
> I feel that "renaming the first directory inside the cache directory to tldr-main" seems a bit brittle/dangerous.

Agree. Why exactly are we keeping the old directory again? Can't we simply keep a list of "deprecated paths" or so that we always delete if we find them (+notify the user / ask them?)? We will overwrite the old dir anyways when we update, right?

> There's another issue with the patch: Right now the check_cache function will simply return with success if the cache directory was renamed,

Shouldn't that be resolved by splitting the cache-updating method into two parts:
1. Check for and apply migrations.
2. Download pages and place them at the right place.

What do you think?

invakid404:
> Agree. Why exactly are we keeping the old directory again? Can't we simply keep a list of "deprecated paths" or so that we always delete if we find them (+notify the user / ask them?)? We will overwrite the old dir anyways when we update, right?

We don't keep the old directory. The problem is, as far as I understand it, that my current implementation assumes that the cache dir will contain a single directory. I've made it find the first directory and move it to the "new" cache location. That was pointed out as potentially dangerous, although I did it to avoid hardcoding "tldr-master", and under the assumption the cache directory was consistent in the first place. I could modify that if you think my reasoning isn't sufficient.

> Shouldn't that be resolved by splitting the cache-updating method into two parts:

> Check for and apply migrations.
> Download pages and place them at the right place.


I've resolved the `check_cache function returning success if cache directory was renamed` by calling check_cache recursively with the date modified of the "old" cache directory.

niklasmohrin:
I am going to be upfront about this: I don't like the current implementation, it doesn't feel right to me. Generally speaking, I wouldn't assume that a function called `check_cache_with_last_update` moves directories around. The recursion seems inappropriate to me too. I don't know, we recurse at most once, this makes it more complicated than it really is. I believe we should go for something like this:

```rust
let mut last_update = when_was_last_update();
if we_have_to_migrations {
  do_migrations();
  last_update = when_was_last_update();
}
// check if cache is not recent enough
```

That said, I don't think the code for this belongs in the method it is currently in. I think that the signature of the new method is too complicated: an `Option` is passed and the `None` case means that something beforehand failed, but the method has to fix that. Really, we should make the caller of the method handle the case that there is no data.

I see by your comment that your approach might work well, but I still don't get why it is superior to / why we haven't explored the "delete all 'deprecated' files" approach.

Let me know what you think / if I missed anything :)

invakid404:
My implementation was following what @dbrgn said initially:

> If the cache is missing (in the None branch of check_cache in main.rs), before showing the warning, first check whether the old cache directory exists, but not the new one. If that is the case, rename the old directory to the new name and log the migration on stderr.

The whole moving folders around thing is to avoid having to redownload tldr-pages unless explicitly told to do so, as far as I understand it. If you have concrete suggestions on what and how to improve, I'm more than happy to change stuff. I just put it together as a thing that works and my approach is by no means final.

invakid404:
> Really, we should make the caller of the method handle the case that there is no data.

check_cache is being called in multiple places iirc, and I think we always want to try and do the migration when we check the cache, so I don't see how that'd help our case. As to the recursion stuff, it was again dbrgn who suggested that calling the function recursively might work. The thing with the "confusing argument" is to allow me to get the modified date before performing any moving. I personally don't find it confusing, but if you have a better idea, I'm down to refactor stuff. I'm by no means a rust wizard and I'm not really familiar with the codebase, so I wouldn't know the best way and place to do all this stuff.

niklasmohrin:
>  avoid having to redownload tldr-pages unless explicitly told to do so

Ah, I see. I think it would be okay to force users to update the cache after updating tealdeer. That would simplify this a lot, we could just do the migration at the one code site that does the update. What are your opinions @invakid404 @dbrgn ?

---

In any case, the migration code should be (more or less) separated from anything else. Then, we can decide to call the migrations from anywhere in the program. I could see that we just have one call to `perform_all_migrations` or so in `main` at some point - the code that checks whether the cache is outdated shouldn't have to know about migrations. 

This is of course only what I think would be best for keeping everything tidy and decoupled. The suggestion from above would also get rid of the recursion and the `Option` argument. What do you think?

---

Have we settled on what to call the cache directory yet? I think that we could use this opportunity to decouple the directory name of the downloaded file from how our xdg cache is structured by just calling the directory `pagecache` or so. It would make it harder for outside inspectors to tell where the files in the cache are coming from - not really a critical loss in my opinion.

dbrgn:
> On decoupling the check_cache function and migrations logic, I believe dbrgn's argument for that was avoiding any significant performance hit. I believe that's why we only try to migrate if check_cache found no cache directory in the first place.

Yes, that was the idea, we should not run migration code on every invocation, only if the cache is missing.

I thought about this for a while, and wasn't sure if a correct solution is possible without adding too much complexity. I then just tried to implement the change myself (that's easier than explaining in words what I mean), and I think I found a good approach (essentially what I originally wrote, but that required some refactoring to enable a clean implementation).

- First, we need to check the cache. By refactoring the cache checking code, we can avoid the recursion.
- Then, once we know that the cache is missing, we need to check for the old directory. In my view, the directory names can be hardcoded.
- The third aspect: We need to know the freshness of the cache after renaming it. I checked and under Linux (at least with ext4), renaming a directory does _not_ change its mtime. And according to [this SO question](https://stackoverflow.com/questions/1025187/rules-for-date-modified-of-folders-in-windows-explorer) the same thing holds true for Windows as well. This means we can rename the folder and don't need to care about its mtime, we simply re-check the freshness after renaming.

I have implemented the cache check refactorings and the migration logic in the `cache-check-refactoring` branch: https://github.com/dbrgn/tealdeer/compare/cache-check-refactoring

@niklasmohrin @invakid404 can you take a look? Do you agree with that approach?

If yes, @invakid404, could you do your renaming changes on top of that branch? (Let me know if you need help with git!)

dbrgn:
By the way, I just noticed that the officially recommended way to download the pages is through https://tldr.sh/assets/tldr.zip (see https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#caching). This ZIP does not contain any top-level directory.

If we use this download URL instead of the github one, we do not need to care about the main branch name. Additionally, it contains fewer files than the old checkout.

My suggestion would be to use this occasion to migrate to the new download URL.

- From now on, put the pages into `<cache_dir>/pages/` (previously: `<cache_dir>/tldr-master/`)
- Download the files from https://tldr.sh/assets/tldr.zip
- When clearing the cache, simply delete the `<cache_dir>/pages/` dir
- Don't do any directory-renaming migration (ignore that commit from my branch), but check for an old `tldr-master` checkout in the cache whenever the cache is explicitly updated (with `--update`). This code can then be removed in a future version. If someone upgrades tealdeer, there will be a "cache missing" warning (that's fine).

@invakid404 do you have time in the coming days to implement this, or should I try to implement the change myself?

invakid404:
I do have time, but I'm a tiny bit confused right now :D. Do you want me to change the base of this PR to be your cache-check-refactoring branch still?

dbrgn:
Yes 🙂

I think the easiest way forward would be if you'd start a new branch from our `master`, then `git cherry-pick 20ea5e607a452e707c62f73722978f418e3f469a` to get the first commit from my branch.

Then, re-add the additional logic from your current branch (but this time with the new download URL) and create a new PR. This one already got quite large/confusing 😄

Does that make sense?

invakid404:
Yeah, sure, I can deal with git. I just want to make sure I understood your final suggestion: we need to download from the link you posted, and instead of doing cache migration checks in check_cache, we'll only check for an existing `tldr-master` folder when explicitly updating the cache, and moving it if present? Or should we just ditch the idea of a migration completely and wipe the cache directory clean?

invakid404:
Superseded by #213.

PR #208, opened 2021-10-03
chore: add funtoo package to list

invakid404:
Hello 👋🏻,

I maintain the tealdeer package in Funtoo Linux, so I thought I might as well mention its existence in the docs :^)


niklasmohrin (approved):
Thank you!

We had a discussion with other package maintainers a while back about the Minimum Supported Rust Version (MSRV) at #190. Do you have any concerns that we should take into account? I know compiling Rust takes an eternity. Feel free to share your thoughts in the linked thread :)

invakid404:
> Thank you!
> 
> We had a discussion with other package maintainers a while back about the Minimum Supported Rust Version (MSRV) at #190. Do you have any concerns that we should take into account? I know compiling Rust takes an eternity. Feel free to share your thoughts in the linked thread :)

Over at Funtoo, our preferred way of installing rust is via the official prebuilt binaries. Pretty sure Gentoo has made the same decision. We have a rust binary package, which is automatically updated when a new official release drops, so we shouldn't have issues in that regard :)

dbrgn (approved):
Thanks @invakid404!

Issue #156, opened 2020-11-29
Add cache directory (or tldr checkout directory) to config


slippycheeze:
G'day.  I use a couple different tldr clients, and so rather than maintaining multiple caches I just keep my own git checkout around and updated regularly.

I can get tealdeer to use that by way of `export TEALDEER_CACHE_DIR="${HOME}/var/cache"` and naming the checkout `tldr-master` under it.  Which is kind of ugly, but whatever.

It'd be great if I could just add `cache = ~/var/cache` (or, better, `tldr = ~/var/cache/tldr`) to the config.toml and never have to worry about the environment variable again.  (in the second example, that'd be the path to the directory with the content, so that you didn't need append anything to the directory to get the page root.)

niklasmohrin:
We could also consider just unpacking the tldr repo directly into the `TEALDEER_CACHE_DIR`. This is only an option, if we don't need to put anything else in the cache, but I wouldn't know what (caching of rendered pages for performance incoming? 😆).

hgaiser:
Hopefully https://github.com/dbrgn/tealdeer/issues/156 will fix this issue :).

dbrgn:
Right now, the cache directory is expected to be a cache under the control of tealdeer. This is shown for example in the PR #206, where we are adding a migration that renames the checkout directory, and - if both exist - deletes the old directory.

If we add the option to externally manage the tldr repo checkout, then enabling this feature should disable all cache management functionality in tealdeer to avoid conflicts. However, that shouldn't be the same as configuring the cache directory. The directory layout inside the cache directory is _not_ an API, and I do not want to guarantee a specific structure.

I do agree that `TEALDEER_CACHE_DIR` is a bit of an outlier. Initially we just had this env variable to configure tealdeer, and no config file. Then we added the config file. Now that this config file exists, there's no reason why the cache directory should be configured through an env var, and not through the config.

TLDR:

- Making the cache directory configurable through the config file makes sense. The env var should then be deprecated (and a deprecation warning should be shown if it's still set).
- We do not guarantee any specific layout inside this cache directory, and tealdeer may modify/rename/delete files inside that directory at any time. Pointing the cache directory to a shared dir is risky.
- If people would like to manage the tldr pages themselves, this should be a separate option and it should disable all cache management functionality. However, I suspect that this would add significant complexity to the sourcecode, and I'd be reluctant to merge such a PR.

niklasmohrin:
This was implemented in #276 :)

[...]

PR #212, opened 2021-10-16
Add cache dir config in configuration file.

hgaiser:
This PR started as just adding a `cache_dir` option to the configuration file, but in the process I had to refactor some functions.

Mainly I moved some functionality for the cache inside the cache struct and let it own a `PathBuf` to the cache dir.

This fixes https://github.com/dbrgn/tealdeer/issues/156 .

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -14,9 +14,9 @@ use walkdir::{DirEntry, WalkDir};
 use zip::ZipArchive;
 
 use crate::{
+    config::MAX_CACHE_AGE,
     error::TealdeerError::{self, CacheError, UpdateError},
     types::OsType,
-    config::MAX_CACHE_AGE,
 };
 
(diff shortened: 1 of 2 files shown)


niklasmohrin:
@hgaiser Do you want to keep working on this, or can we reassign this if someone else comes along? (Both options are equally valid)

hgaiser:
Ah sorry, I don't think I'll have time to work on this anytime soon. Feel free to reassign to someone else 👍

hgaiser:
Yep makes sense! Sorry for not following up on this.

niklasmohrin:
Note that I also have some thoughts in #266

dbrgn:
Replaced by #276.

dbrgn:
> Note that I also have some thoughts in https://github.com/dbrgn/tealdeer/pull/266

Aaaah, I never saw that! Sorry very much about that 😞 I'll take a look at it, to see if we independently got to the same conclusions, or whether we can merge our approaches.

PR #213, opened 2021-10-16
chore: download proper tldr archive

invakid404:
According to the [tldr client specifications](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#caching), downloading https://tldr.sh/assets/tldr.zip is preferred over directly downloading an archive of the master branch.

To achieve this, I've replaced the flate2 crate with the zip crate, and have changed the download URL accordingly. Also, instead of wiping the entire cache dir, I've made `Cache::clear()` only delete directories that are known to contain a tldr-pages archive, i.e. `tldr-master` (the old dir) and `tldr-pages` (the new dir). In the future, we could/should drop the part which deals with the old directory.

This PR also includes changes from @dbrgn's cache refactoring branch.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -340,10 +340,11 @@ impl Cache {
                 let pages_dir = path.join(pages_dir_name);
 
                 if pages_dir.exists() {
-                    fs::remove_dir_all(&pages_dir).map_err(|_| {
+                    fs::remove_dir_all(&pages_dir).map_err(|e| {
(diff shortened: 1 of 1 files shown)


dbrgn:
I see what you did to those commit IDs 😂

dbrgn:
Merged in f7da0b028d571b5232590d303783ad90d84ac373

PR #215, opened 2021-10-17
Add "macos" alias for "osx"

dbrgn:
The Apple computer operating system was renamed from "OS X" to "macOS" in 2016. We should accept "macos" in addition to "osx" in all our CLI APIs, and the docs should also mention this alias instead of "osx".

For backwards compatibility and for compatibility with the tldr client specification, we should still accept "osx" though.

### DIFF
--- a/bash_tealdeer
+++ b/bash_tealdeer
@@ -14,7 +14,7 @@ _tealdeer()
 			return
 			;;
 		-p|--platform)
-			COMPREPLY=( $(compgen -W 'linux osx sunos windows' -- "${cur}") )
+			COMPREPLY=( $(compgen -W 'linux macos sunos windows' -- "${cur}") )
 			return
 			;;
 		--color)
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -13,10 +13,9 @@ ARGS:
 OPTIONS:
     -l, --list                   List all commands in the cache
     -f, --render <FILE>          Render a specific markdown file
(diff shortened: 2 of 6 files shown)


niklasmohrin:
Will you run this by tldr-pages/tldr or should I?

dbrgn:
Would be great if you could do that!

dbrgn:
@niklasmohrin should we wait with this until there's progress in https://github.com/tldr-pages/tldr/issues/7076?

niklasmohrin:
I am not really expecting any new input from the upstream issue unless we actively engage again (and that is totally fair). I guess we could ask tldr maintainers if they want to conclude anything from the current comments and make a decision. Should we?

dbrgn:
@niklasmohrin I pinged them in the chat. I think we can still add the alias in the next version, it doesn't hurt and it's highly unlikely that a spec change will be conflicting.

I'll rebase this PR, so we could merge it for the 1.6 release if we decide to do so.

dbrgn:
Rebased. I also found an option to hide the "possible values" for the `-o` parameter.

I think we should merge this for the 1.5 release.

niklasmohrin:
> Rebased. I also found an option to hide the "possible values" for the `-o` parameter.

Nice. We could also hide the option altogether, don't know if we should though

> I think we should merge this for the 1.5 release.

:+1:

niklasmohrin (approved):

Issue #209, opened 2021-10-08
Spec compliance (`--platform`)


pixelcmtd:
We recently had some internal misconceptions (tldr-pages/tldr#6869) because of clients, especially this one, not complying to the spec by using `-o`/`--os` instead of `-p`/`--platform`. Is that only because of `--pager` and are you planning to fix it?

Things to change in order to conform to the client specification:

- [x] Rename `-o`/`--os` to `-p`/`--platform` → #217 
- [x] When doing a page lookup, ensure that the page name must be lowercased (`tldr AbC should display the page called `abc`) → #227
- [ ] ~~Support the platform `all` which works like the default platform, but lists pages from all platforms when calling `--list` → #229~~

CleanMachine1:
:+1: for fitting spec, makes documentation easier and everyone being able to use the same provided information.

niklasmohrin:
Should we convert your list to issues with a custom label / milestone?

dbrgn:
I decided to drop the issue with the platform type "all" from the spec compliance list, since its place in the spec is disputed (see https://github.com/tldr-pages/tldr/issues/7528 and https://github.com/tldr-pages/tldr/pull/7561). We'll keep tracking those issues, but the compliance work for the client specification v1.5 is done here, I think.

PR #217, opened 2021-10-17
Rename `-o / --os` to `-p / --platform`

dbrgn:
This makes it compliant with the client spec (see #209).

The `-p` flag was previously used for the pager, using the pager now requires using the long flag (`--pager`). Alternatively the pager can be configured through the config file.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,7 +40,7 @@ use crate::{
     error::TealdeerError::ConfigError,
     extensions::Dedup,
     output::print_page,
-    types::{ColorOptions, OsType},
+    types::{ColorOptions, PlatformType},
 };
 
 const NAME: &str = "tealdeer";
@@ -82,12 +82,21 @@ struct Args {
 
     /// Override the operating system
     #[clap(
+        short = 'p',
+        long = "platform",
+        requires = "command",
+        possible_values = ["linux", "osx", "sunos", "windows"],
+    )]
+    platform: Option<PlatformType>,
+
+    /// Deprecated alias of `platform`
+    #[clap(
         short = 'o',
         long = "os",
         requires = "command",
-        possible_values = ["linux", "osx", "sunos", "windows"]
+        possible_values = ["linux", "osx", "sunos", "windows"],
     )]
-    os: Option<OsType>,
+    os: Option<PlatformType>,
 
     /// Override the language
     #[clap(short = 'L', long = "language")]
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -344,6 +348,10 @@ fn main() {
             args.raw = true;
             eprintln!("Warning: The -m / --markdown flag is deprecated, use -r / --raw instead");
         }
+        if args.os.is_some() && args.platform.is_none() {
+            args.platform = args.os;
+            eprintln!("Warning: The -o / --os flag is deprecated, use -p / --platform instead");
+        }
```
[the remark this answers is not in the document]
niklasmohrin: Yeah, that makes sense

src/main.rs
```diff
@@ -78,8 +78,12 @@ struct Args {
     render: Option<String>,
 
     /// Override the operating system [linux, osx, sunos, windows]
+    #[clap(short = 'p', long = "platform", requires = "command")]
```
tranzystorek-io, line 81: Maybe add a `short_alias` and `alias` to `-o`/`--os` instead of having two separate fields

tranzystorek-io: On second thought, that would stand in the way of the deprecation message...

src/main.rs
```diff
@@ -412,9 +425,9 @@ fn main() {
     }
 
     // Specify target OS
-    let os: OsType = match args.os {
-        Some(os) => os,
-        None => OsType::current(),
+    let platform: PlatformType = match args.platform {
+        Some(platform) => platform,
+        None => PlatformType::current(),
     };
```
niklasmohrin, line 431:
```suggestion
    let platform = args.platform.unwrap_or_else(PlatformType::current);
```

src/main.rs
```diff
@@ -356,6 +365,10 @@ fn main() {
             args.raw = true;
             eprintln!("Warning: The -m / --markdown flag is deprecated, use -r / --raw instead");
         }
+        if args.os.is_some() && args.platform.is_none() {
```
niklasmohrin, line 368: But shouldn't we print the warning all the time? Might be confusing if passing both doesn't trigger any alarms. How about something like:
```rust
if args.os.is_some() {
  // print the warning
}
args.platform = args.platform.or(args.os);
```

src/main.rs
dbrgn: Yes, of course 😄

src/main.rs
dbrgn: Ah, you're right, that makes sense.

niklasmohrin (approved):

PR #223, opened 2021-10-24
Add helper functions: print_warning / print_error

dbrgn:
These functions prepend a consistend prefix and they also use colors (when enabled).

Messages are written to stderr.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,6 @@
 
 use std::{env, path::PathBuf, process};
 
-use ansi_term::{Color, Style};
 use app_dirs::AppInfo;
 use atty::Stream;
 use clap::{AppSettings, Parser};
@@ -33,6 +32,7 @@ mod formatter;
 mod line_iterator;
 mod output;
 mod types;
+mod utils;
 
 use crate::{
     cache::{Cache, CacheFreshness, PageLookupResult, TLDR_PAGES_DIR},
@@ -41,6 +41,7 @@ use crate::{
     extensions::Dedup,
     output::print_page,
     types::{ColorOptions, PlatformType},
+    utils::{print_error, print_warning},
 };
 
 const NAME: &str = "tealdeer";
@@ -49,8 +50,6 @@ const APP_INFO: AppInfo = AppInfo {
     author: NAME,
 };
 const ARCHIVE_URL: &str = "https://tldr.sh/assets/tldr.zip";
-#[cfg(not(target_os = "windows"))]
-const PAGER_COMMAND: &str = "less -R";
 
 // Note: flag names are specified explicitly in clap attributes
 // to improve readability and allow contributors to grep names like "clear-cache"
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -0,0 +1,22 @@
...
+/// message will be printed.
+pub fn print_warning(enable_styles: bool, message: &str) {
+    print_msg(enable_styles, message, "Warning: ", Color::Yellow);
+}
+
+/// Print an error to stderr. If `enable_styles` is true, then a red message
+/// will be printed.
+pub fn print_error(enable_styles: bool, message: &str) {
+    print_msg(enable_styles, message, "Error: ", Color::Red);
+}
```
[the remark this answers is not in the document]
niklasmohrin: I don't think that a function named `print_*` should exit the process, there could be one like called `die` or so that uses the `print_critical` function

niklasmohrin (commented):
I am not sure whether this is the right move. Having the function like this means that we need to use `format!` at call sites instead of having the flexibility of a `eprintln` or so. I would prefer using a library for this use case, there is the `log` crate and the [`env_logger`](https://crates.io/crates/env_logger) crate for executables. These can also be configured once at the beginning of the program , so that we don't have to pass `enable_styles` around all the time. What do you think?

src/utils.rs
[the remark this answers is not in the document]
dbrgn: True. Let's skip that for now.

dbrgn:
Hm, I think logging and regular program output should be separate. We don't use an info-level logger to print the tldr pages either. For me, logging should be about "what the program is doing", and not be part of an end user API. On the other hand, the warning about missing cache is not something that should be hidden by a logger reconfiguration.

> Having the function like this means that we need to use format! at call sites instead of having the flexibility of a eprintln or so.

Well, that's not really something we need an external library for. We can easily convert those functions to a macro. However, in the interest of getting out a release soon, I'd postpone that for now 🙂

src/main.rs
```diff
@@ -146,13 +145,13 @@ struct Args {
...
-fn configure_pager() {
-    Pager::with_default_pager(PAGER_COMMAND).setup();
+fn configure_pager(_: bool) {
+    Pager::with_default_pager("less -R").setup();
 }
 
 #[cfg(target_os = "windows")]
-fn configure_pager() {
-    eprintln!("Warning: -p / --pager flag not available on Windows!");
+fn configure_pager(enable_styles: bool) {
```
niklasmohrin, line 153: This one (and the general pattern of passing things like `Args`, `quietly` or `enable_styles` to functions) annoys me a bit. It's convoluted, configuring a pager shouldn't know about printing style. I think we would be better of with a `OutputManager` or so that gets passed as a whole and has `print_error` and `print_warning` as methods. This could be expanded to a struct `ProgramState` or similar that also has the cli arguments and can tell callers things about them - without them having to look at the flags themselves. (I am not sure whether the latter would work out to be useful in practice though.) This is all, of course, nothing for this PR but something that could be done sometime after the next release.

niklasmohrin (approved):
Makes sense :)

src/main.rs
dbrgn: > This one (and the general pattern of passing things like Args, quietly or enable_styles to functions) annoys me a bit. It's convoluted, configuring a pager shouldn't know about printing style.

I fully agree! But I didn't really know how to easily solve it in a better way, and I think the new code is still better than the old.

niklasmohrin: Agreed, let's just leave the comment open and merge this :)

PR #227, opened 2021-12-04
Lowercase page names before lookup

dbrgn:
This adds another allocation that wasn't necessary before, but the overhead is totally negligible.

Part of #209.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -493,7 +493,10 @@ fn main() {
 
     // Show command from cache
     if !args.command.is_empty() {
-        let command = args.command.join("-");
+        // Note: According to the TLDR client spec, page names must be transparently
+        // lowercased before lookup:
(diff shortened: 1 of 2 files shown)


niklasmohrin:
I wonder why they don't have a `make_lowercase` 🤔 I agree, the one allocation doesn't matter. However, I think it would be nice to have the code that handles these what-users-enter-vs-what-files-we-are-looking-for logic in one place. Maybe this change here is better fit right after the code where the words are joined with `'-'`? What do you think?

dbrgn:
> I wonder why they don't have a `make_lowercase` 🤔

You mean in-place? That won't work, because it might be possible that the lowercased string requires more bytes than the uppercased string (depending on some unicode edge cases). It's not all ASCII 🙂

> Maybe this change here is better fit right after the code where the words are joined with `'-'`? What do you think?

Yeah, that would actually make sense, then the user sees the real page name that was used for the lookup in the error message (if the page wasn't found).

PR #229, opened 2021-12-05
Support platform type "all"

dbrgn:
This is the third part of #209.

The goal is supporting the special `all` platform that results in pages
for all platforms being listed when calling `--list`. It's part of the
tldr client specification.

However, `All` should not be a variant of the `PlatformType` enum,
because `Current` isn't a `PlatformType` either. Thus, we accept the
string `all` but convert it into the current platform when parsing.

For consistency, the same is done when no platform is specified, by
introducing yet another possible value `current` which is used by
default. This way, we get rid of the `Option`.

Due to limitations in clap, we had to remove the `requires = "command"`
requirement, however that's actually a good thing because `--list` does
not require a command either, but supports the `--platform` parameter.

To simplify handling of os / platform arguments, a conflict between
`--platform` and `--os` was introduced.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -16,7 +16,7 @@ use zip::ZipArchive;
 
 use crate::{
     error::TealdeerError::{self, CacheError, UpdateError},
-    types::{PathSource, PlatformType},
+    types::{PathSource, PlatformStrategy, PlatformType},
 };
 
 static CACHE_DIR_ENV_VAR: &str = "TEALDEER_CACHE_DIR";
@@ -27,7 +27,7 @@ static TLDR_OLD_PAGES_DIR: &str = "tldr-master";
 #[derive(Debug)]
 pub struct Cache {
     url: String,
-    platform: PlatformType,
+    platform: PlatformStrategy,
 }
 
 #[derive(Debug)]
@@ -64,7 +64,7 @@ pub enum CacheFreshness {
 }
 
 impl Cache {
-    pub fn new<S>(url: S, platform: PlatformType) -> Self
+    pub fn new<S>(url: S, platform: PlatformStrategy) -> Self
     where
         S: Into<String>,
     {
@@ -196,7 +196,7 @@ impl Cache {
 
     /// Return the platform directory.
     fn get_platform_dir(&self) -> &'static str {
-        match self.platform {
+        match self.platform.platform_type {
             PlatformType::Linux { .. } => "linux",
             PlatformType::OsX { .. } => "osx",
             PlatformType::SunOs { .. } => "sunos",
@@ -305,7 +305,7 @@ impl Cache {
         let mut pages = WalkDir::new(platforms_dir)
             .min_depth(1) // Skip root directory
             .into_iter()
-            .filter_entry(|e| self.platform.is_all() || should_walk(e)) // Filter out pages for other architectures
+            .filter_entry(|e| self.platform.list_all || should_walk(e)) // Filter out pages for other architectures
             .filter_map(Result::ok) // Convert results to options, filter out errors
             .filter_map(|e| {
                 let path = e.path();
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -197,10 +197,10 @@ impl Cache {
     /// Return the platform directory.
     fn get_platform_dir(&self) -> &'static str {
         match self.platform {
-            PlatformType::Linux => "linux",
-            PlatformType::OsX => "osx",
-            PlatformType::SunOs => "sunos",
-            PlatformType::Windows => "windows",
+            PlatformType::Linux { .. } => "linux",
+            PlatformType::OsX { .. } => "osx",
+            PlatformType::SunOs { .. } => "sunos",
+            PlatformType::Windows { .. } => "windows",
         }
     }
 
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -305,7 +305,7 @@ impl Cache {
         let mut pages = WalkDir::new(platforms_dir)
             .min_depth(1) // Skip root directory
             .into_iter()
-            .filter_entry(|e| should_walk(e)) // Filter out pages for other architectures
+            .filter_entry(|e| self.platform.is_all() || should_walk(e)) // Filter out pages for other architectures
             .filter_map(Result::ok) // Convert results to options, filter out errors
             .filter_map(|e| {
                 let path = e.path();
--- a/src/types.rs
+++ b/src/types.rs
@@ -78,6 +78,19 @@ impl PlatformType {
     pub fn current(all: bool) -> Self {
         Self::Other { all }
     }
+
+    /// Return whether or not the `all` flag is set.
+    ///
+    /// This flag is only relevant when listing pages: When `all` is set to
+    /// `true`, then the pages for all platforms should be listed.
+    pub fn is_all(self) -> bool {
+        match self {
+            Self::Linux { all }
+            | Self::OsX { all }
+            | Self::SunOs { all }
+            | Self::Windows { all } => all,
+        }
+    }
 }
 
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Deserialize)]
--- a/src/types.rs
+++ b/src/types.rs
@@ -8,19 +8,19 @@ use serde_derive::{Deserialize, Serialize};
 #[serde(rename_all = "lowercase")]
 #[allow(dead_code)]
 pub enum PlatformType {
-    Linux,
-    OsX,
-    SunOs,
-    Windows,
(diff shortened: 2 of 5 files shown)


src/types.rs
```diff
@@ -8,19 +8,19 @@ use serde_derive::{Deserialize, Serialize};
...
 #[allow(dead_code)]
 pub enum PlatformType {
-    Linux,
-    OsX,
-    SunOs,
-    Windows,
+    Linux { all: bool },
+    OsX { all: bool },
+    SunOs { all: bool },
+    Windows { all: bool },
```
niklasmohrin, line 14: Hm, that doesn't feel right, I don't like it :/. Couldn't we model it so there is an `enum Platform` (or `PlatformType`) which has linux, windows, etc. and then there is an `enum PlatformStrategy { All, Priority(Platform) }`? `find_page` could then reject `PlatformStrategy::All` or just stick to the "normal" `Platform` as an argument.

This ofc complicates parsing a bit, but also makes it closer to the actual spec again: We change the type for `--platform`  to `PlatformStrategy` and make it optional again. What do you think?

src/cache.rs
```diff
@@ -305,7 +305,7 @@ impl Cache {
         let mut pages = WalkDir::new(platforms_dir)
             .min_depth(1) // Skip root directory
             .into_iter()
-            .filter_entry(|e| should_walk(e)) // Filter out pages for other architectures
+            .filter_entry(|e| self.platform.is_all() || should_walk(e)) // Filter out pages for other architectures
```
niklasmohrin, line 308: This check / the closure _could_ move into a potential `PlatformStrategy` :D

niklasmohrin (changes requested):
Here I am, doing what I do best: Proposing architectural changes 😅

src/types.rs
dbrgn: We could, but then where do you do the `::current()` lookup?

I'm also not fond of the hack I used, but I wanted to avoid adding a wrapper type. However, if we'd add a wrapper type, why not a struct instead of an enum?

```rust
struct PlatformStrategy {
    platform: PlatformType,
    list_all: bool,
}
```

src/cache.rs
dbrgn: I don't think the platform strategy should deal with file and directory names.

src/types.rs
niklasmohrin: > We could, but then where do you do the ::current() lookup?

I would keep `PlatformType::current` as it is on `main` right now, just returning the correct variant.

> I'm also not fond of the hack I used, but I wanted to avoid adding a wrapper type. However, if we'd add a wrapper type, why not a struct instead of an enum?

Do we want to allow `--platform all` when actually looking for a page _to display_, or just for `--list`? I think the latter. Then, an enum is right. The code for `--list` accepts both variants (by having its argument be `PlatformStrategy`), whereas the normal path only takes a `PlatformType` as an argument. In both cases, if we allowed specifying the `list_all` flag _and_ a specific platform, we need to make choices about prioritization, which is just complication and not covered in the spec either. Or is there any benefit in still considering a specific platform even if `--platform all` was passed? I could totally be missing something :D

src/cache.rs
niklasmohrin: Fair

src/types.rs
dbrgn: > Do we want to allow --platform all when actually looking for a page to display, or just for --list? I think the latter.

Yes, at least that's what the spec says.

> Then, an enum is right. The code for --list accepts both variants (by having its argument be PlatformStrategy)

But the `PlatformStrategy` doesn't contain a "resolved" platform type when the variant is `All`.

The type "all" is not something that conflicts with a concrete platform type. It behaves like the "current" platform type (which is resolved in `main`). The only difference is that when listing the pages, all platforms are included. Otherwise it acts like "current".

The listing logic lives within the cache code. The cache code must be initialized with a platform type. Thus, if we only pass in `enum PlatformStrategy { All, Priority(Platform) }`, every part of the caching code that deals with the platform type must match on the strategy, and handle the error case `All` (because that doesn't contain a concrete platform type). In contrast, in the current implementation, everything just works as before.

I see two ways to implement the "all" platform type cleanly without the `{ all: bool }` enum hack:

- A: The current approach
- B: A wrapper type like this:

```rust
struct PlatformStrategy {
    platform: PlatformType,
    list_all: bool,
}
```

The `--platform` parameter type would be `PlatformStrategy`. The cache would be initialized with the `PlatformStrategy` instance (because that's where listing the pages is done), while every other part of the code that deals with the platform type would take a `PlatformType` as before.

niklasmohrin: Ah, we are designing two different interfaces right now 😅 

> Yes, at least that's what the spec says.

But where? This is what the requirement in our issue says, but in the client spec I can only find it mentioned once in the table where the flags are listed. I would think about `--list` as a subcommand rather than an addition to the main use case. It kind of is, the `--list` option triggers an `exit(0)` before we look into our current pages dir, so why bother resolving the platform for looking up pages? Am I not seeing something?

> But the PlatformStrategy doesn't contain a "resolved" platform type when the variant is All.

... which matches what I believe to be the correct interpretation

dbrgn: > But where? This is what the requirement in our issue says, but in the client spec I can only find it mentioned once in the table where the flags are listed.

Yes, you're right, this is not really specified right now. And you're right that we could accept `all` only for `--list`. However, that means that we have to conditionally accept some values for the `--platform` parameter, which might be possible to implement with clap (or with custom logic), but it adds a lot of complexity that isn't really needed. Also, do we put `all` in the autocompletion files for `--platform` or not?

> It kind of is, the --list option triggers an exit(0) before we look into our current pages dir, so why bother resolving the platform for looking up pages? Am I not seeing something?

The listing happens through the cache struct, which is initialized with a `PlatformType` right now. If we change this to be a `PlatformStrategy`, then every time we match on it (to access the actual platform type), we need to handle the case where it's `All`, but where we're not actually doing a command listing, but something else. However, we don't even know what the command was, that's not something that the cache should care about.

If we wanted an API like this, we'd need to separate the cache logic that might need access to the platform-specific pages from the cache logic that only lists all page. Otherwise we have functions like this:

```rust
/// Return the platform directory.
fn get_platform_dir(&self) -> &'static str {
    match self.platform_strategy {
        PlatformStrategy::Priority(PlatformType::Linux) => "linux",
        PlatformStrategy::Priority(PlatformType::OsX) => "osx",
        PlatformStrategy::Priority(PlatformType::SunOs) => "sunos",
        PlatformStrategy::Priority(PlatformType::Windows) => "windows",
        PlatformStrategy::All => panic!("Called get_platform_dir even though it was `All`, which should only be used for listing"),
    }
}
```

...or even worse, propagate this as an error, which means that the calling code also needs to care about this situation.

I like APIs where invalid state cannot be represented in the type system. The cache needs to know the platform type, and we _know_ the platform type (it's `PlatformType::current()`). So I don't see why we should avoid putting it in the `PlatformStrategy`. Thus, it should be a product type (i.e. a struct), not a sum type (i.e. an enum). The enum with `{ all: boolean }` was simply a hack to emulate a product type, which I agree isn't too clean.

dbrgn: Btw, I'll rewrite it for variant B, I agree that it's cleaner. And then we have actual code to discuss.

dbrgn: Pushed 🙂

niklasmohrin: Okay, I can see your point. I feel like both interpretations make sense, so I guess we should ask upstream for clarification? (I am not really confortable to commit to your suggestion with the risk of having a breaking change later, but my suggestion could likely be too complicated.) Should I open an issue?

dbrgn: Feel free to open a request for clarification, but if there isn't any activity within - let's say - a week, I'd like to go ahead and merge this. I think that treating `all` like `current` - except for the places where considering all platforms makes sense (which is only `--list`, I think) - would be a very logical and easy to implement approach. If someone would specify `-p all` for another command, then it would simply work as it would without any `-p` argument (so it's essentially being ignored). However, this approach should probably be documented, so opening an issue would be the right thing to do.

If the spec would require that `-p all` may only be accepted in combination with `--list`, it would simply make argument validation harder for all clients.

niklasmohrin: See tldr-pages/tldr#7528 :)

dbrgn:
Discussion in https://github.com/tldr-pages/tldr/issues/7528 and https://github.com/tldr-pages/tldr/pull/7561 is still ongoing, but I'll remove this from the milestone: This definitely isn't one of the important compatibility issues we're having.

dbrgn:
Closing for now.

PR #230, opened 2021-12-05
Introduce command_or_file arg group

### LINKED ISSUE #219: Cannot combine `-f / --render` with other options
This command fails, even though it should work:

    cargo run -- --raw -f tar.md

This is a regression introduced by #108, because of the `requires`-constraints in commands like `raw`.

First, write a regression test to catch this bug.

Then find out how to tell clap that "either a command or a -f file is required".

dbrgn:
This allows to say "this command line parameter requires either a command or a raw file".

Fixes #219.

(Clap is pretty awesome!)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,7 +20,7 @@ use std::{env, path::PathBuf, process};
 
 use app_dirs::AppInfo;
 use atty::Stream;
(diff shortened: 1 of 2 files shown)


niklasmohrin (approved):
Seems correct :)

PR #231, opened 2021-12-05
Move pager configuration near render logic

### LINKED ISSUE #169: `tldr --update` output is displayed by pager

tranzystorek-io:
Fixes #169

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -11,10 +11,14 @@ use crate::{
     line_iterator::LineIterator,
 };
 
-// Set up display pager
+/// Set up display pager
+///
+/// SAFETY: this function may be called multiple times
 #[cfg(not(target_os = "windows"))]
 fn configure_pager(_: bool) {
-    pager::Pager::with_default_pager("less -R").setup();
+    use std::sync::Once;
+    static INIT: Once = Once::new();
+    INIT.call_once(|| pager::Pager::with_default_pager("less -R").setup());
 }
 
 #[cfg(target_os = "windows")]
--- a/src/output.rs
+++ b/src/output.rs
@@ -11,14 +11,33 @@ use crate::{
     line_iterator::LineIterator,
 };
 
+// Set up display pager
+#[cfg(not(target_os = "windows"))]
+fn configure_pager(_: bool) {
+    pager::Pager::with_default_pager("less -R").setup();
+}
+
+#[cfg(target_os = "windows")]
+fn configure_pager(enable_styles: bool) {
+    use crate::utils::print_warning;
+    print_warning(enable_styles, "--pager flag not available on Windows!");
+}
+
 /// Print page by path
 pub fn print_page(
     lookup_result: &PageLookupResult,
     enable_markdown: bool,
+    enable_styles: bool,
+    use_pager: bool,
     config: &Config,
 ) -> Result<()> {
     // Create reader from file(s)
     let reader = lookup_result.reader()?;
+
+    // Configure pager if applicable
+    if use_pager || config.display.use_pager {
+        configure_pager(enable_styles);
+    }
 
     // Lock stdout only once, this improves performance considerably
     let stdout = io::stdout();
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,8 +21,6 @@ use std::{env, path::PathBuf, process};
 use app_dirs::AppInfo;
 use atty::Stream;
 use clap::{AppSettings, ArgGroup, Parser};
-#[cfg(not(target_os = "windows"))]
-use pager::Pager;
 
 mod cache;
 mod config;
@@ -153,17 +151,6 @@ struct Args {
     // while TLDR specification requires `-v` to be used.
     #[clap(short = 'v', long = "version")]
     version: bool,
-}
-
-/// Set up display pager
-#[cfg(not(target_os = "windows"))]
-fn configure_pager(_: bool) {
-    Pager::with_default_pager("less -R").setup();
-}
-
-#[cfg(target_os = "windows")]
-fn configure_pager(enable_styles: bool) {
-    print_warning(enable_styles, "--pager flag not available on Windows!");
 }
 
 /// The cache should get updated if this was requested by the user, or if auto
@@ -412,11 +399,6 @@ fn main() {
         }
     };
 
-    // Set up pager
-    if args.pager || config.display.use_pager {
-        configure_pager(enable_styles);
-    }
-
     // Show various paths
     if args.show_paths {
         show_paths(&config);
@@ -433,7 +415,7 @@ fn main() {
     // If a local file was passed in, render it and exit
     if let Some(file) = args.render {
         let path = PageLookupResult::with_page(file);
-        if let Err(ref e) = print_page(&path, args.raw, &config) {
+        if let Err(ref e) = print_page(&path, args.raw, enable_styles, args.pager, &config) {
             print_error(enable_styles, e);
             process::exit(1);
         } else {
@@ -496,7 +478,9 @@ fn main() {
             &languages,
             config.directories.custom_pages_dir.as_deref(),
         ) {
-            if let Err(ref e) = print_page(&lookup_result, args.raw, &config) {
+            if let Err(ref e) =
+                print_page(&lookup_result, args.raw, enable_styles, args.pager, &config)
+            {
                 print_error(enable_styles, e);
                 process::exit(1);
             }


niklasmohrin:
Could you write a test so that the discussed refactor later doesn't break this again? I think it would be really cool if you mocked the pager with some shell script that just writes a file to a tempdir when called. Then you would just have to set the script as the pager through the environment variable. If you don't want to, we can also open a new issue for someone else to write the test though :)

tranzystorek-io:
@niklasmohrin I'd be happy to, just a few issues I can see:

- I guess you mean writing an empty file to the `/tmp` directory to verify that a pager was run, if so, I'm guessing it would need to be cleaned_up after the test?
- With the new `clap-derive` parsing, I cannot do, `tldr --pager --update` without giving a command, I'd have to somehow modify the `display.use_pager` config in the toml file, and I think the `TestEnv::write_config` method doesn't help here

dbrgn:
> With the new `clap-derive` parsing, I cannot do, `tldr --pager --update` without giving a command

Hm, that actually sounds like a useful combination though. If you want, you could change that behavior in a separate PR:

- Create a new `ArgGroup` similar to [`command_or_file`](https://github.com/dbrgn/tealdeer/blob/1a3624d011684e1f381820f5abb3c4b69dd65c00/src/main.rs#L64) called `action`. It would depend on `&["command", "render", "pager"]`. (Maybe it can also depend on `&["command_or_file", "pager"]` instead, I'm not sure if that works.)
- The `pager` could then be changed from `requires = "command_or_file"` to `requires = "action"`.

> I guess you mean writing an empty file to the `/tmp` directory to verify that a pager was run, if so, I'm guessing it would need to be cleaned_up after the test?

I'm not sure what @niklasmohrin had in mind, but cleanup sounds like a good thing. The thing is that it'll be a bit tricky. You cannot write the file to a static path, because then multiple tests would overwrite the file. And if you use a random path, you need some way to communicate that path between the test and the mock pager.

Ideally you would somehow be able to pass a filepath from the test to the mock pager. Then the test could reuse the `TempDir` struct that's already being used in `TestEnv`, which features auto-cleanup after the test is done.

tranzystorek-io:
> Hm, that actually sounds like a useful combination though.

I'm not sure what you have in mind, when we just want to run an update, a pager seems completely unnecessary anyway

dbrgn:
@tranzystorek-io uh, sorry, you're right. I was thinking of `tldr --pager --list` 🙈

dbrgn:
> and I think the `TestEnv::write_config` method doesn't help here

Why is that? I think it should work.

tranzystorek-io:
Nevermind, I misunderstood how `write_config` works, thought it takes a preexisting config.toml file 😅

tranzystorek-io:
> The tests would need to have a dev-dependency on that sub-crate though, so that it's always built first.

AFAIK a binary crate cannot be a dependency, unfortunately

dbrgn:
@tranzystorek-io hm, you're right. you can have both a bin.rs and a lib.rs in your crate, but if you add a dependency, then only the lib part will be built.

niklasmohrin:
My initial idea was to have a short script that creates a file (it would be even easier with `PAGER="touch testenvcache/tealdeer_log.txt`) which would be a very easy check whether or not the pager was run. I think it would be fine to not have a super portable test for this because we only want to check when a pager is spawned, which should behave the same on all platforms. The problem is that the `pager` crate by default checks whether stdout is a tty and doesn't spawn the pager if not - as it is in when running subprocesses in tests. I found a way to have it use the stdout from the tests - that means it closes the file descriptor on exit though :D We could probably make some `dup(2)` magic if we want to follow this path, but it is a bit hacky already. For reference, this is my final result:

```rust
#[test]
fn test_pager_is_called() {
    let testenv = TestEnv::new();
    testenv.add_entry("thecommand", "thecontent");
    let logfile = testenv.cache_dir.path().join("tealdeer_log.txt");

    assert!(!logfile.exists());
    testenv
        .command()
        .env("PAGER", format!("touch {}", logfile.display()))
        .arg("--pager") // <- Comment and uncomment this line
        .arg("thecommand")
        .stdout(unsafe { std::process::Stdio::from_raw_fd(std::io::stdout().as_raw_fd()) })
        .assert()
        .success();
    assert!(logfile.exists());
}
```

niklasmohrin:
@tranzystorek-io @dbrgn I guess testing is too difficult for now, let's leave this for later  <sub>/ never</sub>

dbrgn:
Since both calls are directly followed by `print_page`, and since we want to use the pager only for printing the page, wouldn't it be better to set up the pager in `print_page`? (It has access to the config already.)

niklasmohrin:
Sounds like a good idea, we should probably make sure to a) check whether we can call `configure_pager` multiple times or wrap it in a `Once` or so? (I wouldn't expect `print_page` to have the contract we would otherwise attach to it). What do you think?

tranzystorek-io:
One thing I'm worrying is that `configure_pager` needs the `enable_styles` flag, so it would need to be passed to `print_page`, which seems like a separate, unrelated concern.

dbrgn:
The `enable_styles` flag is part of `Config`, which is already passed to `print_page`.

niklasmohrin:
I agree that (even without the proposed addition) `print_page` does a lot of things, I suppose the `enable_markdown` branching is more similar to the pager thing than it is to actually printing to the terminal. Maybe we should move the two printing blocks into separate "functionality" functions and leave all of the rest (pager, markdown) in one "business logic" function. What do you think?

dbrgn:
I don't really see an issue with `print_page`, the function deals with outputting the page in the proper mode to the terminal, with or without markdown, with or without pager. Locking stdout is also "preparation logic", but it clearly has to do with "page printing".

However, if you feel this could be improved, feel free to propose an improvement in a PR 🙂

dbrgn:
In any case, the goal here is to fix #169, and moving the pager setup logic into `print_page` would solve this. Further improvements can be done later.

tranzystorek-io:
> The `enable_styles` flag is part of `Config`, which is already passed to `print_page`.

I can't find it in `Config`, in main it seems to come only from the command line args (the `--color` flag)

dbrgn:
Uhh, sorry @tranzystorek-io, you're right of course (I was mixing up `config.display.use_pager` and `enable_styles`).

Maybe we should handle styles for printing error messages differently instead of passing around the flag everywhere (maybe a global static or thread-local variable?). However, for now I'd say passing it to `print_page` is OK (better than repeating the `configure_pager` call twice).

niklasmohrin:
> Maybe we should handle styles for printing error messages differently instead of passing around the flag everywhere (maybe a global static or thread-local variable?).

Didn't we discuss an `OutputManager` or something similar at some point?

tranzystorek-io:
For now, I moved the `configure_pager` logic to the output module, as well as rearranged `print_page` arguments as discussed

src/output.rs
```diff
@@ -10,15 +10,34 @@ use crate::{
     line_iterator::LineIterator,
 };
 
+// Set up display pager
+#[cfg(not(target_os = "windows"))]
+fn configure_pager(_: bool) {
+    pager::Pager::with_default_pager("less -R").setup();
```
dbrgn, line 16: Could you wrap this in [Once](https://doc.rust-lang.org/std/sync/struct.Once.html) and add a note to the function docs, indicating that the function may be called multiple times safely? (Also, function docs should start with triple-slash.)

dbrgn (changes requested):
One last change request.

src/output.rs
tranzystorek-io: Done! 😄

src/output.rs
```diff
@@ -11,15 +14,38 @@ use crate::{
     line_iterator::LineIterator,
 };
 
+static INIT: Once = Once::new();
```
niklasmohrin, line 17: Sorry for all the back and forth: You can move this static into the function body of `configure_pager`, but only the linux version so that the linter is happy on windows too :)

Other than that, lets get this merged

tranzystorek-io: Done, nice catch!

niklasmohrin (approved):
Very nice!

I also ran a quick benchmark, `Once` doesn't seem to make any difference :+1:

dbrgn (approved):
Thanks!

PR #238, opened 2021-12-29
Concatenate page and patch bytes before processing

### LINKED ISSUE #181: Custom patches should not need to be formatted like a full page

dbrgn:
This way a patch file does not need a well-formed header anymore.

Fixes #181 and thus the last open feature/bug issue for the 1.5.0 milestone! 🎉

Replaces #202.

### DIFF
--- a/docs/src/usage_custom_pages.md
+++ b/docs/src/usage_custom_pages.md
@@ -4,7 +4,7 @@ Tealdeer allows creating new custom pages, overriding existing pages as well as
 extending existing pages.
 
 The directory, where these custom pages and patches can be placed, follows OS
-conventions. On Linux, for example, the default location is
+conventions. On Linux for instance, the default location is
 `~/.local/share/tealdeer/pages/`. To print the path used on your system, simply
 run `tldr --show-paths`.
 
@@ -32,6 +32,9 @@ Sometimes you don't want to fully replace an existing upstream page, but just
 want to extend it with your own examples that you frequently need. In this
 case, use a file called `<command>.patch`, it will be appended to existing
 pages.
+
+> **Note:** Because the patch file will be concatenated directly to the page
+> file, it needs to start with an empty line.
 
 Path:
 
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -387,4 +387,55 @@ impl Cache {
 /// Unit Tests for cache module
 #[cfg(test)]
 mod tests {
+    use super::*;
+
+    use std::{
+        fs::File,
+        io::{Read, Write},
+    };
+
+    #[test]
+    fn test_reader_with_patch() {
+        // Write test files
+        let dir = tempfile::tempdir().unwrap();
+        let page_path = dir.path().join("test.page");
+        let patch_path = dir.path().join("test.patch");
+        {
+            let mut f1 = File::create(&page_path).unwrap();
+            f1.write_all(b"Hello\n").unwrap();
+            let mut f2 = File::create(&patch_path).unwrap();
+            f2.write_all(b"World").unwrap();
+        }
+
+        // Create chained reader from lookup result
+        let lr = PageLookupResult::with_page(page_path).with_optional_patch(Some(patch_path));
+        let mut reader = lr.reader().unwrap();
+
+        // Read into a Vec
+        let mut buf = Vec::new();
+        reader.read_to_end(&mut buf).unwrap();
+
+        assert_eq!(&buf, b"Hello\nWorld");
+    }
+
+    #[test]
+    fn test_reader_without_patch() {
+        // Write test file
+        let dir = tempfile::tempdir().unwrap();
+        let page_path = dir.path().join("test.page");
+        {
+            let mut f = File::create(&page_path).unwrap();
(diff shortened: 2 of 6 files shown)


dmaahs2017 (approved):

docs/src/usage_custom_pages.md
```diff
@@ -33,6 +33,9 @@ want to extend it with your own examples that you frequently need. In this
 case, use a file called `<command>.patch`, it will be appended to existing
 pages.
 
+> **Note:** Because the patch file will be concatenated directly to the page
+> file, it needs to start with an empty line.
+
```
niklasmohrin, line 38: Any way we can work around that? An intermediate reader that just outputs one empty line? There is an [`impl Read for &[u8]`](https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-2)

src/cache.rs
```diff
@@ -49,8 +48,36 @@ impl PageLookupResult {
...
+        //
+        // Note: It might be worthwhile to create our own struct that accepts
+        // the page and patch files and that will read them sequentially,
+        // because it avoids the boxing below. However, the performance impact
+        // would first need to be shown to be significant using a benchmark.
+        Ok(BufReader::new(if let Some(patch_file) = patch_file_opt {
+            Box::new(page_file.chain(patch_file)) as Box<dyn Read>
+        } else {
+            Box::new(page_file) as Box<dyn Read>
+        }))
```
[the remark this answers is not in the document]
dbrgn: I prefer the current approach, because opening the file and creating a reader are clearly separated.

docs/src/usage_custom_pages.md
dbrgn: If we want to fix this, we should implement a custom reader, similar to what @dmaahs2017 did. Because then we can look at the ending of the page and the beginning of the patch, and only output the empty line if necessary.

However, I'd start with the current approach for now, it's easy to understand (pure concatenation) and doesn't do any magic. We can always add more heuristics, but we can't take them away without making it a breaking change.

src/cache.rs
```diff
@@ -49,8 +48,36 @@ impl PageLookupResult {
...
-        iter::once(self.page_path.as_path()).chain(self.patch_path.as_deref())
+    /// Create a buffered reader that sequentially reads from the page and the
+    /// patch, as if they were concatenated.
+    ///
+    /// This will return an error if either the page file or the patch file
+    /// cannot be opened.
+    pub fn reader(&self) -> Result<BufReader<Box<dyn Read>>, String> {
+        // Open page file
+        let page_file = File::open(&self.page_path)
+            .map_err(|msg| format!("Could not open page file at {:?}: {}", self.page_path, msg))?;
```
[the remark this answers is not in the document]
dbrgn: Yes, I've definitely thought about that, and it probably makes sense to start using it. However, for now using Strings is perfectly OK in an application (as opposed to a library).

niklasmohrin:
We could avoid the dynamic dispatch if we use https://docs.rs/either/latest/either/enum.Either.html#impl-Read, don't know if it is overkill though

docs/src/usage_custom_pages.md
niklasmohrin: I wouldn't say tt's really a breaking change, the patches feature hasn't been released yet - we could add a note in the release, but I don't feel like more effort is needed. As a user, I would even prefer if I could remove all the empty lines in my personal patches over having the compatibility.

Then, it's not magic either, in the code, it is just one more `chain`:
```rust
Box::new(page_file.chain(b"\n".as_slice()).chain(patch_file))
```

dbrgn:
Whoa, I didn't know that `Either` has a blanket-`Read`-impl. This is nice!

dbrgn:
Ahh, wait, `Either` is a separate crate and not in the stdlib. Then I don't think it's worth the added dependency 🙂

docs/src/usage_custom_pages.md
dbrgn: Ok, you're right, and I guess it's very unlikely that an official page already ends with an empty line. I'll add that change!

dbrgn:
Great, I'll merge this once CI passes. Thanks @dmaahs2017 and @niklasmohrin for the review and comments! I think the patch system will be a very useful feature in the next release.

niklasmohrin:
Fair enough - we can always come back to this thought if we feel like optimizing

dbrgn:
Hehe, `as_slice()` was added only in Rust 1.57, which hasn't been stable for a month (as per our MSRV policy). I'll use `&b"\n"[..]` unless there's a better way.

Issue #244, opened 2022-01-01
Homebrew binary for Apple Silicon M1 panicks with --update


storopoli:
This is the new 1.5.0 version:

```bash
❯ tldr  --version
tealdeer 1.5.0

❯ brew info tealdeer
tealdeer: stable 1.5.0 (bottled)
Very fast implementation of tldr in Rust
https://github.com/dbrgn/tealdeer
Conflicts with:
  tldr (because both install `tldr` binaries)
/opt/homebrew/Cellar/tealdeer/1.5.0 (12 files, 4.3MB) *
  Poured from bottle on 2022-01-01 at 11:05:57
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/tealdeer.rb
License: Apache-2.0
==> Dependencies
Build: rust ✘
==> Caveats
zsh completions have been installed to:
  /opt/homebrew/share/zsh/site-functions
```

Here's the full backtrace:

```bash
❯ RUST_BACKTRACE=full tldr -u
thread 'main' panicked at 'Client::new(): reqwest::Error { kind: Builder, source: MissingOrMalformedExtensions }', /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/github.com-1ecc6299db9ec823/reqwest-0.11.8/src/blocking/client.rs:798:38
stack backtrace:
   0:        0x102dfd988 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::ha47ac0a9192e239e
   1:        0x102c804f4 - core::fmt::write::hfbe6a7b828c32513
   2:        0x102def8ec - std::io::Write::write_fmt::h34cd061a987d18fd
   3:        0x102defe80 - std::panicking::default_hook::{{closure}}::h2b5c2a08263c3d54
   4:        0x102df09f8 - std::panicking::rust_panic_with_hook::h36a6e8d335334d30
   5:        0x102dfe020 - std::panicking::begin_panic_handler::{{closure}}::h71decad72c0b1843
   6:        0x102dfdf8c - std::sys_common::backtrace::__rust_end_short_backtrace::hfd636421a82c4798
   7:        0x102df02c4 - _rust_begin_unwind
   8:        0x102e886f4 - core::panicking::panic_fmt::hebdff93de232aa26
   9:        0x102e88c68 - core::result::unwrap_failed::hf00c663ffab666bd
[...]

PR #247, opened 2022-01-02
Cache: Return error if HTTP client cannot be created

dbrgn:
Apparently `Client::new` can panic (see #244), so let's return the error instead.


dbrgn:
Hm, I'm currently converting the codebase to use anyhow, which also affects these error messages, so I'll close this PR.

PR #253, opened 2022-01-09
Handle cache download errors

### LINKED ISSUE #252: Fails to update cache: thread 'main' panicked (Arch Linux)
When I run `RUST_BACKTRACE=full tldr --update`, this is the error I get:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidArchive("Could not find central directory end")', src/cache.rs:166:33
stack backtrace:
0:     0x560e37b784fd - <unknown>
1:     0x560e379d62ec - <unknown>
2:     0x560e37b76d84 - <unknown>
3:     0x560e37b774d0 - <unknown>
4:     0x560e37b7685a - <unknown>
5:     0x560e37b978f8 - <unknown>
6:     0x560e37b97876 - <unknown>
7:     0x560e37b97832 - <unknown>
8:     0x560e37962b60 - <unknown>
9:     0x560e37962e42 - <unknown>
10:     0x560e37bae307 - <unknown>
11:     0x560e37bc3dd5 - <unknown>
12:     0x560e37bb63a3 - <unknown>
13:     0x560e37bbfb79 - <unknown>
14:     0x7f4d4a849b25 - __libc_start_main
15:     0x560e3796d54e - <unknown>
16:                0x0 - <unknown>
```

About my system:
OS: Arch Linux
tealdeer version: 1.5.0

dbrgn:
With this fix, when an invalid ZIP is downloaded:

```
  Error: Could not update cache

  Caused by:
      0: Could not decompress downloaded ZIP archive
      1: invalid Zip archive
```

When a non-successful HTTP error status is returned:

```
  Error: Could not update cache

  Caused by:
      0: Could not download tldr pages from https://tldr.sh/assets/tldr.zap
      1: HTTP status client error (404 Not Found) for url (https://tldr.sh/assets/tldr.zap)
```

Anyhow is awesome 🙂

Fixes #252.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -150,7 +150,11 @@ impl Cache {
         let client = builder
             .build()
             .context("Could not instantiate HTTP client")?;
-        let mut resp = client.get(&self.url).send()?;
+        let mut resp = client
+            .get(&self.url)
+            .send()?
+            .error_for_status()
+            .with_context(|| format!("Could not download tldr pages from {}", &self.url))?;
         let mut buf: Vec<u8> = vec![];
         let bytes_downloaded = resp.copy_to(&mut buf)?;
         debug!("{} bytes downloaded", bytes_downloaded);
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -2,7 +2,7 @@ use std::{
     env,
     ffi::OsStr,
     fs::{self, File},
-    io::{BufReader, Cursor, Read, Seek},
+    io::{BufReader, Cursor, Read},
     path::{Path, PathBuf},
     time::{Duration, SystemTime},
 };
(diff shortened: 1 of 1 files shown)


src/cache.rs
```diff
@@ -150,16 +150,20 @@ impl Cache {
...
         debug!("{} bytes downloaded", bytes_downloaded);
         Ok(buf)
     }
 
     /// Decompress and open the archive
-    fn decompress<R: Read + Seek>(reader: R) -> ZipArchive<R> {
-        ZipArchive::new(reader).unwrap()
+    fn decompress<R: Read + Seek>(reader: R) -> Result<ZipArchive<R>> {
+        Ok(ZipArchive::new(reader)?)
     }
```
niklasmohrin, line 168: Is there any benefit from having this as a method now? it would probably be easier to just inline it

src/cache.rs
```diff
@@ -150,16 +150,20 @@ impl Cache {
         let client = builder
             .build()
             .context("Could not instantiate HTTP client")?;
-        let mut resp = client.get(&self.url).send()?;
+        let mut resp = client
+            .get(&self.url)
+            .send()?
```
niklasmohrin, line 155: We don't provide any `context` here, but the errors are pretty clear so we can leave it like this :)

Nice find with `error_for_status`!

src/cache.rs
dbrgn: Yeah, that's true.

niklasmohrin (approved):

PR #254, opened 2022-01-09
Add note about auto-updates when cache is missing

### LINKED ISSUE #250: Auto update cache by default?
Hi!

Thanks for writing tealdeer, it's been great to use.

I've found having to manually update the cache to be confusing, especially on first-use. Consider the reference tldr package versus tealdeer:

```sh
❯ paru -S tldr
❯ tldr tldr

  Displays simple help pages for command-line tools, from the tldr-pages project.
  More information: <https://tldr.sh>.
```

```sh
❯ paru -S tealdeer
❯ tldr tldr
Warning: Cache not found. Please run `tldr --update`.
```

I had to dig throug the documentation and GitHub issues to find https://github.com/dbrgn/tealdeer/issues/90 and https://dbrgn.github.io/tealdeer/config_updates.html. At which point I had to seed the config:

```sh
❯ tldr --seed-config
```

Then enable auto cache update:

```diff
[updates]
-auto_update = false
+auto_update = true
auto_update_interval_hours = 720
```

Does it make sense to enable this behaviour by default?

dbrgn:
Output:

screenshot-20220109-153929

Fixes #250.

@niklasmohrin @uhthomas what do you think?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,10 +81,20 @@ fn check_cache(args: &Args, enable_styles: bool) -> CheckCacheResult {
             CheckCacheResult::CacheFound
         }
         CacheFreshness::Missing => {
-            print_warning(
+            print_error(
                 enable_styles,
-                "Cache not found. Please run `tldr --update`.",
+                &anyhow::anyhow!(
+                    "Page cache not found. Please run `tldr --update` to download the cache."
+                ),
             );
+            println!("\nNote: You can optionally enable automatic cache updates by adding the");
+            println!("following config to your config file:\n");
+            println!("  [updates]");
+            println!("  auto_update = true\n");
+            println!("The path to your config file can be looked up with `tldr --show-paths`.");
+            println!("To create an initial config file, use `tldr --seed-config`.\n");
+            println!("You can find more tips and tricks in our docs:\n");
+            println!("  https://dbrgn.github.io/tealdeer/config_updates.html");
             CheckCacheResult::CacheMissing
         }
     }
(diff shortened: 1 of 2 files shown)


niklasmohrin:
To be honest, I think it is a bit verbose and might be more intimidating than helpful for first-time users. How do you fell about:
```plain
Error: Cache not found. Please run `tldr --update` to download the cache.

Your first time using tealdeer? You can find some tips and tricks to get started at
    https://dbrgn.github.io/tealdeer/getting_started.html
```

... and then we put the text you have here in a subsection of `configuration` (or it already exists, I don't know). In the getting-started page, we could put some links and maybe some config recommendations for ease of use (like `auto-update`). What do you think?

(I also changed the "Warning" to "Error", because this is an error; I think we should change this on main :D)

dbrgn:
I thought about it again, and I'd stick to the current version of the PR (with slight rewording). The message will only be shown when a cache is missing, meaning that tealdeer wasn't yet initialized on that system (so usually just once). In this case, it's good if the information is right there, without having to open the webbrowser and visit a website first. (I think I wouldn't do that myself, I'd just copy-paste the command and miss the bit about auto updates.)

Here's an updated version:

screenshot-20220205-012018

@niklasmohrin can you live with that? 🙂

dbrgn:
Ah, and I changed the warning to an error.

niklasmohrin:
Code looks good; I personally still think that the note is a bit long and the error is a bit lost in all of the text, but that may just be me (one other person I asked said it's clear to them)

PR #256, opened 2022-01-15
Generate shell completion scripts

### LINKED ISSUE #220: Auto-generate completion files through clap
Replace our manually-written completion files for bash, zsh and fish with auto-generated completion files.

Clap supports this through https://github.com/clap-rs/clap/tree/master/clap_generate.

These completion files don't need to be checked in, instead the release build script should generate them. Additionally, the README could contain a section on how to generate these completions (might be interesting for packagers).

SimplyDanny:
Shell completion scripts are generated in every release build.

The build script requires access to the definition of the command line arguments and options `Args`. This struct is moved into its own file to be able to import it separately together with its `types` dependency.

`clap_complete` names the completion scripts `tldr.bash`, `tldr.fish` and `_tldr`, respectively. As it seems, these defaults cannot be changed. If the previous names are preferred, the files must be renamed in the build script after their generation.

The scripts are put into the `completion` directory where they are ignored from Git.

This PR closes #220.


SimplyDanny:
Hm ... I just noticed that the important dynamic completion of page names is unfortunately not yet supported by clap according to https://github.com/clap-rs/clap/issues/1232.

niklasmohrin:
That is quite unfortunate and, for me personally, a deal breaker. I still like the move of `Args` and of the completion files into their own directory (even when it is included in the three :D) and would keep those changes.

@dbrgn How do you feel about the generated completions?

dbrgn:
Hmm, that's unfortunate indeed, I did't realize that this is not possible. I agree that this is a deal breaker. For now, I'd close this, and if someone has the motivation, that person could open a feature request in the clap repo.

I also agree that moving the completions to a subdirectory makes sense.

Regarding the moving of CLI definitions into a separate module: At first I didn't like it, but after some more thinking about it, I think it's a good change.

@SimplyDanny do you want to open a new PR with the first commit, plus a commit that moves the completion files into a  `completions/` directory?

SimplyDanny:
Sure. I will prepare separate PRs for the accepted changes.

Keep in mind, though, that changes to the location of the completion scripts require updates of the packages (e.g. the Homebrew formula). I only generated them into an own folder, because their names would have changed anyway.

dbrgn:
> Keep in mind, though, that changes to the location of the completion scripts require updates of the packages (e.g. the Homebrew formula). I only generated them into an own folder, because their names would have changed anyway.

Yes, I'm aware, but I think the change is worth it, because in the long run there may be additional completions as well, and if we keep them all in the root directory then it may get a bit messy.

(The main reason why I didn't rename/move them before was that I wanted to investigate clap-generated completions first, in which case we could have removed the completions from the repository entirely and only generate them for the releases 🙂)

dbrgn:
Superseded by #258 and #259. Thanks @SimplyDanny!

PR #257, opened 2022-01-16
Add `no-auto-update` flag

cyqsimon:
- Adding an empty `rustfmt.toml` enforces default formatter settings regardless of user's global settings
  - See https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#configuring-rustfmt
- Currently if the user enables auto update in the config file, there is no way to temporarily disable it via CLI
  - This is a problem when for example, the network is not available or is very slow. The client simply errors and refuses to display anything, despite the fact that the cache may be perfectly good (albeit a little old)
  - Adding `no-auto-update` flag allows the client to stay usable in such situations

I will need some help with the completion scripts since I have no experience with these.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,9 +49,8 @@ const APP_INFO: AppInfo = AppInfo {
 };
 const ARCHIVE_URL: &str = "https://tldr.sh/assets/tldr.zip";
 
-/// The cache should get updated if this was requested by the user, or if auto
-/// updates are enabled and the cache age is longer than the auto update interval,
-/// unless it is explicitly disabled by the user.
+/// The cache should be updated if it was explicitly requested,
+/// or if an automatic update is due and allowed.
 fn should_update_cache(args: &Args, config: &Config) -> bool {
     args.update
         || (!args.no_auto_update
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,10 +50,12 @@ const APP_INFO: AppInfo = AppInfo {
 const ARCHIVE_URL: &str = "https://tldr.sh/assets/tldr.zip";
 
 /// The cache should get updated if this was requested by the user, or if auto
-/// updates are enabled and the cache age is longer than the auto update interval.
+/// updates are enabled and the cache age is longer than the auto update interval,
+/// unless it is explicitly disabled by the user.
 fn should_update_cache(args: &Args, config: &Config) -> bool {
     args.update
-        || (config.updates.auto_update
+        || (!args.no_auto_update
+            && config.updates.auto_update
             && Cache::last_update().map_or(true, |ago| ago >= config.updates.auto_update_interval))
 }
 
--- a/completion/bash_tealdeer
+++ b/completion/bash_tealdeer
@@ -6,7 +6,7 @@ _tealdeer()
 	_init_completion || return
 
 	case $prev in
-		-h|--help|-v|--version|-l|--list|-u|--update|-c|--clear-cache|-p|--pager|-r|--raw|--show-paths|--seed-config|-q|--quiet)
+		-h|--help|-v|--version|-l|--list|-u|--update|--no-auto-update|-c|--clear-cache|-p|--pager|-r|--raw|--show-paths|--seed-config|-q|--quiet)
 			return
 			;;
 		-f|--render)
--- a/completion/fish_tealdeer
+++ b/completion/fish_tealdeer
@@ -3,19 +3,20 @@
 # https://github.com/dbrgn/tealdeer/
 #
 
-complete -c tldr -s h -l help        -d 'Print the help message.' -f
-complete -c tldr -s v -l version     -d 'Show version information.' -f
-complete -c tldr -s l -l list        -d 'List all commands in the cache.' -f
-complete -c tldr -s f -l render      -d 'Render a specific markdown file.' -r
-complete -c tldr -s p -l platform    -d 'Override the operating system.' -xa 'linux macos sunos windows'
-complete -c tldr -s u -l update      -d 'Update the local cache.' -f
-complete -c tldr -s c -l clear-cache -d 'Clear the local cache.' -f
-complete -c tldr -s p -l pager       -d 'Use a pager to page output.' -f
-complete -c tldr -s r -l raw         -d 'Display the raw markdown instead of rendering it.' -f
-complete -c tldr -s q -l quiet       -d 'Suppress informational messages.' -f
-complete -c tldr      -l show-paths  -d 'Show file and directory paths used by tealdeer.' -f
-complete -c tldr      -l seed-config -d 'Create a basic config.' -f
-complete -c tldr      -l color       -d 'Controls when to use color.' -xa 'always auto never'
+complete -c tldr -s h -l help           -d 'Print the help message.' -f
+complete -c tldr -s v -l version        -d 'Show version information.' -f
+complete -c tldr -s l -l list           -d 'List all commands in the cache.' -f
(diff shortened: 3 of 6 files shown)


cyqsimon:
Also I think the `Args` struct for clap could use some cleanup. More guard clauses in `main()` can be accomplished with `requries` and `conflicts_with` using clap, for example. But that belongs in a separate PR.

niklasmohrin:
Rustfmt change makes sense to me, good catch!

About the auto update: Auto update is a convenience feature and having to specify this flag is inconvenient. What if we only warn if auto updates fail because one is offline? @dbrgn

cyqsimon:
> Auto update is a convenience feature and having to specify this flag is inconvenient. What if we only warn if auto updates fail because one is offline?

I'm not particularly convinced that automatic detection of network availability/condition is a full solution, mostly for the reason that oftentimes network failure/non-availability manifests in the form of a timeout after an extensive wait. Imagine being on a train in a remote region with questionable cellular coverage and using your mobile hotspot for internet - it's pretty likely that the update doesn't fail immediately but suffers from terrible speed and frequent retries. Truth be told, this is in fact my exact situation a few days ago.

That being said I do agree that if auto update fails, it's better to make it a warning rather than a hard error. So this change would be a welcomed one in my books - I just don't think it solves the problem fully. I would insist that there needs to be a manual "off switch" for situations where the network is not completely dead but pretty unreliable. At least having the option of typing an additional flag and getting a response reliably is better than having to wait out a network timeout, or worse, not being to use the client at all.

niklasmohrin:
Okay, I can see your point. If we implement this, I would make the setting more generic, like "offline" or so. I did a quick search whether there is a convention for this kind of setting, but I couldn't find one. If there would be some (linux) convention that, let's say the `OFFLINE` env variable, is set for this scenario, I would be totally in favor of respecting that (but not adding tealdeer specific settings). With an extra flag, say `--offline`, I feel like this is a big change for a rather niche usecase. I am not strictly against it, it's just a lot of work that is expected from the program and it involves this system of settings that turn things on and off in different places; after all, tealdeer is offline by default

dbrgn:
The empty `rustfmt.toml` file is a good idea indeed!

To keep this discussion focussed on the no-auto-update flag, I'll cherry-pick that commit onto the main branch.

dbrgn:
I don't think there's a generic "offline"-variable, or at least I've never heard of such a thing. It would also open up a lot of questions, e.g. what happens if you run `OFFLINE=1 tldr --update`?

I think having a `--no-auto-update` file is fine, it's a simple change and I don't expect that it'll add more complexity in the future.

dbrgn (changes requested):
@cyqsimon can you rebase your branch against `main` and also update the usage docs with the changed ?

    cargo run -- --help > docs/src/usage.txt

cyqsimon:
I'm not very familiar with this workflow of cherry picking commits in the middle of a PR. Hopefully I haven't managed to screw everything up.

Also as a reminder the shell completion scripts haven't been updated yet.

cyqsimon:
I updated the completion scripts but I haven't checked them. As aforementioned, I do not have experience working with those so I might be making beginner mistakes. Please review these changes rather cautiously.

fish_tealdeer
```diff
@@ -3,19 +3,20 @@
...
 #
 
-complete -c tldr -s h -l help        -d 'Print the help message.' -f
-complete -c tldr -s v -l version     -d 'Show version information.' -f
-complete -c tldr -s l -l list        -d 'List all commands in the cache.' -f
-complete -c tldr -s f -l render      -d 'Render a specific markdown file.' -r
-complete -c tldr -s p -l platform    -d 'Override the operating system.' -xa 'linux macos sunos windows'
-complete -c tldr -s u -l update      -d 'Update the local cache.' -f
-complete -c tldr -s c -l clear-cache -d 'Clear the local cache.' -f
-complete -c tldr -s p -l pager       -d 'Use a pager to page output.' -f
```
cyqsimon, line 13: This `-s p` is a mistake right?

dbrgn: Yes, well spotted, the short version of that should be removed. A fix in a separate PR (once this is merged) would be welcome 🙂

dbrgn (approved):
LGTM now. @niklasmohrin do you agree?

> I'm not very familiar with this workflow of cherry picking commits in the middle of a PR. Hopefully I haven't managed to screw everything up.

Everything looks correct 🙂 If you have any future questions regarding git, don't hesitate to ask, we're happy to help.

src/main.rs
```diff
@@ -101,6 +101,10 @@ struct Args {
     #[clap(short = 'u', long = "update")]
     update: bool,
 
+    /// Disable auto-update, override config file
```
niklasmohrin, line 104: I feel like this could be improved: When reading this without knowing about what config options there are, this reads as if auto_update would be on by default. I would much rather have something like

> Disables auto-update, if it was set in the config file.

or

> Ignore the auto-update config setting.

or

> If auto-update is configured, disable it for this run.

My priority here are bottom to top, I like the last one best and the top one the least :)

---

Secondly, the actual key for the setting in config.toml is `auto_update`, with an underscore. This is the first time we mention specific config options in the usage text, so a decision is to be made here. I think it should be "auto_update", what do you think?

src/main.rs
```diff
@@ -154,10 +158,12 @@ struct Args {
 }
 
 /// The cache should get updated if this was requested by the user, or if auto
-/// updates are enabled and the cache age is longer than the auto update interval.
+/// updates are enabled and the cache age is longer than the auto update interval,
+/// unless it is explicitly disabled by the user.
```
niklasmohrin, line 162: Code is fine, I wonder if we want to trim this comment a bit though:

> The cache should be updated if it was explicitly requested, or if an automatic update is due and allowed.

@dbrgn We usually disagree about comments :sweat_smile: , what do you think about this one?

niklasmohrin (changes requested):
Code change looks good, I want to discuss some texts though :^)

src/main.rs
dbrgn: Your suggestion looks good.

src/main.rs
dbrgn: > If auto-update is configured, disable it for this run

That's also my preferred variant.

> I think it should be "auto_update", what do you think?

No strong opinion, but I think "auto-update" or "auto update" would be fine as well, referring to "the feature" and not "the config variable".

src/main.rs
cyqsimon: Excellent way to word this.

src/main.rs
cyqsimon: My attitude when it comes to comments is rather liberal (e.g. I will use expletives without hesitation if the situation calls for it), so I don't have a strong opinion on this either. But clarity is never a bad thing.

dbrgn:
Ahh, now the branch conflicts with the pull request #259 that was merged just recently 🙈

@cyqsimon if you want to update the pull request yourself:

```
# Only if you don't already have a remote...
git remote add upstream https://github.com/dbrgn/tealdeer/

# Update upstream source
git fetch upstream

# Ensure you're on the no-auto-update branch
git checkout no-auto-update

# Rebase and fix conflicts
git rebase upstream/main
```

If you want us to merge or rebase your commits for you, that's no problem either, just let us know!

cyqsimon:
Done

dbrgn (approved):

fish_tealdeer
dbrgn: I opened a PR here: https://github.com/dbrgn/tealdeer/pull/262

PR #259, opened 2022-01-17
Move shell completion scripts to their own directory

SimplyDanny:
The names of the scripts are unchanged. We could rename them to the scheme used by clap-generate to be prepared. 😉 What do you think?

Is there something else to update? Scripts? Documentation?


niklasmohrin:
There is a mention of the bash and fish completion files in `Cargo.toml`, I am not sure why though :thinking: @dbrgn ?

Other than that, I can't find any other places

There should definitely be a mention in `CHANGELOG.md`, but this will be added whenever we release the next time. I am not sure if we need to explicitly notify package maintainers about the change, if we mention it in the changelog - but that is a task for us, not you :)

dbrgn:
Could you also update the `upload-completions` job in `.github/workflows/release.yml`?

> There is a mention of the bash and fish completion files in `Cargo.toml`, I am not sure why though 🤔 @dbrgn ?

Hm, I think those could/should be removed as well.

docs/src/installing.md
```diff
@@ -57,6 +57,9 @@ To enable the log output, set the `RUST_LOG` env variable:
 
 ## Autocompletion
 
-- *Bash*: copy `bash_tealdeer` to `/usr/share/bash-completion/completions/tldr`
-- *Fish*: copy `fish_tealdeer` to `~/.config/fish/completions/tldr.fish`
-- *Zsh*: copy `zsh_tealdeer` to `/usr/share/zsh/site-functions/_tldr`
+Shell completion scripts are located in the folder `completion`. 
+Just copy them to their designated locations:
```
dbrgn, line 61:
```suggestion
Just copy them to their designated location:
```

dbrgn (changes requested):

docs/src/installing.md
SimplyDanny: Changed.

niklasmohrin (approved):

SimplyDanny:
> Could you also update the upload-completions job in .github/workflows/release.yml?

Done.

> Hm, I think those could/should be removed as well.

Done.

> The names of the scripts are unchanged. We could rename them to the scheme used by clap-generate to be prepared. 😉 What do you think?

I guess you missed this question. Shall we keep the names for now or use the ones produced by clap-generate already?

niklasmohrin:
In my opinion we should stick to the current names, _even if we had clap-generate_. The names are better suited for use in this repo and you have to copy them anyways, so renaming doesnt add any complexity then either

dbrgn (approved):

Issue #261, opened 2022-01-20
-bash: _init_completion: command not found


brokosz:
tealdeer 1.5.0 installed via homebrew on bash 5.1.16(1) gives the above error when  trying to autocomplete. I was able to work around this by changing lines 5 & 6 in the bash completion file from:

```
local cur prev words cword
 _init_completion || return
```
To:
```
local cur prev
_get_comp_words_by_ref cur prev
```

dbrgn:
What happens if you type `type _init_completion` in your shell?

Do you know what `_init_completion` and `_get_comp_words_by_ref` actually do (or in other words, can you explain what your change does)? Unfortunately I know bash completion files only very superficially.

brokosz:
You probably know more about bash completion than I do. 😁 Turns out I was on an old bash-completions package (and why `_init_completion` didn't work for me) so i'll close this issue. Fwiw - looks like the functions are pretty well documented in here: https://github.com/scop/bash-completion/blob/master/bash_completion#L339

PR #262, opened 2022-01-21
Clean up shell completions

dbrgn:


dbrgn:
Potentially include a fix for #261 in here as well...

dbrgn:
Merging since #261 was resolved.

Issue #260, opened 2022-01-18
Support Android Targets


leleliu008:
https://github.com/dbrgn/tealdeer/archive/v1.5.0.tar.gz

![21](https://user-images.githubusercontent.com/3040852/149950669-5295f210-72b9-4d54-9395-54e528593ede.PNG)

dbrgn:
Android is not currently supported. Just out of curiosity, what kind of Android device do you have that would benefit from a tldr installation?

leleliu008:
HUAWEI MatePad Pro
![matepad-pro](https://user-images.githubusercontent.com/3040852/149956737-1bdef0f8-f478-4c41-ab04-c4bcd987399b.jpg)

dbrgn:
Ok 🙂 I guess it shouldn't be too hard to add proper Android support.

leleliu008:
```patch
--- a/src/types.rs
+++ b/src/types.rs
@@ -43,7 +43,7 @@
 }
 
 impl PlatformType {
-    #[cfg(target_os = "linux")]
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     pub fn current() -> Self {
         Self::Linux
     }
@@ -66,6 +66,7 @@
 
     #[cfg(not(any(
         target_os = "linux",
+        target_os = "android",
         target_os = "macos",
         target_os = "freebsd",
         target_os = "netbsd",
```

dbrgn:
A pull request would be welcome. Note that "android" is a dedicated platform type supported by tldr (with some Android-specific pages available), so the `PlatformType` should be extended with an `Android` variant (and not just fall back to Linux).

niklasmohrin:
A PR for this should also add a job to github actions that makes sure the build still works for android; I am not sure whether we can test in a simulated android environment, a little research from future implementers would be nice :)

leleliu008:
We usually just check if can be built for Android and do not check if can be run on Android. Because ARM-based emulators were too slow. but In the real world, 99% Android devices are ARM-based.

dbrgn:
Android support should be working since #267 was merged. Could you test building the `main` branch?

niklasmohrin:
I wouldn't consider this issue done until we build in CI, it shouldn't be difficult (see [my comment in the other PR](https://github.com/dbrgn/tealdeer/pull/274#discussion_r897300717)). I don't think we need to run tests for android in CI, but if its cheap we can do that too.

[...]

PR #263, opened 2022-01-21
Github Actions: add android.yml to build for Android

leleliu008:
Signed-off-by: leleliu008 <leleliu008@gmail.com>

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,27 @@
+name: cross compile with android-ndk
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  cross-compile-for-android:
+    strategy:
+      fail-fast: false
+      matrix:
+        target-abi: [armeabi-v7a, arm64-v8a, x86, x86_64]
+
+    runs-on: macos-10.15
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - run:  brew install tree file
+
+      - run:  patch -p1 < android.patch
+
+      - run:  ./cargo-ndk.sh --help
+
+      - run:  ./cargo-ndk.sh ${{ matrix.target-abi }} 21 -vv
--- a/cargo-ndk.sh
+++ b/cargo-ndk.sh
@@ -1,0 +1,285 @@
+#!/bin/sh
+
+set -e
+
+COLOR_RED='\033[0;31m'          # Red
+COLOR_GREEN='\033[0;32m'        # Green
+COLOR_YELLOW='\033[0;33m'       # Yellow
+COLOR_BLUE='\033[0;94m'         # Blue
+COLOR_PURPLE='\033[0;35m'       # Purple
+COLOR_OFF='\033[0m'             # Reset
+
+success() {
+    printf '%b\n' "${COLOR_GREEN}[✔] $*${COLOR_OFF}"
+}
+
+die() {
+    printf '%b\n' "${COLOR_RED}💔  $*${COLOR_OFF}" >&2
+    exit 1
+}
+
+run() {
+    printf '%b\n' "${COLOR_PURPLE}==>${COLOR_OFF} ${COLOR_GREEN}$*${COLOR_OFF}"
+    eval "$*"
+    echo
+}
+
+getvalue() {
+    if [ $# -eq 0 ] ; then
+        cut -d= -f2
+    else
+        echo "$1" | cut -d= -f2
+    fi
+}
+
+check_if_pie_executable_for_the_given_abi() {
+    case $1 in
+        armeabi-v7a) file "$2" | grep -q 'ELF 32-bit LSB pie executable, ARM, EABI5'   ;;
+        arm64-v8a)   file "$2" | grep -q 'ELF 64-bit LSB pie executable, ARM aarch64,' ;;
+        x86)         file "$2" | grep -q 'ELF 32-bit LSB pie executable, Intel 80386,' ;;
+        x86_64)      file "$2" | grep -q 'ELF 64-bit LSB pie executable, x86-64,'      ;;
+        '')          die "check_if_pie_executable_for_the_given_abi <ABI> <FILEPATH> , <ABI> is not given." ;;
+        *)           die "check_if_pie_executable_for_the_given_abi <ABI> <FILEPATH> , unrecognized <ABI> : $1"
+    esac
+}
+
+build_and_install() {
+    CARGO_INSTALL_ARGS=$@
+
+    unset CARGO_PROJECT_DIR
+    unset CARGO_INSTALL_DIR
+
+    while [ -n "$1" ]
+    do
+        case $1 in
+            --path)
+                shift
+                CARGO_PROJECT_DIR="$1"
+                if [ -z "$CARGO_PROJECT_DIR" ] ; then
+                    die "--path <PATH> , <PATH> must not be empty."
+                fi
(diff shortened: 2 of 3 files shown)


.github/workflows/android.yml
```diff
@@ -0,0 +1,27 @@
...
+        target-abi: [armeabi-v7a, arm64-v8a, x86, x86_64]
+
+    runs-on: macos-10.15
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - run:  brew install tree file
+
+      - run:  patch -p1 < android.patch
```
niklasmohrin, line 23: Why not just apply the patch and persist that change?

.github/workflows/android.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches: [ main ]
+
+jobs:
+  cross-compile-for-android:
+    strategy:
+      fail-fast: false
+      matrix:
+        target-abi: [armeabi-v7a, arm64-v8a, x86, x86_64]
+
+    runs-on: macos-10.15
```
niklasmohrin, line 16: Why do we need to build on mac? It usually takes longer until one gets a mac machine for actions and they are slower than the linux counterparts

.github/workflows/android.yml
```diff
@@ -0,0 +1,27 @@
+name: cross compile with android-ndk
```
niklasmohrin, line 1: I think we can just put this job into the workflow file we already have

cargo-ndk.sh
```diff
@@ -0,0 +1,285 @@
+#!/bin/sh
```
niklasmohrin, line 1: Where does this file come from? It it related to [this repository with the same name](https://github.com/bbqsrc/cargo-ndk)? I would definitely prefer if we wouldn't have to vendor this script in our repository (and surely not in the root directory)

niklasmohrin (changes requested):
Thank you for contributing! I have some questions about the setup before we can merge this

.github/workflows/android.yml
leleliu008: Do you mean why not I file a pull request to apply the android.patch?

.github/workflows/android.yml
leleliu008: I can change it to `Ubuntu-20.04`

.github/workflows/android.yml
leleliu008: Ok, I will do it.

.github/workflows/android.yml
niklasmohrin: Yes, let's just apply the patch and commit these changes instead of having to maintain the patch file

cargo-ndk.sh
leleliu008: this file is written by me and it is nothing to do with [cargo-ndk project](https://github.com/bbqsrc/cargo-ndk)

I wanna use [cargo-ndk project](https://github.com/bbqsrc/cargo-ndk), but it has bugs, finnaly I decide write my own version in shell. it is very simple and readable. 

I developed a tool call [ndk-pkg](https://github.com/leleliu008/ndk-pkg), if you want keep the workflow simple, you can use this tool.

```
name: Ubuntu

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

jobs:
  test:

    runs-on: ubuntu-20.04

    steps:
      - uses: actions/checkout@v2

      - run: curl -LO https://raw.githubusercontent.com/leleliu008/ndk-pkg/master/bin/ndk-pkg

      - run: chmod +x ndk-pkg

      - run: ./ndk-pkg update

      - name: create formula
        run:  |
          cat > ~/.ndk-pkg/repos.d/offical/formula/tealdeer.sh << EOF
          package set summary "Very fast implementation of tldr in Rust"
          package set git.url "https://github.com/dbrgn/tealdeer.git"
          package set src.url "dir://${{ github.workspace }}"
          package set license "Apache-2.0"
          package set dep.pkg "openssl"
          package set bsystem "cargo"
          EOF

      - run: ./ndk-pkg install tealdeer --tree
```

.github/workflows/android.yml
leleliu008: I have said in https://github.com/dbrgn/tealdeer/issues/260

have you read the patch? do you accept this patch file. I'm fairly new to Rust language. I have no ablity to write perfect code in Rust now.

cargo-ndk.sh
leleliu008: I have port this project to `ndk-pkg`, here is a github action example https://github.com/leleliu008/ndk-pkg-formula-repository/runs/4992613697?check_suite_focus=true

.github/workflows/android.yml
niklasmohrin: The patch seems to go in the right direction, not a lot of Rust is required (and we can help you / tell you what to do to fix errors if they come up). Just run `patch -p1 < android.patch`, remove `android.patch` and push all of these changes here (if you need git help, feel free to reach out too) :)

.github/workflows/android.yml
niklasmohrin: In the other workflow we use `ubuntu-latest`, so that should do it :)

cargo-ndk.sh
niklasmohrin: Ah that looks interesting :+1: If you have the tealdeer formula in your package base, we don't need to run `cargo-ndk` here, right? I am happy with applying the `android.patch` and committing this here, so you don't need to patch the code in the formula. Additionally, we can run a standard `cargo build` for the android target in CI so that it doesn't break out of the blue (testing is harder). Does that sound good to you?

leleliu008: > If you have the tealdeer formula in your package base, we don't need to run cargo-ndk here, right?

In my ndk-pkg formula  repo, I use the stable tarball file https://github.com/dbrgn/tealdeer/archive/v1.5.0.tar.gz , not fetch from the github source tree HEAD

> we can run a standard `cargo build` for the android target in CI so that it doesn't break out of the blue (testing is harder).

just run `cargo build --target armv7-linux-androideabi` is not enough. some certain environment variables must be set, that's why I wrote `cargo-ndk.sh`.

niklasmohrin:
I must admit, I am not comfortable with `cargo-ndk`, this just adds too much code to maintain to this repository. In #274, I noted that there is the possibility of using `cross` which cross-compiles using docker containers. This works great, [`bat`](https://github.com/sharkdp/bat) uses it, I use it in another project and it is very easy to use with the `actions-rs/cargo` action we already use. About the compile error from the issue: We should just remove this function altogether, because the `Other` variant does not exist anyways

dbrgn:
> In #274, I noted that there is the possibility of using `cross` which cross-compiles using docker containers.

Does that just build for ARM, or actually on an Android device?

niklasmohrin:
You can choose any of the rust targets (for example [this list](https://github.com/niklasmohrin/clircle/blob/main/.github/workflows/ci.yml#L36)), we would probably be most interested in `aarch64-linux-android` and friends. If I understand it correctly, cross uses a docker container for cross compiling and qemu to run tests - or maybe they also compile in qemu. If we just build without testing, we don't need to worry about any ndk stuff, otherwise, we may have to a little bit of code to hook into that. I don't really know anything about android development, but that's just another reason why I don't want to host these specifics in our repo :D

niklasmohrin:
@leleliu008 Do you want to continue working on this, or should we close this PR for now?

leleliu008:
Since the feature has been implemented in #274, this is no need to go on. Please feel free to close it.

PR #266, opened 2022-01-30
Implement `cache_dir` config file option

### LINKED ISSUE #156: Add cache directory (or tldr checkout directory) to config

niklasmohrin:
Closes #156, supersedes #212, unless that catches up and gets merged first :)

From the commit message:

> This required that the `Cache` always holds a `PathBuf` with its base
> path and a cached `PathBuf` of the pages directory. This implies that
> some cache operations (such as clearing) are now instance methods
> instead of class methods, because the instance variables are used
> (we always re-resolved the paths before).
> 
> On the way, I stumbled upon some oddities and added `FIXME`s. Please let
> me know what you think about them :)

TODO
- [x] deprecate env var
- [ ] ~~`paths` module?~~ another time

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -164,12 +164,17 @@ impl Default for RawUpdatesConfig {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 struct RawDirectoriesConfig {
     #[serde(default)]
+    pub cache_dir: Option<PathBuf>,
+    #[serde(default)]
     pub custom_pages_dir: Option<PathBuf>,
 }
 
 impl Default for RawDirectoriesConfig {
     fn default() -> Self {
         Self {
+            cache_dir: None,
+            // FIXME: should this "business-logic" really be in `*Raw*DirectoriesConfig`?
+            // To me, this seems odd, especially since it can still be `None` afterwards 🤔
             custom_pages_dir: get_app_root(AppDataType::UserData, &crate::APP_INFO)
                 .map(|path| {
                     // Note: The `join("")` call ensures that there's a trailing slash
@@ -238,6 +243,7 @@ pub struct UpdatesConfig {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct DirectoriesConfig {
+    pub cache_dir_override: Option<PathBuf>,
     pub custom_pages_dir: Option<PathBuf>,
 }
 
@@ -270,6 +276,7 @@ impl From<RawConfig> for Config {
                 ),
             },
             directories: DirectoriesConfig {
+                cache_dir_override: raw_config.directories.cache_dir,
                 custom_pages_dir: raw_config.directories.custom_pages_dir,
             },
         }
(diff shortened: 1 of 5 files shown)


niklasmohrin:
I noticed that while developing, I have to think about preconditions / invariants of the cache directory a lot. Most recently, I noticed that in `list_pages` we dont error if the cache doesn't exist and in `clear` we check for existence again. Apart from the mental overhead, this is also unnecessary work. At first I thought we could wrap each of the invariants  into a `Once` so that we can just sprinkle them everywhere without performance drawbacks, but this would also make every function fallible. I now think that we should use this pattern:
```rust
struct Cache<State: CacheState> {
    base_path: PathBuf,
    state: State,
}

struct Unverified {
    pages_path: Lazy<PathBuf>,
    base_metadata: Lazy<Metadata>,
    pages_metadata: Lazy<Metadata>,
}
struct Verified {
    pages_path: PathBuf,
    base_metadata: Metadata,
    pages_metadata: Metadata,
}

impl Cache<Unverified> {
    pub fn verify(self) -> Cache<Verified> { ... }
}
```
This way, we can lock some operations behind a verification step on the type level and make a clear distinction which steps are needed to be able to perform a certain operation on the cache. (I thought the pattern was called type-state-pattern, but I couldn't find it again; I think I remember `rocket` used it as a creational pattern at some point)

src/config.rs
```diff
@@ -163,13 +163,18 @@ impl Default for RawUpdatesConfig {
...
     #[serde(default)]
     pub custom_pages_dir: Option<PathBuf>,
 }
 
 impl Default for RawDirectoriesConfig {
     fn default() -> Self {
         Self {
+            cache_dir: None,
+            // FIXME: should this "business-logic" really be in `*Raw*DirectoriesConfig`?
+            // To me, this seems odd, especially since it can still be `None` afterwards 🤔
```
dbrgn, line 177: You're right. This would be addressed in the latest commit in #276 ("Move custom_pages_dir business logic from RawConfig to Config").

dbrgn:
Your suggestion to use the type-state pattern sounds interesting, but is there any reason to construct a cache that isn't backed by a valid directory?

I think the invariant should be:

- Once the `Cache` constructor is finished, the cache directory must exist, be a directory and be writable
- Tealdeer is a run-once program. If permissions change between the creation of the cache and the actual command, an error or even a panic is fine

niklasmohrin:
Closing in favor of #276

PR #273, opened 2022-05-28
Auto-generate shell completions with `clap_complete`

cyqsimon:
Having the completion scripts auto-generated by `clap_complete` makes far more sense than manually maintaining the completion scripts IMO.


cyqsimon:
I think CI needs fixing

dbrgn:
This issue was tracked in #220 and a resolution attempt was started in #256. However, we noticed that the generated completions do not support dynamically generated values (i.e. completing the TLDR pages that are installed on the system). I haven't yet checked your proposal, does it solve these issues?

niklasmohrin:
Thank you for contributing! This topic has come up before, see #220. Back then we found that it isn't possible to have completion for the page names, but only the flags. We feel like this is a pretty big cost and moved forward with our hand crafted completion files. if this still isn't possible, I would stick to our decision from then. :/

cyqsimon:
Sorry I haven't considered this feature. I'll investigate if it's possible to integrate this feature with `clap_complete`.

cyqsimon:
After some digging around, I do concur with your conclusions in #256.

Although it's certainly possible (and quite easy TBH) to have `clap_complete` generate scripts, then modify them with a patch script, it's another piece of puzzle that requires maintenance and frankly very hacky. It's probably not worth going down that rabbit hole either.

Therefore I do agree that manually maintaining the completion scripts are the best solution for now. Closing PR.

PR #279, opened 2022-07-05
Search pages in all platforms

### LINKED ISSUE #278: Search other Platforms when command isn't found
Taken from the [TLDR client specification](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md):

> If a page is not available for either the host platform or the common platform, then clients SHOULD search other platforms and display a page from there - along with a warning message.

On Windows, `tealdeer` outputs this:

```console
$ tldr fuser
Warning: Page `fuser` not found in cache.
```

This lead me to believe that `fuser` didn't have a page, and I mistakenly submitted a [PR](https://github.com/tldr-pages/tldr/issues/8160)

I suspect that I'm not the only person to have had this problem, and I would appreciate if we could get this fixed.

marchersimon:
Fixes #278

I don't know if the implementation is efficient, but at least it feels logical.
What I've done:

- I created a `PriorityQueue`
- The current platforms has the highest priority, followed by `common`, and then all others
- Search for a page in this order, until it is found

It seems to work and doesn't break anything (hopefully)

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -13,6 +13,7 @@ use log::debug;
 use reqwest::{blocking::Client, Proxy};
 use walkdir::{DirEntry, WalkDir};
 use zip::ZipArchive;
+use priority_queue::PriorityQueue;
 
 use crate::types::{PathSource, PlatformType};
 
@@ -290,15 +291,27 @@ impl Cache {
 
         // Try to find a platform specific path next, append custom patch to it.
         let platform_dir = self.get_platform_dir();
-        if let Some(page) =
-            Self::find_page_for_platform(&page_filename, &cache_dir, platform_dir, &lang_dirs)
-        {
-            return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
-        }
 
-        // Did not find platform specific results, fall back to "common"
-        Self::find_page_for_platform(&page_filename, &cache_dir, "common", &lang_dirs)
-            .map(|page| PageLookupResult::with_page(page).with_optional_patch(patch_path))
+        let mut platforms = PriorityQueue::new();
+        platforms.extend(vec![
+            ("linux", 0),
+            ("osx", 0),
+            ("sunos", 0),
+            ("windows", 0),
+            ("android", 0),
+            ("common", 1)].clone());
+        // update the priority of the current platform to 2
+        platforms.push(platform_dir, 2);
+        
+        // Cycle through all platforms in this order: current_platform - common - all other platforms
+        for platform in platforms.into_sorted_iter() {
+            if let Some(page) =
+                Self::find_page_for_platform(&page_filename, &cache_dir, platform.0, &lang_dirs)
+                {
+                    return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
+                }
+        }
+        return None;
     }
 
     /// Return the available pages.
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -13,7 +13,6 @@ use log::debug;
 use reqwest::{blocking::Client, Proxy};
 use walkdir::{DirEntry, WalkDir};
 use zip::ZipArchive;
-use priority_queue::PriorityQueue;
 
 use crate::types::{PathSource, PlatformType};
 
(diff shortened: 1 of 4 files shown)


niklasmohrin:
Thanks! This seems like a good change, I have some comments about implementation though. Firstly, I don't think we need an efficient priority queue - these smart data structures are sometimes even slower for small inputs than just doing the straight-forward thing (see for example `Vec::clear_duplicates` in `extensions.rs`). Using a vector and just sorting it by priority is probably faster and doesn't need the dependency. Secondly, you repeat the possible platform dirs, but this should really be something like `Platform::all()`. This way, we don't duplicate this knowledge between multiple places.

In theory, we are doing extra IO work because we are checking the current platform and `common` twice, but I think this shouldn't make much of a difference (and if it does, we can fix that in another PR).

Please let us know if you need any help or have questions, note that both @dbrgn and I might take some time to respond (like with this PR)

marchersimon:
@niklasmohrin @dbrgn I've now made the changes

src/cache.rs
```diff
@@ -289,16 +289,25 @@ impl Cache {
...
+                Self::find_page_for_platform(&page_filename, &cache_dir, platform, &lang_dirs)
+                {
+                    return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
+                }
         }
-
-        // Did not find platform specific results, fall back to "common"
-        Self::find_page_for_platform(&page_filename, &cache_dir, "common", &lang_dirs)
-            .map(|page| PageLookupResult::with_page(page).with_optional_patch(patch_path))
+        return None;
```
niklasmohrin, line 310: `clippy` tells us that this `return` is unneeded. In general, when handling `Some`, `None`, iterators, etc. manually, there is a good chance that there is already some helper function from the standard library that does exactly what we want. In this case, it is `find_map`. Using that, we can express the above as:
```rust
directories_to_search
    .into_iter()
    .find_map(|platform_dir| {
        Self::find_page_for_platform(&page_filename, &cache_dir, platform_dir, &lang_dirs)
    })
    .map(|page| PageLookupResult::with_page(page).with_optional_patch(patch_path))
```

This is nice! But there is one more flaw in there. We are using `into_iter` on our freshly created vector. Maybe we can spare the allocation? Of course, the creation of our `directories_to_search` is basically just building an iterator and collecting it, so we can just skip the "collecting" step. Finally, let's use this opportunity to also give names to the iterators themselves and remove the comments - now, everything is clear from the code:

```rust
let main_directories = [current_platform_dir, "common"];
let secondary_directories = PlatformType::all()
    .iter()
    .copied()
    .filter(|&p| p != self.platform)
    .map(Self::directory_name_for);
let mut directories_to_search = main_directories
    .iter()
    .copied()
    .chain(secondary_directories);
```
Now, we can remove the `into_iter` below and arrive at what I believe to be the optimal solution.

niklasmohrin (commented):
Sorry for the long wait. This looks simpler already, nice work :)

I think this is 100% there logically, however I feel like we can improve the readability a bit when employing some Rust patterns and helpers. Please see my comments below and feel free to ask if something is unclear or if you disagree with any of my suggestions :)

marchersimon:
Thanks for the extensive feedback @niklasmohrin. Unfortunately this is where my Rust knowledge ends. I haven't really worked with slices, lifetime specifiers or other Rust patters yet. I'm still learning and it could take some while 'till i get there. But if you want to, you can finish my PR.

niklasmohrin:
No worries, this is what review is for :) You can take your time if you want to move forward with the PR and understand the proposed changes. Feel free to ask questions at any time, we might take some time to respond though.

if you really dont want to, I can also apply the review. Please let me know which way you prefer :)

matthew-e-brown:
I just ran into this issue myself: I use a Windows machine for work, but spend a lot of time SSH'd into remote Linux boxes. So, things like `tldr a2enmod` don't appear unless I want to manually `-p linux` each time, or install and run tealdeer on the remote Linux box, which isn't always feasible.

I've given a quick read through this PR, and I must commend @niklasmohrin on his review style—very warm feedback for somebody who is/was new to Rust. So, @marchersimon, I'd like to add my encouragement for you to come back and move forward on this PR now that a few months have passed. How do you feel about moving forwards? It would be great to see it implemented. I have faith in you! 😄

I would of course be willing to help as well, or even take it over from you if that's what you'd like. As someone with ties to teaching, though, I would much prefer to see it be done by somebody who is/was using it as a chance to learn a new language.

Cheers. 🙂

niklasmohrin:
@matthew-e-brown Due to the inactivity, I think it would be fair for you to take this PR over if you want to

niklasmohrin:
I will close this PR, so that this can be finished in another one. If anyone is basing their work no this, consider adding a `Co-authored-by: Simon Marcher <marchersimon@zohomail.eu>` for partial credit (unless you wouldn't want that @marchersimon) :)

PR #280, opened 2022-07-11
added missing fish completion for language

bagohart:
I tried to use this project and noticed that not all flags have tab completion for fish, so I added the missing one for `--language`.
The one for `--config-path` is missing, too, but since the flag is deprecated, it shouldn't be included anyway.


niklasmohrin (approved):
Thank you!

PR #284, opened 2022-08-21
Switch to 2021 Edition and maintenance

niklasmohrin:
I saw that CI was failing, see commit messages for changes and rationale.


dbrgn (approved):
Looks great, thanks!

PR #285, opened 2022-09-08
Add custom pages to list output

### LINKED ISSUE #205: `tldr --list` should include customize pages
`tldr --list` should include customize pages. It helps zsh/bash autocomplete.

Olavhaasie:
Fix #205 

I have added custom pages to the output of list by reading from the `custom_pages_dir` from the config and walking that directory for any pages. This also includes custom pages in the autocompletion.

This is my first contribution, so please let me know if something has to be changed or added :innocent:

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -302,7 +302,7 @@ impl Cache {
     }
 
     /// Return the available pages.
-    pub fn list_pages(&self) -> Result<Vec<String>> {
+    pub fn list_pages(&self, custom_pages_dir: Option<&Path>) -> Result<Vec<String>> {
         // Determine platforms directory and platform
         let (cache_dir, _) = Self::get_cache_dir()?;
         let platforms_dir = cache_dir.join(TLDR_PAGES_DIR).join("pages");
@@ -328,7 +328,7 @@ impl Cache {
         let mut pages = WalkDir::new(platforms_dir)
             .min_depth(1) // Skip root directory
             .into_iter()
-            .filter_entry(|e| should_walk(e)) // Filter out pages for other architectures
+            .filter_entry(should_walk) // Filter out pages for other architectures
             .filter_map(Result::ok) // Convert results to options, filter out errors
             .filter_map(|e| {
                 let path = e.path();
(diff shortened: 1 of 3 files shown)


src/cache.rs
```diff
@@ -341,6 +341,31 @@ impl Cache {
...
+            let is_page = |entry: &DirEntry| -> bool {
+                let path = entry.path();
+                let extension = &path.extension().and_then(OsStr::to_str).unwrap_or("");
+                entry.file_type().is_file() && extension == &"page"
+            };
+            let to_stem = |entry: DirEntry| -> Option<String> {
+                entry
+                    .path()
+                    .file_stem()
+                    .and_then(|stem| stem.to_str().map(Into::into))
```
niklasmohrin, line 355:
```suggestion
                    .file_stem()
                    .and_then(OsStr::to_str)
                    .map(str::to_string)
```

src/cache.rs
```diff
@@ -341,6 +341,31 @@ impl Cache {
...
             })
             .collect::<Vec<String>>();
+
+        if let Some(custom_pages_dir) = custom_pages_dir {
+            let is_page = |entry: &DirEntry| -> bool {
+                let path = entry.path();
+                let extension = &path.extension().and_then(OsStr::to_str).unwrap_or("");
+                entry.file_type().is_file() && extension == &"page"
+            };
+            let to_stem = |entry: DirEntry| -> Option<String> {
```
niklasmohrin, line 351: We could move this above and use this function in the `pages` `filter_map` (line 334) too (we don't need the `let path = entry.path()` anymore then and can just use `entry.path()` in the line below)

src/cache.rs
```diff
@@ -341,6 +341,31 @@ impl Cache {
...
+            let mut custom_pages = WalkDir::new(custom_pages_dir)
+                .min_depth(1)
+                .max_depth(1)
+                .into_iter()
+                .filter_entry(is_page)
+                .filter_map(Result::ok)
+                .filter_map(to_stem)
+                .collect::<Vec<String>>();
+
+            pages.append(&mut custom_pages);
```
niklasmohrin, line 366: This is a little anti-pattern: We collect into a vector, only to consume the elements right afterwards. We can let `custom_pages` just be the iterator we construct (by _not_ `collect`ing) and extend `pages` from that using `pages.extend(custom_pages);`. This way, we don't have the overhead of allocating the vector :)

src/cache.rs
```diff
@@ -341,6 +341,31 @@ impl Cache {
                 }
             })
             .collect::<Vec<String>>();
+
+        if let Some(custom_pages_dir) = custom_pages_dir {
+            let is_page = |entry: &DirEntry| -> bool {
+                let path = entry.path();
+                let extension = &path.extension().and_then(OsStr::to_str).unwrap_or("");
+                entry.file_type().is_file() && extension == &"page"
```
niklasmohrin, line 349: This can be simplified: We dont need to convert to `str` from `OsStr`, because `OsStr` has an `impl PartialEq<&str>` already. Instead of `unwrap_or("")`, we can use `unwrap_or_default` to get the empty `OsStr`. Finally, we don't need to borrow both sides of the equality check, like so:
```suggestion
                let extension = entry.path().extension().unwrap_or_default();
                entry.file_type().is_file() && extension == "page"
```

niklasmohrin (changes requested):
Thank you for your submission! I think that you hit the right spot of the codebase with your addition and that the change is useful. I left some comments about small implementation details which I would like to see added, other than that the code looks good. If you feel like it, you can also apply some of my comments to the rest of the function, as they also apply there, but don't feel obliged to do more than you came to do. Finally, I think we should add an integration test to verify that the new feature is working as intended (and to have this check in the future so that we don't break it again). To do so, we already have some setup in `tests/lib.rs`, you can add a test that checks that a custom page shows up in `tldr --list`. If you need any help, feel free to reach out :)

Olavhaasie:
Thanks for the useful feedback! It really reduced the complexity of the code and I learned something new.
As you said I also applied some changes to the already existing code, similar to the ones you suggested. Also, I added a test case for listing custom pages.

niklasmohrin (approved):
@dbrgn ?

dbrgn (approved):
Looks great! Thanks @Olavhaasie for the contribution and @niklasmohrin for the review!

PR #286, opened 2022-09-16
Add a `--edit` option

BuonOmo:
This option will help contributing, as the _edit this page on github_ button does on https://tldr.ostera.io/man.

This is a minimum viable product. If you feel this is interesting, I could go further and open an editor if the page is a local page or if it contains a patch. To do so, I was thinking of refactoring a tiny bit `PageLookupResult` to contain those information.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -85,6 +85,15 @@ pub(crate) struct Args {
     )]
     pub markdown: bool,
 
+    /// Open file on Github for editing
+    #[clap(
+        long = "edit",
+        short = 'e',
+        requires = "command",
+        conflicts_with = "raw"
+    )]
+    pub edit: bool,
+
     /// Suppress informational messages
     #[clap(short = 'q', long = "quiet")]
     pub quiet: bool,
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,7 +45,7 @@ use crate::{
     cli::Args,
     config::{get_config_dir, get_config_path, make_default_config, Config},
     extensions::Dedup,
-    output::print_page,
+    output::{open_browser, print_page},
     types::{ColorOptions, PlatformType},
     utils::{print_error, print_warning},
 };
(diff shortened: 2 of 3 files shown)


niklasmohrin (changes requested):
Thank you for your contribution! Here are my thoughts on this:

1. I don't think we want to open the browser, we rejected automatic folder creation some time ago because we felt that it was unexpected from a tool like tealdeer. I feel the same about navigating to a webpage; if we have this flag, let's just let it display the link. Modern terminal emulators let you click on it and open it in your preferred browser (with any additional arguments that might be needed). (same for opening an editor)
2. I am a bit worried that we could run into conflicts with proposed flags from [the client specification](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md) later if they ever decide that there should be a `--edit` flag (see for example https://github.com/tldr-pages/tldr/issues/1073). I would prefer if we / you first started a discussion at tldr-pages/tldr about possible recommendations from the spec, so that they at least know about our plans. Best case would be if a full description of what the command should do (that still allows us to show paths to custom pages etc) could be worked out before merging an implementation here
3. (once this is settled, we can take a closer look at the implementation)

Hence, I am "requesting changes" for now.

BuonOmo:
@niklasmohrin definitely agree with all of your points. I don't think I'll have the time to dive into specification discussion. I'll just ping them there about this idea to give them more insights on how it could be done.

I could also, as an alternative proposition, just add the link in the presentation of the command directly:

```
$ tldr command
  Command forces the shell to execute the program and ignore any functions, builtins and aliases with the same name.
  More information: <https://manned.org/command>.

  Execute the `ls` program literally, even if an `ls` alias exists:

      command ls

  Display the path to the executable or the alias definition of a specific command:

      command -v command_name

Edit this page on GitHub: https://github.com/tldr-pages/tldr/edit/main/pages/common/command.md
```

niklasmohrin:
That's okay :) About appending it to the output: My first impression is that the tldr pages are explicitly designed to be compact and I would hate it if users would have to scroll up one line just because we have this "contribution advertisement". As a user, this could possibly annoy me into switching to another client. I am against changing the default behavior in this way.

Another thought for the spec discussion: This output could be an idea for `--edit`. We could make it mean that the flag doesn't work like a subcommand, but rather an additional option to the default behavior (i.e. `tldr command` is unchanged and `tldr --edit command` outputs what you wrote above). This would also allow for having this flag permanently enabled by config or so.

_However_, I think I like the initial proposal better, because we can then put more information about editing, not just one line. :D

BuonOmo:
Closing for now, we'll see how specs evolve. In the mean time I'll be using my branch, and anyone in favor of that could. I'll update it if requested

PR #288, opened 2022-09-24
Replace ansi_term

dbrgn:
ansi_term is not actively maintained anymore (see https://rustsec.org/advisories/RUSTSEC-2021-0139). Replace it with a suitable alternative.

I looked at termcolor, owo-colors and yansi, and ended up picking yansi:

- It is very simple
- It does not do terminal color support checking (we already do that in tealdeer)
- Its color enum type is almost identical to the one from ansi_term, so migrating is easy
- It's developed by Sergio Benitez (of Rocket fame)

The only change from a config API point of view is that "purple" is now renamed to "magenta", but "purple" still works.

(Based on #287, merge that one first!)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,15 +18,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.65"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1111,7 +1102,6 @@ dependencies = [
 name = "tealdeer"
 version = "1.5.0"
 dependencies = [
- "ansi_term",
  "anyhow",
  "app_dirs2",
  "assert_cmd",
@@ -1129,6 +1119,7 @@ dependencies = [
  "tempfile",
  "toml",
  "walkdir",
+ "yansi",
  "zip",
 ]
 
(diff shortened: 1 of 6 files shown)


dbrgn:
Note: We could replace the custom enable_colors logic with [Paint::disable()](https://docs.rs/yansi/latest/yansi/#disabling), but we'd have to test that it doesn't have unintended interactions with our tests (due to global state).

niklasmohrin:
Beyond tests: Does the `disable` functionality not have any performance impact? All methods that actually produce bytes from a `Paint` have to load from an atomic boolean before deciding how to paint :thinking: Let's measure before merging

dbrgn:
Yes, that's true. I suspect there's no significant impact on real-world use cases like rendering a tldr page. I'm fairly sure the overhead will be insignificant compared to the I/O we're doing. However, I'll try to measure it.

dbrgn:
```
[main] $ cargo build --release
[main] $ mv target/release/tldr tldr-ansi-term
[main] $ git switch replace-ansi-term
[replace-ansi-term] $ cargo build --release
[replace-ansi-term] $ mv target/release/tldr tldr-yansi
```

...and then:

screenshot-20220925-174600

If you run the benchmark multiple times, you get varying results. I'd say that the change is not significant. (I didn't expect it to be.)

niklasmohrin:
Okay, I also reproduced these performance results :+1:

niklasmohrin:
I wish they had Cargo features to remove a) the wrapping and b) the enable/disable stuff because I don't want to think about that :D

niklasmohrin (approved):

niklasmohrin:
About using `disable`: I think I had some thoughts about `OutputManager` some time back to better separate concerns with IO, but never came to implement / design it to the end

dbrgn:
> I wish they had Cargo features to remove a) the wrapping and b) the enable/disable stuff because I don't want to think about that :D

Feel free to open an issue on the yansi issue tracker 🙂

PR #290, opened 2022-10-01
Remove deprecated commands

dbrgn:
Three commands were deprecated in the last release of tealdeer (1.5.0, almost a year ago) and can be removed now:

- `--config-path` (use `--show-paths` instead)
- `-o / --os` (use `-p/--platform` instead)
- `-m / --markdown` (use `-r/--raw` instead)


dbrgn:
Merging without review, I assume this will not be controversial.

PR #293, opened 2022-10-01
Documentation fixes

dbrgn:
Some leftovers, fixes and improvements.

### DIFF
--- a/docs/src/installing.md
+++ b/docs/src/installing.md
@@ -40,10 +40,6 @@ Build and install the tool via cargo...
 
     $ cargo install tealdeer
 
-*(Note: You might need to install OpenSSL development headers, otherwise you get
-a "failed to run custom build command for openssl-sys" error message. The
-package is called `libssl-dev` on Ubuntu.)*
-
 ## Build From Source
 
 Debug build with logging enabled:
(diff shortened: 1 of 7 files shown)


docs/src/config_directories.md
```diff
@@ -11,7 +11,7 @@ exist, it will be created.
     [directories]
     cache_dir = "/home/myuser/.tealdeer-cache/"
 
-If no `cache_dir` is specified, tealdeer will fall back to a location that
```
niklasmohrin, line 14: Are you sure? So far I thought we have the lowercased word as the "brand name"

niklasmohrin (approved):

docs/src/installing.md
```diff
@@ -40,17 +40,17 @@ Build and install the tool via cargo...
 
 ## Build From Source
 
-Debug build with logging enabled:
```
niklasmohrin, line 47: Hm, this reads as if you have to set it for building, not at run time; unless that really is the case, I think the previous note was better

docs/src/config_directories.md
dbrgn: It's a bit inconsistent 🙈 In the docs, so far we always capitalized the name (except for this line above). In the README, we did not. I don't really mind either way, do you have a preference? I'd slightly tend towards capitalizing it as a proper name (looks less strange at the beginning of a sentence).

docs/src/installing.md
dbrgn: Fair point. What about the fixupped version?

niklasmohrin: Looks good

docs/src/config_directories.md
dbrgn: > I think I would go with lowercase in general, but capitalize at the beginning of a sentence.

Works for me.

Issue #268, opened 2022-02-13
The `--quiet` flag doesn't silence all warnings


niklasmohrin:
In fish, when typing `tldr t<Tab>`, I get additional output printed to stderr:
```plain
tldr tThe cache hasn't been updated for 34 days.
You should probably run `tldr --update` soon.

tabula          textql            tlmgr-update         trap
tac             theharvester      tlp                  trash
tail            thunar            tlp-stat             trash-cli
tailscale       thunderbird       tmpmail              travis
```

This is because we use a plain `tldr --list` for the completion. The zsh script avoids this by redirecting `2>/dev/null`, the bash one either has the same problem or bash completions automatically silence this, I haven't checked. I suppose `2>/dev/null` is okay, but I would expect `--quiet` to suffice for this and prefer using that instead.

newsch:
I can confirm this is a problem with the bash completion as well:
```
$ source completion/bash_tealdeer
$ touch -d '31 days ago' ~/.cache/tealdeer/tldr-pages
$ tldr tThe cache hasn't been updated for 31 days.
You should probably run `tldr --update` soon.
```

FWIW, `-q` doesn't suppress the "Cache not found. Please run `tldr --update`." message.

Walker-00:
Is this been fixed?, both "fish" and "bash" works fine for me. Shouldn't this close?

niklasmohrin:
Completion seems to work for some other reason, but the original point that `tldr --quiet` should not display warnings or errors still stands

sidsri14:
The remaining warning leak is fixed: the old-naming custom pages warning now respects --quiet (PR #520), with a regression test covering both modes.

PR #295, opened 2022-10-04
Add auto generate completions for some shells

SergioRibera:
This is a requested function that solves problem #268 by generating automatic autocompletions and allows updating the cache

- [x] Check format
- [x] Check clippy
- [x] test commands implementation

### Supported Shells
- Bash
- Zsh
- Fish
- Powershell
- Elvish

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,6 +143,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "3.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f7a2e0a962c45ce25afce14220bc24f9dade0a1787f185cecf96bfba7847cd8"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "3.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


dbrgn:
Thanks! This issue (auto-generated completions) was originally tracked in https://github.com/dbrgn/tealdeer/issues/220 and a resolution attempt was started in https://github.com/dbrgn/tealdeer/pull/256 and https://github.com/dbrgn/tealdeer/pull/273. However, we noticed that the generated completions do not support dynamically generated values (i.e. completing the TLDR pages that are installed on the system). Your proposal probably doesn't have a solution for dynamic completions (#256) either, right?

SergioRibera:
> Thanks! This issue (auto-generated completions) was originally tracked in #220 and a resolution attempt was started in #256 and #273. However, we noticed that the generated completions do not support dynamically generated values (i.e. completing the TLDR pages that are installed on the system). Your proposal probably doesn't have a solution for dynamic completions (#256) either, right?

You are right, I guess I had not seen that, but I think I can implement it with the base I have so far, I will update the PR and let you know if I get something or not to close this PR, thank you very much.

niklasmohrin:
@SergioRibera Do you want to keep working on this, or should we close this PR for now?

SergioRibera:
I've been busy with other things, so I think you can close the PR for now.

niklasmohrin:
Alright, no problem :)

PR #296, opened 2022-10-11
Docs: improve grammar

CyrusYip:


niklasmohrin (approved):
Thank you!

dbrgn:
Nice catch - even in the diff of this PR I only realized the mistake after reading it three times 😂

PR #297, opened 2022-10-22
Fix path source for custom pages dir

dbrgn:
The two values were inverted 😂

I think when this is merged, I'll do an 1.6.1 release, since this bug can be a bit confusing to users.


niklasmohrin (approved):

dbrgn:
Thanks for the reminder, almost forgot about it 🙃

PR #298, opened 2022-10-22
Migrate to Clap v4, bump MSRV to 1.75

### LINKED ISSUE #350: Support for clap version 4
Hi.

I am part of the Debian rust team and we would like to reduce the number of clap versions we are shipping. Is there any chance of an update to support clap version 4?

dbrgn:
Again, this is a pretty big update with multiple breaking changes that requires changing how our argument parsing is set up. (If I ignore the effort to upgrade, this update does contain some nice API improvements, so I guess in the long term it's worth it. The blogpost [is an interesting read](https://epage.github.io/blog/2022/09/clap4/).)

Fixes #350.

### DIFF
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -1,30 +1,27 @@
-tealdeer 1.6.1
+tealdeer 1.6.1: A fast TLDR client
 Danilo Bargen <mail@dbrgn.ch>, Niklas Mohrin <dev@niklasmohrin.de>
-A fast TLDR client
 
-USAGE:
-    tldr [OPTIONS] [COMMAND]...
+Usage: tldr [OPTIONS] [COMMAND]...
 
-ARGS:
-    <COMMAND>...    The command to show (e.g. `tar` or `git log`)
+Arguments:
+  [COMMAND]...  The command to show (e.g. `tar` or `git log`)
 
-OPTIONS:
-    -l, --list                    List all commands in the cache
-    -f, --render <FILE>           Render a specific markdown file
-    -p, --platform <PLATFORMS>    Override the operating system [possible values: linux, macos,
-                                  windows, sunos, osx, android, freebsd, netbsd, openbsd]
-    -L, --language <LANGUAGE>     Override the language
-    -u, --update                  Update the local cache
-        --no-auto-update          If auto update is configured, disable it for this run
-    -c, --clear-cache             Clear the local cache
-        --pager                   Use a pager to page output
-    -r, --raw                     Display the raw markdown instead of rendering it
-    -q, --quiet                   Suppress informational messages
-        --show-paths              Show file and directory paths used by tealdeer
-        --seed-config             Create a basic config
-        --color <WHEN>            Control whether to use color [possible values: always, auto,
-                                  never]
-    -v, --version                 Print the version
-    -h, --help                    Print help information
+Options:
+  -l, --list                  List all commands in the cache
+  -f, --render <FILE>         Render a specific markdown file
+  -p, --platform <PLATFORMS>  Override the operating system [possible values: linux, macos, sunos,
+                              windows, android, freebsd, netbsd, openbsd]
+  -L, --language <LANGUAGE>   Override the language
+  -u, --update                Update the local cache
+      --no-auto-update        If auto update is configured, disable it for this run
+  -c, --clear-cache           Clear the local cache
+      --pager                 Use a pager to page output
+  -r, --raw                   Display the raw markdown instead of rendering it
+  -q, --quiet                 Suppress informational messages
+      --show-paths            Show file and directory paths used by tealdeer
+      --seed-config           Create a basic config
+      --color <WHEN>          Control whether to use color [possible values: always, auto, never]
+  -v, --version               Print the version
+  -h, --help                  Print help
 
 To view the user documentation, please visit https://dbrgn.github.io/tealdeer/.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,32 +2,38 @@
 
 use std::path::PathBuf;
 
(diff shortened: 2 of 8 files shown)


dbrgn:
I think everything should work fine now, but the PR is currently blocked on #352.

docs/src/usage.txt
```diff
@@ -1,30 +1,26 @@
...
-        --seed-config             Create a basic config
-        --color <WHEN>            Control whether to use color [possible values: always, auto,
-                                  never]
-    -v, --version                 Print the version
-    -h, --help                    Print help information
+Options:
+  -l, --list                  List all commands in the cache
+  -f, --render <FILE>         Render a specific markdown file
+  -p, --platform <PLATFORMS>  Override the operating system [possible values: linux, macos, windows,
+                              sunos, android]
```
dbrgn, line 13: The help now doesn't list aliases anymore. I changed the code to use "macos" as main name and "osx" as alias (not vice versa), so that "macos" is shown in the help by default. "osx" still works though.

docs/src/usage.txt
```diff
@@ -1,30 +1,27 @@
-tealdeer 1.6.1
+tealdeer 1.2.3: A fast TLDR client
 Danilo Bargen <mail@dbrgn.ch>, Niklas Mohrin <dev@niklasmohrin.de>
-A fast TLDR client
```
dbrgn, line 3: The default help output in clap now doesn't list version and author anymore.

I updated the help template to include those two again. The version is now combined with the description to get more compact output.

docs/src/usage.txt
zedseven: For it to show aliases, you now have to use `visible_alias`, if I recall correctly. You can have a mix of visible and invisible aliases.

src/cli.rs
```diff
@@ -36,61 +46,62 @@ pub(crate) struct Args {
...
+    #[arg(long = "seed-config")]
     pub seed_config: bool,
 
     /// Control whether to use color
-    #[clap(
+    #[arg(
         long = "color",
         value_name = "WHEN",
-        possible_values = ["always", "auto", "never"]
+        value_parser = EnumValueParser::<ColorOptions>::new(),
```
niklasmohrin, line 97: We don't need to be so verbose here, we can use `#[arg(value_enum)]` or even not specify anything and clap will figure this out on its own (same for the other value enum)
```suggestion
```

src/cli.rs
```diff
@@ -36,61 +46,62 @@ pub(crate) struct Args {
...
     )]
+    //possible_values = ["always", "auto", "never"]
```
niklasmohrin, line 99:
```suggestion
```

src/cli.rs
```diff
@@ -2,32 +2,42 @@
...
 
 // Note: flag names are specified explicitly in clap attributes
 // to improve readability and allow contributors to grep names like "clear-cache"
 #[derive(Parser, Debug)]
-#[clap(about = "A fast TLDR client", author, version)]
-#[clap(
-    after_help = "To view the user documentation, please visit https://dbrgn.github.io/tealdeer/."
+#[command(
+    about = "A fast TLDR client",
+    version = "1.2.3",
```
niklasmohrin, line 18: Clap can read the version from `Cargo.toml`:
```suggestion
    version,
```

niklasmohrin (commented):
Looks mostly good, some thoughts:

src/cli.rs
dbrgn: Oh, nice 🙂

src/types.rs
```diff
@@ -96,18 +99,16 @@ pub enum ColorOptions {
...
-    fn from_str(s: &str) -> Result<Self> {
-        match s {
-            "always" => Ok(Self::Always),
-            "auto" => Ok(Self::Auto),
-            "never" => Ok(Self::Never),
-            other => Err(anyhow!(
-                "Unknown color option: {}. Possible values: always, auto, never",
-                other
-            )),
+impl clap::ValueEnum for ColorOptions {
```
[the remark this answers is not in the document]
dbrgn: Ah, true. I can't really use it for `PlatformType` though, due to the aliases. (Unless there are enum field attributes, but the Clap documentation doesn't seem to mention those...)

docs/src/usage.txt
dbrgn: It seems that `visible_alias` is only available on commands and args, but not on `PossibleValue`.

dbrgn:
@niklasmohrin updated! Should be ready now, I think.

dbrgn:
MSRV had to be bumped due to clap dependencies. I picked 1.75, which was released in December 2023.

src/cli.rs
```diff
@@ -36,61 +42,56 @@ pub(crate) struct Args {
...
+    #[arg(long = "color", value_name = "WHEN")]
     pub color: Option<ColorOptions>,
 
     /// Print the version
     // Note: We override the version flag because clap uses `-V` by default,
     // while TLDR specification requires `-v` to be used.
-    #[clap(short = 'v', long = "version")]
-    pub version: bool,
+    #[arg(short = 'v', long = "version", action = ArgAction::Version)]
+    pub version: Option<bool>,
```
niklasmohrin, line 96: This is a bit weird, I would have thought that this type should only have two values, not three as `Option<bool>` - but `bool` and `Option<()>` both fail, so I guess this is the way to go?

niklasmohrin (approved):
Looks good :)

src/cli.rs
dbrgn: Thanks to GitHub code search, I found many results with `Option<bool>`, but also some results that used `()`. And that actually works 🙂

Issue #45, opened 2018-09-01
Allow specifying multiple operating systems


jedahan:
So sometimes I'm using tealdeer on my laptop which is macOS, and looking up a utility that is for linux on another box. I have to `-o linux` after seeing that the command is not found.

My workaround is `function tld { tldr $* || tldr -o linux $* }`

I was wondering how you'd feel making this the default behavior.

dbrgn:
Hm, what should happen if a command exists both on Linux and on SunOS (with different TLDR-pages), and you call `tldr` from macOS? :slightly_smiling_face:

jedahan:
Perfect is the enemy of good?

I don't like to make things more complicated, but if you are concerned about confusion with automatic fallback, maybe one of all of:
 * configurable fallback order as cli arguments with a default
 * an indicator of the current fallback at the top or bottom of the rendering
 * maybe making this an upstream thing tldr-pages/tldr with symlinks as fallbacks?

BrainMaestro:
Maybe it could fail, but let the user know that the page exists on another os?

dbrgn:
I'd definitely favor an option where you could specify multiple operating systems in order of precedence (`tldr -o osx,linux`). Now that we have configfile support, the default setting could even be configurable. Alternatively, you could alias `tldr` to `tldr -o osx,linux`.

niklasmohrin:
This issue was closed in #300 :)

PR #300, opened 2022-10-25
Allow querying multiple platforms

jj-style:
Fix for an old issue #45 - to allow specifying multiple `platform`s on the cli.
Behaviour is to try get the tldr page for each platform given - so if a page doesn't exist for one platform but does for another, you still get the page displayed.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -299,7 +299,7 @@ impl Cache {
         let platforms_dir = self.pages_dir().join("pages");
         let platform_dirs: Vec<&'static str> = platforms
             .iter()
-            .map(|&p| Cache::get_platform_dir(p))
+            .map(|&p| Self::get_platform_dir(p))
             .collect();
 
         // Closure that allows the WalkDir instance to traverse platform
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -20,7 +20,6 @@ static TLDR_OLD_PAGES_DIR: &str = "tldr-master";
 
 #[derive(Debug)]
 pub struct Cache {
-    platform: PlatformType,
     cache_dir: PathBuf,
 }
 
@@ -86,12 +85,11 @@ pub enum CacheFreshness {
 }
 
 impl Cache {
-    pub fn new<P>(platform: PlatformType, cache_dir: P) -> Self
+    pub fn new<P>(cache_dir: P) -> Self
     where
         P: Into<PathBuf>,
     {
         Self {
-            platform,
             cache_dir: cache_dir.into(),
         }
     }
@@ -211,8 +209,8 @@ impl Cache {
     }
 
     /// Return the platform directory.
-    fn get_platform_dir(&self) -> &'static str {
-        match self.platform {
+    fn get_platform_dir(platform: PlatformType) -> &'static str {
+        match platform {
             PlatformType::Linux => "linux",
             PlatformType::OsX => "osx",
             PlatformType::SunOs => "sunos",
@@ -247,6 +245,7 @@ impl Cache {
         name: &str,
         languages: &[String],
         custom_pages_dir: Option<&Path>,
+        platforms: &[PlatformType],
     ) -> Option<PageLookupResult> {
         let page_filename = format!("{}.md", name);
         let patch_filename = format!("{}.patch", name);
@@ -275,12 +274,14 @@ impl Cache {
 
         let patch_path = Self::find_patch(&patch_filename, custom_pages_dir);
 
-        // Try to find a platform specific path next, append custom patch to it.
-        let platform_dir = self.get_platform_dir();
-        if let Some(page) =
-            Self::find_page_for_platform(&page_filename, &pages_dir, platform_dir, &lang_dirs)
-        {
-            return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
+        // Try to find a platform specific path next, in the order supplied by the user, and append custom patch to it.
+        for &platform in platforms {
+            let platform_dir = Cache::get_platform_dir(platform);
+            if let Some(page) =
+                Self::find_page_for_platform(&page_filename, &pages_dir, platform_dir, &lang_dirs)
+            {
+                return Some(PageLookupResult::with_page(page).with_optional_patch(patch_path));
+            }
         }
 
         // Did not find platform specific results, fall back to "common"
@@ -289,10 +290,17 @@ impl Cache {
     }
 
     /// Return the available pages.
-    pub fn list_pages(&self, custom_pages_dir: Option<&Path>) -> Vec<String> {
+    pub fn list_pages(
+        &self,
+        custom_pages_dir: Option<&Path>,
+        platforms: &[PlatformType],
+    ) -> Vec<String> {
         // Determine platforms directory and platform
         let platforms_dir = self.pages_dir().join("pages");
-        let platform_dir = self.get_platform_dir();
+        let platform_dirs: Vec<&'static str> = platforms
+            .iter()
+            .map(|&p| Cache::get_platform_dir(p))
+            .collect();
 
         // Closure that allows the WalkDir instance to traverse platform
         // specific and common page directories, but not others.
@@ -303,7 +311,7 @@ impl Cache {
                 None => return false,
             };
             if file_type.is_dir() {
-                return file_name == "common" || file_name == platform_dir;
+                return file_name == "common" || platform_dirs.contains(&file_name);
             } else if file_type.is_file() {
                 return true;
             }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,13 +2,13 @@
 
 use std::path::PathBuf;
 
-use clap::{AppSettings, ArgGroup, Parser};
+use clap::{AppSettings, ArgAction, ArgGroup, Parser};
 
 use crate::types::{ColorOptions, PlatformType};
 
 // Note: flag names are specified explicitly in clap attributes
 // to improve readability and allow contributors to grep names like "clear-cache"
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, Clone)]
 #[clap(about = "A fast TLDR client", author, version)]
 #[clap(
     after_help = "To view the user documentation, please visit https://dbrgn.github.io/tealdeer/."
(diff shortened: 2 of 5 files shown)


src/main.rs
```diff
@@ -281,7 +281,10 @@ fn main() {
     }
 
     // Specify target OS
-    let platform: PlatformType = args.platform.unwrap_or_else(PlatformType::current);
+    let platforms: Vec<PlatformType> = match args.platform {
+        Some(ref p) => p.clone(),
+        None => vec![PlatformType::current()],
+    };
```
niklasmohrin, line 287: Hm, this seems wrong. We don't need ownership of the vector, a slice should totally suffice. I guess we are running into borrowing issues later, but we should fix those instead.

Depending on how difficult this turns out to be, we can also postpone a refactor, but I would not want this kind of fixing to pile up

src/main.rs
```diff
@@ -294,85 +297,95 @@ fn main() {
...
-    // Check cache presence and freshness
-    if !cache_updated
-        && (args.list || !args.command.is_empty())
-        && check_cache(&cache, &args, enable_styles) == CheckCacheResult::CacheMissing
-    {
-        process::exit(1);
-    }
+    for platform in platforms {
+        // Instantiate cache. This will not yet create the cache directory!
+        let cache = Cache::new(platform, &config.directories.cache_dir.path);
```
niklasmohrin, line 310: So far, we only ever had one cache for the whole program, which makes sense, considering that we have all of the tldr-pages checked out in one directory. Right now, this code has a bug where we check whether the same directory needs to be updated (or cleared) for every platform the a user passes. I think it also lists some commands more than once, if there are multiple versions for different platforms.

I think we should stick to only having one `cache`. About adding support for multiple platforms, we could either make `Cache` have its own vector of platforms, or, just pass the platforms when they are needed. I like the latter the most, it will also lead to a solution where we don't need an owned `Vec` and don't have lifetimes on `Cache` (only in the methods that will then take `platforms: &[Platform]`).

src/main.rs
```diff
@@ -294,85 +297,95 @@ fn main() {
...
-                .map(PathWithSource::path),
-        ) {
-            if let Err(ref e) =
-                print_page(&lookup_result, args.raw, enable_styles, args.pager, &config)
-            {
-                print_error(enable_styles, e);
-                process::exit(1);
-            }
+                .map(PathWithSource::path);
+            println!("{}", cache.list_pages(custom_pages_dir).join("\n"));
```
niklasmohrin, line 340: I wonder if we want to display some disambiguation here: What should the output be, if we run `tldr --platform linux --platform windows --list cd`? @dbrgn 

We should check the client specification from tldr-pages/tldr for that (or ask them for a recommendation)

tests/lib.rs
```diff
@@ -562,6 +562,34 @@ fn test_pager_flag_enable() {
...
 }
 
+#[test]
+fn test_multiple_platform_command_search() {
+    let testenv = TestEnv::new();
+    testenv.add_os_entry("linux", "linux-only", "this command only exists for linux");
+
+    testenv
+        .command()
+        .args(["--platform", "macos", "--platform", "linux", "linux-only"])
```
niklasmohrin, line 572: Do we want to guarantee behavior if the page occurs for multiple platforms? Maybe try platforms in the order they were specified? In that case, we would also need a test for that

niklasmohrin (changes requested):
Thank you for contributing! I have some change requests below, the main one is passing `platforms` to the methods that need it and not storing it in the `Cache`. This also includes handling multiple platforms in these methods instead of running them multiple times. Reviews might take a while here as @dbrgn and I are both relatively busy; feel free to ping us again here if we haven't checked in in a while or if you feel like we forgot you

jj-style:
Thank you for the code review and suggestions! I agree about passing in `platform` where it's needed and refactoring it out of the `Cache`.
I'll have a crack at fixing these issues, I'm still somewhat new to rust but hopefully should be doable and will be a good learning experience!

p.s. thanks for the awesome project - it saves my bacon multiple times a day!

niklasmohrin:
Feel free to ask about any problems you encounter, I will try to answer without unreasonable delays

jj-style:
Happy new year! Sorry it's taken me a while to respond to your feedback. I've given it an attempt to remove the `platform` from the `Cache` and pass in `&[PlatformType]` on the methods that need it. It definitely feels better than looping the main program for each platform and creating multiple instances of the `cache`

src/main.rs
```diff
@@ -280,8 +280,10 @@ fn main() {
         create_config_and_exit(enable_styles);
     }
 
-    // Specify target OS
-    let platform: PlatformType = args.platform.unwrap_or_else(PlatformType::current);
+    let platforms = match args.platform {
+        Some(ref platforms) => platforms.clone(),
+        None => vec![PlatformType::current()],
+    };
```
niklasmohrin, line 286: We can avoid the clone here by borrowing into the `Option`:
```suggestion
    const FALLBACK_PLATFORMS: &'static [PlatformType] = &[PlatformType::current()];
    let platforms = args
        .platform
        .as_ref()
        .map_or(FALLBACK_PLATFORMS, Vec::as_slice);
```

src/main.rs
```diff
@@ -295,7 +297,7 @@ fn main() {
     }
 
     // Instantiate cache. This will not yet create the cache directory!
-    let cache = Cache::new(platform, &config.directories.cache_dir.path);
+    let cache = Cache::new(config.directories.cache_dir.path.clone());
```
niklasmohrin, line 300: Is there any particular reason for the change from `&` to `.clone`?

src/main.rs
```diff
@@ -364,8 +370,8 @@ fn main() {
                     enable_styles,
                     &format!(
                         "Page `{}` not found in cache.\n\
-                         Try updating with `tldr --update`, or submit a pull request to:\n\
-                         https://github.com/tldr-pages/tldr",
+                            Try updating with `tldr --update`, or submit a pull request to:\n\
+                            https://github.com/tldr-pages/tldr",
```
niklasmohrin, line 374: Please revert the changes on these lines :)

src/cache.rs
```diff
@@ -289,10 +290,17 @@ impl Cache {
...
+        &self,
+        custom_pages_dir: Option<&Path>,
+        platforms: &[PlatformType],
+    ) -> Vec<String> {
         // Determine platforms directory and platform
         let platforms_dir = self.pages_dir().join("pages");
-        let platform_dir = self.get_platform_dir();
+        let platform_dirs: Vec<&'static str> = platforms
+            .iter()
+            .map(|&p| Cache::get_platform_dir(p))
```
niklasmohrin, line 302: Nit: This should use `Self` instead of `Cache` :)

niklasmohrin (commented):
Thanks, looks good :)

Some of my previous comments are still open, in particular what should happen with `--list`, and whether we want to guarantee that we respect the order the platforms were given in. For the latter, I think we can always prioritize platforms that came first (as we already do) - so we would just need a test that covers this. @dbrgn Do you have opinions on behavior and whether we want to try to incorporate this into the spec?

jj-style:
Thanks for the feedback! I have now implemented the suggestions you have made.

With regards to what should happen with `--list` when multiple platforms are provided - the current behaviour in `cache.rs` with the `pages.sort().dedup()` implies that the order the platforms are given is not important and the combined ordered set of pages should be listed.
I have added a test for this behaviour however, if we decide we want to list the pages grouped in order by the platforms supplied I can make this change

src/cli.rs
```diff
@@ -39,9 +39,10 @@ pub(crate) struct Args {
     #[clap(
         short = 'p',
         long = "platform",
+        action = ArgAction::Append,
         possible_values = ["linux", "macos", "windows", "sunos", "osx", "android"],
     )]
-    pub platform: Option<PlatformType>,
+    pub platform: Option<Vec<PlatformType>>,
```
niklasmohrin, line 45: I think it should now be:
```suggestion
    pub platforms: Option<Vec<PlatformType>>,
```
Can you verify that this change doesn't change the CLI argument to `--platforms` then? I think it should work like this though

niklasmohrin (commented):
Thanks, looks very good now! I have some final remarks, mostly about testing, below.

There is also [one comment from before](https://github.com/dbrgn/tealdeer/pull/300#discussion_r1008561492) that I think hasn't been addressed yet. A check should be added to make sure that when running `tldr --platform linux --platform windows xyz`, a page for linux should be preferred, even if `xyz` exists for both platforms. I think it could be tested in `test_multiple_platform_command_search`.

I thought about the behavior for `--list` some more and I think the current version is probably what we want. I am fine with the current behavior and if we ever need to change it, well, we can still change it :^)

src/cli.rs
jj-style: Good spot yeah I have updated and tested this. It is still `--platform` from the `clap` `long` annotation

jj-style:
thank you for the continuous feedback and help with this. I have pushed the changes for the testing :)

niklasmohrin (approved):
Looks good to me! Thank you for your contribution and sorry for the long wait :)

niklasmohrin (approved):
Sorry for the long wait! Your PR looks good and will be included in the next release. Thanks!

dbrgn:
Hi, some (very late, I know) follow-up questions @niklasmohrin @jj-style:

- Was it always clear that the API should be `tldr --platform linux --platform windows` as opposed to something like `tldr --platform linux,windows`? Or was that second alternative considered?
- The [client specification](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md) does not mention the behavior in case of multiple pages. Niklas asked this in [this comment](https://github.com/dbrgn/tealdeer/pull/300#discussion_r1008560304), but I didn't see that it was addressed. Are there any other clients that support specifying multiple platforms? How does their API look like?
- (Also, the help seems confusing to me now, because it mixes a singular keyword with a plural parameter name. But I can fix that in a separate PR.)

niklasmohrin:
- I dont remember exactly, this is just the default behavior of clap for this. I think I slightly prefer the current version, because it passes multiple platforms via multiple arguments and a parsing error could point directly to which platform is invalid. 
- I thought that the tests assert a reasonable behavior - platforms are checked in the order they are passed - but I dont fully remember and would have to dig in again. If the spec decides something else, we can change it again
- Agree, this could be improved. Maybe a note that it can be passes multiple times would be better

jj-style:
I did not check if or how other clients support multiple platforms, sorry.
I agree with  @niklasmohrin about using the default clap behaviour but happy to see if there are other ways of doing it in clap that provide a more concise interface. 

Happy to update the help text in another PR either way!

dbrgn:
Help text PR is here: #375

PR #303, opened 2022-12-11
Allow building with native-tls

dbrgn:
Right now we only support building with Rustls. However, there are quite a few architectures that aren't yet supported by "ring" (the crypto library used by Rustls), for example MIPS, PowerPC or SPARC.

To offer an alternative, I added the "native-tls" feature that can be used instead of "native-roots". When used, the native TLS stack is used instead of Rustls (i.e. SChannel on Windows, Secure Transport on macOS and OpenSSL otherwise).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,9 +46,20 @@ filetime = "0.2.10"
 
 [features]
 default = ["native-roots"]
+logging = ["env_logger"]
+
+# Reqwest (the HTTP client library) can handle TLS connections in three
+# different modes:
+#
+# - Rustls with native roots
+# - Rustls with WebPK roots
+# - Native TLS (SChannel on Windows, Secure Transport on macOS and OpenSSL otherwise)
+#
+# Exactly one of the three variants must be selected. By default, Rustls with
+# native roots is enabled.
 native-roots = ["reqwest/rustls-tls-native-roots"]
 webpki-roots = ["reqwest/rustls-tls-webpki-roots"]
-logging = ["env_logger"]
+native-tls = ["reqwest/native-tls"]
 
 [profile.release]
 lto = true
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,10 +18,16 @@
 
 #[cfg(any(
     all(feature = "native-roots", feature = "webpki-roots"),
-    not(any(feature = "native-roots", feature = "webpki-roots")),
+    all(feature = "native-roots", feature = "native-tls"),
+    all(feature = "webpki-roots", feature = "native-tls"),
+    not(any(
+        feature = "native-roots",
+        feature = "webpki-roots",
+        feature = "native-tls"
+    )),
 ))]
 compile_error!(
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -17,11 +17,22 @@
 #![allow(clippy::too_many_lines)]
 
 #[cfg(any(
+    all(
+        feature = "native-roots",
+        feature = "webpki-roots",
+        feature = "native-tls"
+    ),
```
niklasmohrin, line 24: This should already be covered by the cases below, right?

src/main.rs
```diff
@@ -17,11 +17,22 @@
...
+    all(feature = "webpki-roots", feature = "native-tls"),
+    not(any(
+        feature = "native-roots",
+        feature = "webpki-roots",
+        feature = "native-tls"
+    )),
 ))]
 compile_error!(
-    "exactly one of feature \"native-roots\" and feature \"webpki-roots\" must be enabled"
+    "exactly one of features \"native-roots\", \"webpki-roots\" or \"native-tls\" must be enabled"
```
niklasmohrin, line 35:
```suggestion
    "exactly one of the features \"native-roots\", \"webpki-roots\" or \"native-tls\" must be enabled"
```

Cargo.toml
```diff
@@ -46,8 +46,12 @@ filetime = "0.2.10"
 
 [features]
 default = ["native-roots"]
+# Use Rustls with native roots
```
niklasmohrin, line 49: For me, the comment doesn't provide any more information over the line below, I am in favor of removing it again

Cargo.toml
```diff
@@ -46,8 +46,12 @@ filetime = "0.2.10"
...
 native-roots = ["reqwest/rustls-tls-native-roots"]
+# Use Rustls with WebPK roots
 webpki-roots = ["reqwest/rustls-tls-webpki-roots"]
+# Use native TLS (SChannel on Windows, Secure Transport on macOS and OpenSSL otherwise)
```
niklasmohrin, line 53: Here I am not sure, instinctively, I would say that
a) this comment can only become outdated
b) if you are compiling for a platform that doesn't support the other options, you probably know what this means
c) if you don't know what it means and are unaware what this means, you are probably not looking into `Cargo.toml` in the first place

If we want to bring attention to these features, we could have a small section in the README or documentation about building, here I don't see much of a benefit.

src/main.rs
dbrgn: True!

Cargo.toml
dbrgn: The idea was to highlight/explain the three different options to handle TLS:

- RusTLS with WebPKI roots
- RusTLS with native roots
- Native TLS

I'm not sure if this belongs in the docs, since configuring this is rather niche, and people who want to customize the build process probably look at `Config.toml` anyways. But maybe a single intro section would be better, than a line above every option.

dbrgn:
@niklasmohrin updated! I think it's better than before 🙂 Good to merge now?

niklasmohrin (approved):

PR #305, opened 2022-12-18
Add scoop package manager as supported installation method

agrmohit:
Adds scoop as a supported installation method for tealdeer in docs.

It is present in the `main` repository of scoop which is enabled by default. So the only step needed to install `tealdeer` from scoop is `scoop install tealdeer`.


docs/src/installing.md
```diff
@@ -25,6 +25,7 @@ Tealdeer has been added to a few package managers:
 - openSUSE: [`tealdeer`](https://software.opensuse.org/package/tealdeer?search_term=tealdeer)
 - Solus: [`tealdeer`](https://packages.getsol.us/shannon/t/tealdeer/)
 - Void Linux: [`tealdeer`](https://github.com/void-linux/void-packages/tree/master/srcpkgs/tealdeer)
+- Scoop: [`tealdeer`](https://github.com/ScoopInstaller/Main/blob/master/bucket/tealdeer.json)
```
dbrgn, line 28: Thanks! Can you move this up two lines? (Entries are in alphabetical order.)

dbrgn (changes requested):

dbrgn:
Thanks!

PR #310, opened 2023-01-10
Docs: Fix typo (irregular space → regular space)

YDX-2147483647:
It was a no-break space (U+00A0, `&nbsp;`), which causes mdBook to ignore the line break and [render wrongly](https://dbrgn.github.io/tealdeer/installing.html#package-managers).

Instead, a regular space (U+0020) should be used.

## Before

- openSUSE: [tealdeer](https://software.opensuse.org/package/tealdeer?search_term=tealdeer) - Scoop: [tealdeer](https://github.com/ScoopInstaller/Main/blob/master/bucket/tealdeer.json)
- Solus: [tealdeer](https://packages.getsol.us/shannon/t/tealdeer/)

## After

- openSUSE: [tealdeer](https://software.opensuse.org/package/tealdeer?search_term=tealdeer)
- Scoop: [tealdeer](https://github.com/ScoopInstaller/Main/blob/master/bucket/tealdeer.json)
- Solus: [tealdeer](https://packages.getsol.us/shannon/t/tealdeer/)

### DIFF
--- a/docs/src/installing.md
+++ b/docs/src/installing.md
@@ -23,7 +23,7 @@ Tealdeer has been added to a few package managers:
 - NetBSD: [`sysutils/tealdeer`](https://pkgsrc.se/sysutils/tealdeer)
 - Nix: [`tealdeer`](https://nixos.org/nixos/packages.html#tealdeer)
 - openSUSE: [`tealdeer`](https://software.opensuse.org/package/tealdeer?search_term=tealdeer)
-- Scoop: [`tealdeer`](https://github.com/ScoopInstaller/Main/blob/master/bucket/tealdeer.json)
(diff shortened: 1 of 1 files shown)


YDX-2147483647:
I guess #305 introduces this by accident.

niklasmohrin:
Thanks!

You are right, the next line is treated as if the bullet point goes on because there is no distinct `- ` there. Actually, after the "Scoop", there is another weird space.

Before your PR:
image
After your PR:
image

Could you fix the second one as well?

YDX-2147483647:
😺 I didn't notice that.

niklasmohrin (approved):
Thanks!

PR #314, opened 2023-02-07
Skip line without allocation

qknogxxb:


src/line_iterator.rs
```diff
@@ -65,11 +65,42 @@ impl<R: BufRead> Iterator for LineIterator<R> {
...
+                                    }
+                                }
+                                // inform BufRead how many bytes should be consumed
+                                r.consume(amt);
+                                // no more loops are needed if the byte b'\n' has been found in the buffer
+                                if amt != buf_len {
+                                    break;
+                                }
+                            }
+                            Ok(())
```
niklasmohrin, line 101: I fell into a bit of a rabbit hole with the implementation. At first, it can be simplified by using `Iterator::position`:
```rust
loop {
    let buf = r.fill_buf()?;

    if buf.is_empty() {
        return Ok(());
    }

    if let Some(offset) = buf.iter().position(|&c| c == b'\n') {
        r.consume(offset + 1);
        return Ok(());
    }

    let buf_len = buf.len();
    r.consume(buf_len);
}
```

However, this still feels too complicated - we don't really care about buffers and filling and consuming them! We really just want to read from the reader until there is a newline. So, we look at the bytes of the reader and `find` the first newline character or error. By `find`ing up to this point, the reader is advanced exactly as far as we want. Finally, we discard whatever the result of the search was into `()`:
```rust
<&mut R as std::io::Read>::bytes(r)
    .find(|b| matches!(b, Ok(b'\n') | Err(_)))
    .transpose()
    .map(|_newline_char| ())
```
This version is far more concise, but a bit harder to understand because we manipulate `Option<Result<...>>` and need to avoid `Deref` on `r`. However, it works for any type that is `Read`, not only `BufRead` and is closer to what we actually want to do.

I like the second version more, but am fine with either of the two. @dbrgn what do you think?

src/line_iterator.rs
```diff
@@ -65,11 +65,42 @@ impl<R: BufRead> Iterator for LineIterator<R> {
                     } else {
                         // It's the new format! Drop next line.
                         // (Hmm, is there a way to do this without an allocation?)
-                        let mut devnull = String::new();
-                        if let Err(e) = self.reader.read_line(&mut devnull) {
+                        // let mut devnull = String::new();
+                        // if let Err(e) = self.reader.read_line(&mut devnull) {
```
niklasmohrin, line 69:
```suggestion
```

niklasmohrin (commented):
Thank you for contributing! I have some comments about the implementation and where in the codebase it should be placed below:

src/line_iterator.rs
```diff
@@ -65,11 +65,42 @@ impl<R: BufRead> Iterator for LineIterator<R> {
...
+                        if let Err(e) = skip_one_line(&mut self.reader) {
                             warn!("Could not read line from reader: {:?}", e);
                             return None;
                         }
+
+                        fn skip_one_line<R: BufRead>(r: &mut R) -> std::io::Result<()> {
```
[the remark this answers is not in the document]
niklasmohrin: Note that my latter comment, which suggests making this only use `Read`, would mean that we make an extension trait for `Read`, not `BufRead` :)

src/line_iterator.rs
qknogxxb: Your solution is much better! Now we can do the same thing in only 3 lines of codes, not in 30 or more. You are genius!

src/line_iterator.rs
```diff
@@ -64,9 +64,10 @@ impl<R: BufRead> Iterator for LineIterator<R> {
                         self.format = TldrFormat::V1;
                     } else {
                         // It's the new format! Drop next line.
-                        // (Hmm, is there a way to do this without an allocation?)
-                        let mut devnull = String::new();
-                        if let Err(e) = self.reader.read_line(&mut devnull) {
+                        if let Err(e) = Read::bytes(&mut self.reader)
+                            .find(|b| matches!(b, Ok(b'\n') | Err(_)))
+                            .transpose()
+                        {
```
dbrgn, line 70: That looks quite elegant (and _much_ better than the original version 😄).

@niklasmohrin in your version, you mapped the result to a `()`, but as proposed by @qknogxxb I don't think we need that at all.

Furthermore, you used `<&mut R as std::io::Read>::bytes(...)` That doesn't seem necessary, as can be seen here, right? Or was it something else you had in mind?

qknogxxb:
@dbrgn Your summary is very professional.

@niklasmohrin It's very kind of you to stroke so many times, I received your kindness and encouragement. This is my first PR in github.

src/line_iterator.rs
niklasmohrin: I was still writing with the function signature in mind, but inlining seems like a good call from @qknogxxb :)

The extra syntax was just to get around derefs, didn't notice that `Read::bytes` already picks up the correct impl. (`self.reader.by_ref().bytes()` still tries to use `BufReader::bytes`, not `(&mut BufReader)::bytes`)

niklasmohrin (approved):
Thanks for picking this up :)

If you feel like it, you could extend the test at the end of the file to make sure that the output of `LineIterator` is as expected after the skipped line. It's not a must-have though. Do you want to try it, or not? You can also change your mind later and ask for help of course :)

niklasmohrin:
Oh, and @dbrgn let's not forget to squash when this is done :)

src/line_iterator.rs
qknogxxb: It seems that the compiler is not smart enough to infer how far it should to deref and see whether is it properly.
The compiler always deref to the end of the deref chain. A hint to compiler is necessary anyway.

qknogxxb:
@niklasmohrin 

Yes, I would like to do some contributions and practices.
I notice that there is a test at the end of the file, which I believe it's enough to ensure the new code is correct.
And what is exactly thrown to your `devnull` (definately a clever name) is `"=========\n"`, the same what `<Bytes as Iterator>::find` skips.

```rust
    #[test]
    fn test_first_line_new_format() {
        let input = "The Title\n=========\n\n";
        let mut lines = LineIterator::new(input.as_bytes());
        let title = lines.next().unwrap();
        assert_eq!(title, LineType::Title("The Title".to_string()));
        let empty = lines.next().unwrap();
        assert_eq!(empty, LineType::Empty);
    }
```

dbrgn:
@qknogxxb yes, I think you're right, that test should cover the functionality.

dbrgn (approved):
Thanks @qknogxxb for your contribution! I'll merge this after a rebase & squash.

qknogxxb:
@dbrgn
And thank you too for inventing this convinent tool!
I can say `tldr` prevent me from wasting at least 20 minutes per day to look up documents.

niklasmohrin:
I still think the test could be improved ... :D

I mean, it tests that the `=` disappear, but it is not clear whether we would have a leading `\n` when continuing to read (which is removed by `trim_end` later by coincidence of the test case). Something like
```rust
let input = "\
The Title
=======
I am important!
";
```
could be used to ensure the reader doesn't forget leave the newline or reads too far. In general, I would want the test to give me confidence that we don't have off-by-ones like the following would:
```rust
let mut previous = 0;
if let Err(e) = Read::bytes(&mut self.reader)
    .find(|b| {
        let res = previous == b'\n';
        previous = *b.as_ref().unwrap();
        res
    })
    .transpose()
```
... which also looks reasonable at first glance :^)

dbrgn:
Further PRs with improvements are always welcome of course 🙂

qknogxxb:
@niklasmohrin
You mentioned a new word "off-by-ones". I guess this means a special overflow bug which exactly overflow one byte.
Specifically in your example code, the next adjacent `char` ("I" in "I am important") after `\n` will be consumed unexpectedly.
I will try to improve the test on this situation.

PR #318, opened 2023-02-20
Add config file option to set preferred language

### LINKED ISSUE #251: Add config file option to set prefered language
Since #96 it is possible to specify a language via the command line argument `-L/--language`, this works quite well. Now I would like to always use a specific language, differing from the `LANG` set as an environment variable.

I played around with the config file and had a look at the code but it seems to not be supported as a config file option. Would it be possible to add this? I'm imagening something like an additional section in the config file:

```toml
[language]
language = "en"
```

And of course: Thank's a lot for tealdeer, it really is charmingly fast and straightforward to use!

cjwcommuny:
This pull request enable users to write the following config:

```toml
[display]
language = "zh_CN"
```

to override the language.

Closes #251.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,6 +30,7 @@ compile_error!(
     "exactly one of the features \"native-roots\", \"webpki-roots\" or \"native-tls\" must be enabled"
 );
 
+use std::iter::once;
 use std::{env, process};
 
 use app_dirs::AppInfo;
@@ -202,17 +203,26 @@ fn init_log() {
 #[cfg(not(feature = "logging"))]
 fn init_log() {}
 
-fn get_languages(env_lang: Option<&str>, env_language: Option<&str>) -> Vec<String> {
+fn get_languages(
+    command_line_lang: Option<&str>,
+    config_lang: Option<&str>,
+    env_lang: Option<&str>,
+    env_language: Option<&str>,
+) -> Vec<String> {
     // Language list according to
     // https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#language
 
-    if env_lang.is_none() {
-        return vec!["en".to_string()];
-    }
-    let env_lang = env_lang.unwrap();
+    let env_language = env_language.unwrap_or("").split(':');
 
-    // Create an iterator that contains $LANGUAGE (':' separated list) followed by $LANG (single language)
-    let locales = env_language.unwrap_or("").split(':').chain([env_lang]);
+    let locales: Vec<&str> = if let Some(command_line_lang) = command_line_lang {
+        vec![command_line_lang]
+    } else if let Some(config_lang) = config_lang {
+        vec![config_lang]
+    } else if let Some(env_lang) = env_lang {
+        env_language.chain(once(env_lang)).collect()
+    } else {
+        env_language.collect()
+    };
 
     let mut lang_list = Vec::new();
     for locale in locales {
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -30,6 +30,7 @@ compile_error!(
     "exactly one of the features \"native-roots\", \"webpki-roots\" or \"native-tls\" must be enabled"
 );
 
+use std::iter::once;
 use std::{env, process};
```
dbrgn, line 34: Nit: Can you merge these imports for consistency?

```
use std::{env, iter::once, process};
```

src/main.rs
```diff
@@ -202,17 +203,26 @@ fn init_log() {
 #[cfg(not(feature = "logging"))]
 fn init_log() {}
 
-fn get_languages(env_lang: Option<&str>, env_language: Option<&str>) -> Vec<String> {
+fn get_languages(
```
dbrgn, line 206: Maybe add a comment indicating the precedence?

```rust
/// Return language(s) with the following precedence:
///
/// 1. CLI language parameter (`-L`)
/// 2. Language as set in the config file
/// 3. If the `$LANG` env var is set: `$LANGUAGE` (list separated by `:`) followed by `$LANG` (single language)
/// 4. Fallback: English
```

src/main.rs
```diff
@@ -202,17 +203,26 @@ fn init_log() {
...
-    let locales = env_language.unwrap_or("").split(':').chain([env_lang]);
+    let locales: Vec<&str> = if let Some(command_line_lang) = command_line_lang {
+        vec![command_line_lang]
+    } else if let Some(config_lang) = config_lang {
+        vec![config_lang]
+    } else if let Some(env_lang) = env_lang {
+        env_language.chain(once(env_lang)).collect()
+    } else {
+        env_language.collect()
+    };
```
dbrgn, line 225: I don't think this logic conforms to https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#language anymore, because the fast-path (`if env_lang.is_none() { return ... }`) was removed.

dbrgn (changes requested):
Thanks! I left a few comments.

src/main.rs
cjwcommuny: Added.

src/main.rs
cjwcommuny: Fixed.

src/main.rs
cjwcommuny: Merged.

src/main.rs
```diff
@@ -202,17 +202,32 @@ fn init_log() {
 #[cfg(not(feature = "logging"))]
 fn init_log() {}
 
-fn get_languages(env_lang: Option<&str>, env_language: Option<&str>) -> Vec<String> {
+/// Return language(s) with the following precedence:
+///
+/// 1. CLI language parameter (`-L`, `--language`)
+/// 2. Language as set in the config file
+/// 3. If the `$LANG` env var is set: `$LANGUAGE` (list separated by `:`) followed by `$LANG` (single language)
+/// 4. Fallback: English
```
niklasmohrin, line 210: This is not exactly what we want - the "Fallback: English" should only apply to the environment variables, not to the CLI or config file setting. Hence I suggest the following:
- Leave `get_languages` as is, maybe only rename it to `get_languages_from_env` or so and remove the current `get_languages_from_env`
- Make a new function that takes `args` and `config` and tries `args`, then `config` and finally uses `get_languages_from_env`.
- To allow for testing without setting environment variables, the new function should only redirect to another function with a signature like your current `get_languages`:
```rust
fn the_new_function_impl(
    command_line_lang: Option<&str>,
    config_lang: Option<&str>,
    env_lang: Option<&str>,
    env_language: Option<&str>,
) -> Vec<String> {
```

niklasmohrin (changes requested):

niklasmohrin:
@cjwcommuny Do you want to continue working on this, or should we close this PR for now?

PR #319, opened 2023-04-22
Use XDG conventions on macOS too

### LINKED ISSUE #311: Respect XDG_* variables on macos if set
When the user explicitly has set the `XDG_*` variables, tealdeer should use them on macOS like it does on Linux, falling back to Mac conventions if not set.

utkarshgupta137:
Fixes #311.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,18 +24,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
-name = "app_dirs2"
-version = "2.5.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "47a8d2d8dbda5fca0a522259fb88e4f55d2b10ad39f5f03adeebf85031eba501"
-dependencies = [
- "jni",
- "ndk-context",
- "winapi",
- "xdg",
-]
-
-[[package]]
 name = "assert_cmd"
 version = "2.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -115,12 +103,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a20104e2335ce8a659d6dd92a51a767a0c062599c73b343fd152cb401e828c3d"
 
 [[package]]
-name = "cesu8"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
-
-[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -166,16 +148,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "combine"
-version = "4.6.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35ed6e9d84f0b51a7f52daf1c7d71dd136fd7a3f41a8462b8cdb8c78d920fad4"
-dependencies = [
- "bytes",
- "memchr",
-]
-
-[[package]]
 name = "core-foundation"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dbrgn:
Hi, thanks for the PR. Some questions before I take a closer look:

- In the base configuration (if no XDG variables are set), will this use the same directories as before on non-macOS systems? Have you verified that this does not modify the directory paths on Windows and Linux?
- In [#311](https://github.com/dbrgn/tealdeer/issues/311), the request was to use XDG conventions _if_ XDG variables are set, but not by default. This makes sense, as we should follow the local OS conventions by default. Does etcetera implement this logic? From what I can tell, it will always use XDG conventions. This would be a breaking change for Tealdeer users on macOS (not just the cache will be created a second time in a different location without the old cache being cleaned up, but if a config file is present in the previous location, it will be ignored).

utkarshgupta137:
> Hi, thanks for the PR. Some questions before I take a closer look:
> 
>     * In the base configuration (if no XDG variables are set), will this use the same directories as before on non-macOS systems? Have you verified that this does not modify the directory paths on Windows and Linux?
> 
>     * In [#311](https://github.com/dbrgn/tealdeer/issues/311), the request was to use XDG conventions _if_ XDG variables are set, but not by default. This makes sense, as we should follow the local OS conventions by default. Does etcetera implement this logic? From what I can tell, it will always use XDG conventions. This would be a breaking change for Tealdeer users on macOS (not just the cache will be created a second time in a different location without the old cache being cleaned up, but if a config file is present in the previous location, it will be ignored).

Ugh, looks like app_dirs2 doesn't follow the convention on Windows of most other "dirs" crates out there. Linux users won't be affected, but I can modify the code to retain the same behavior on Windows. I don't use CLI tools on Windows, so I've no strong opinions about it.

No, right now this PR makes no effort to maintain backward compatibility on macOS. IMO, no macOS users expect the config file to be located in `~/Library/Application Support`. Almost every CLI tool which uses the `dirs`/`dirs-next` crates & other clones has open or closed issues of users wanting to use the XDG conventions on macOS. But I've yet to come across a case where the inverse is true. I've written more about this [here](https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-1464129466).
Secondly, I don't think it is a good idea to half-ass the XDG convention. If a user who has previously not set it decides to set it for whatever reason, then they may be confused why some unrelated tool is broken.
However, since you probably want to maintain backward compatibility, I can add migration code to move the old data to the new location, similar to this: https://github.com/cantino/mcfly/pull/349/files#diff-4f91d834bba4a3a432715f3ccf4ebc7c10d7ecfb974f1e78f56108dd80581a53R392-R403

utkarshgupta137:
Fixed & test on Windows:
```
PS C:\Users\Utkarsh Gupta\Downloads\tealdeer-main> ..\tealdeer-windows-x86_64-msvc.exe --show-paths
Config dir:       C:\Users\Utkarsh Gupta\AppData\Roaming\tealdeer\tealdeer\ (OS convention)
Config path:      C:\Users\Utkarsh Gupta\AppData\Roaming\tealdeer\tealdeer\config.toml
Cache dir:        C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer (OS convention)
Pages dir:        C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer\tldr-pages\
Custom pages dir: C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer\pages\ (OS convention)

PS C:\Users\Utkarsh Gupta\Downloads\tealdeer-main> .\target\release\tldr.exe --show-paths
Config dir:       C:\Users\Utkarsh Gupta\AppData\Roaming\tealdeer\tealdeer\ (Windows convention)
Config path:      C:\Users\Utkarsh Gupta\AppData\Roaming\tealdeer\tealdeer\config.toml
Cache dir:        C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer (Windows convention)
Pages dir:        C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer\tldr-pages\
Custom pages dir: C:\Users\Utkarsh Gupta\AppData\Local\tealdeer\tealdeer\pages (Windows convention)
```

utkarshgupta137:
I should mention that is weird that app_dirs2 uses the same location for cache & user data. The popular [`directories`](https://github.com/dirs-dev/directories-rs#projectdirs) and most of its clones put the data & config together, not the data & cache.

utkarshgupta137:
@dbrgn If you want, I can add a fallback to `~/Library/Application Support` as well.

utkarshgupta137:
@dbrgn ping

niklasmohrin:
Let's discuss the transition period in the issue thread first :)

dbrgn:
@utkarshgupta137 thanks for your work so far. Do you plan to carry on with this PR? If yes, it would need to be updated as per https://github.com/dbrgn/tealdeer/issues/311#issuecomment-1937787422

utkarshgupta137:
Sorry, I don't use this tool very much & generally don't have enough bandwidth to continue. But I would be happy to review PRs or assist otherwise.

PR #322, opened 2023-05-02
Change custom page files to use a `.md` extension.

### LINKED ISSUE #320: Custom Pages With `.md` Extension
I really like the custom pages functionality, but it's not clear to me why the custom pages use a `.page` extension when the stock pages use `.md`.

Is the reason simply that the patch files use `.patch`?

The reason I'm interested in this is that I'd like to store my custom pages in an [Obsidian](https://obsidian.md/) vault as notes. This way, I can maintain command usage information in my regular notes app and have it accessible from the command line with tealdeer. Unfortunately, Obsidian expects the files to have a `.md` extension, which complicates things.

It would be fantastic to be able to define a custom extension or use `.md` for the custom pages.
I'm not sure about how the patch files would have to be handled, however. For me, I don't see myself using patches, but it would be great to come to a solution that works for them too.

I'm happy to work on a PR for this, but before doing anything, I want to make sure I understand the decisions that were made, and what the best design would be.

zedseven:
My rationale for this PR is in #320. Closes #320.

To summarise, the lack of a `.md` extension causes problems for external tools working with the custom page files, and my solution is to simply add the extension to the end.

e.g.
- `{name}.page.md`
- `{name}.patch.md`

This *will* be a breaking change for anyone using custom pages.


niklasmohrin:
As I outlined in the issue thread, let's first find out how we want to handle this breaking change if we want it. Happy to continue discussion here once we are all on the same page :)

dbrgn:
As commented in https://github.com/dbrgn/tealdeer/issues/320#issuecomment-1937775105, I think it's valuable to print a warning when pages with `.page` or `.patch` extension are found in the custom pages directory. This will be a small performance hit, but we can remove it again after 1-2 releases.

@zedseven would you be willing to add such a check?

dbrgn:
Could you add a breaking change warning to the docs?

```diff
diff --git a/docs/src/usage_custom_pages.md b/docs/src/usage_custom_pages.md
index 7024abd..d5d0f89 100644
--- a/docs/src/usage_custom_pages.md
+++ b/docs/src/usage_custom_pages.md
@@ -1,5 +1,13 @@
 # Custom Pages and Patches

+> ⚠️ **Breaking change in version 1.7.0:** The file name extension for custom
+> pages and patches was changed:
+>
+> - `<name>.page` → `<name>.page.md`
+> - `<name>.patch` → `<name>.patch.md`
+>
+> If you have custom pages or patches, you need to rename them.
+
 Tealdeer allows creating new custom pages, overriding existing pages as well as
 extending existing pages.
```

zedseven:
I've added the temporary check (when looking up any command), added your warning to the docs, rebased the branch onto the latest `main`, and fixed a bug I noticed with my initial implementation.

src/cache.rs
```diff
@@ -407,6 +409,32 @@ impl Cache {
...
+                    let extension = entry.path().extension();
+                    if let Some(extension) = extension {
+                        extension == "page" || extension == "patch"
+                    } else {
+                        false
+                    }
+                } else {
+                    false
+                }
+            });
```
dbrgn, line 435: This check could have been written without the `WalkDir` crate, just using `fs::read_dir`. But I started rewriting it, and noticed that the `WalkDir` API is clearly nicer, so let's stick to it 🙂

src/cache.rs
```diff
@@ -395,6 +413,39 @@ impl Cache {
...
+        if old_custom_pages_exist {
+            print_warning(
```
dbrgn, line 437: I changed this code to use `print_warning`.

dbrgn (approved):
Great, thanks! I pushed some minor tweaks.

PR #324, opened 2023-05-10
Update CI workflows

kbdharun:
## Changes

- Remove unmaintained rust actions and use the components directly in the [GitHub action runners](https://github.com/actions/runner-images).
- Update runners.
- Update toolchain to https://github.com/dtolnay/rust-toolchain.
- Bump the `actions/checkout` version to `v3` (Node 16).
- Bump the `actions/upload-artifact` version to `v3` (Node 16).
- Bump the `actions/download-artifact` version to `v3` (Node 16).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,7 @@ jobs:
        - uses: dtolnay/rust-toolchain@stable
          with:
            toolchain: stable
+           components: clippy
        - name: run clippy lints
          run: cargo clippy --features logging
 
@@ -45,6 +46,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
+          components: rustfmt
       - name: run rustfmt
         run: cargo fmt --all -- --check
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,14 @@ jobs:
     name: run tests
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
-        rust: ['1.62', stable]
+        platform: [ubuntu-latest, macos-latest, windows-latest, ubuntu-20.04]
+        toolchain: [stable, 1.62.1]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ matrix.toolchain }}
       - name: Build with default features
         run: cargo build
       - name: Build with logging and webpki roots
@@ -28,12 +31,20 @@ jobs:
   clippy:
     runs-on: ubuntu-latest
     steps:
+       - uses: actions/checkout@v3
+       - uses: dtolnay/rust-toolchain@stable
+         with:
+           toolchain: stable
        - name: run clippy lints
          run: cargo clippy --features logging
 
   fmt:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
       - name: run rustfmt
         run: cargo fmt --all -- --check
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,12 +13,12 @@ jobs:
     name: run tests
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest, ubuntu-20.04]
+        platform: [ubuntu-latest, macos-latest, windows-latest]
         toolchain: [stable, 1.62.1]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.toolchain }}
       - name: Build with default features
@@ -29,10 +29,11 @@ jobs:
         run: cargo test
 
   clippy:
+    name: run clippy lints
     runs-on: ubuntu-latest
     steps:
        - uses: actions/checkout@v3
-       - uses: dtolnay/rust-toolchain@stable
+       - uses: dtolnay/rust-toolchain@master
          with:
            toolchain: stable
            components: clippy
@@ -40,10 +41,11 @@ jobs:
          run: cargo clippy --features logging
 
   fmt:
+    name: run rustfmt
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
           components: rustfmt
@@ -60,6 +62,9 @@ jobs:
         with:
           mdbook-version: '0.4.4'
       - name: Build
+        uses: dtolnay/rust-toolchain@master
+        with:
+            toolchain: stable
         run: cargo build
       - name: Ensure that docs can be built
         run: cd docs && mdbook build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
+        rust: ['1.62', stable]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,10 +61,11 @@ jobs:
         uses: peaceiris/actions-mdbook@v1
         with:
           mdbook-version: '0.4.4'
-      - name: Build
+      - name: Setup toolchain
         uses: dtolnay/rust-toolchain@master
         with:
             toolchain: stable
+      - name: Build
         run: cargo build
       - name: Ensure that docs can be built
         run: cd docs && mdbook build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,79 +14,39 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        rust: ['1.62', stable]
     runs-on: ${{ matrix.platform }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          profile: minimal
-          default: true
+      - uses: actions/checkout@v3
       - name: Build with default features
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
+        run: cargo build
       - name: Build with logging and webpki roots
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --features logging,webpki-roots --no-default-features
+        run: cargo build --features logging,webpki-roots --no-default-features
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
+        run: cargo test
 
   clippy:
-    name: run clippy lints
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          default: true
-          components: clippy
-      - uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --features logging
+       - name: run clippy lints
+         run: cargo clippy --features logging
 
   fmt:
-    name: run rustfmt
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        profile: minimal
-        default: true
-        components: rustfmt
-    - uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
+      - name: run rustfmt
+        run: fmt --all -- --check
 
   docs:
     name: build docs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
           mdbook-version: '0.4.4'
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          default: true
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
+        run: cargo build
       - name: Ensure that docs can be built
         run: cd docs && mdbook build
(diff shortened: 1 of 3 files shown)


kbdharun:
> Hi :) These changes seem nice, but they make it a bit intransparent as to what versions are used. For example, I don't see which version of the rust compiler is running (and we also want to build with two specific versions: the msrv and stable). 

Sorry for the delay, I wasn't feeling well. I have updated the workflow now to support multiple versions of Rust, a test action run can be found [here](https://github.com/kbdharun/tealdeer/actions/runs/4959978670) [Note: the failed actions are because I am running it in a fork, it should run fine here, also by the way can you approve the workflow run]

> About the OS update: not sure either, but it should probably be fine

Yep, it is better to use the latest stable versions to have faster CIs and additional features. Let me know If you guys need me to revert it.

niklasmohrin:
Hi, I think it is still not correct: You have added the line back in that starts more CI runs with different values for the `rust` variable, but this variable is never used (note the use of `${{ matrix.rust }}` before when installing the toolchain). Currently, it is still the preinstalled version of Rust that is used, which is not what we want for the reasons I wrote before.

I was unaware that actions-rs is not maintained anymore, so switching to something else seems like a good idea. From what I can tell from https://github.com/actions-rs/toolchain/issues/216 and for example https://github.com/jonhoo/rust-ci-conf/commit/362696ab8007ef1a4779885a398286856cacf555 the world is moving to dtolnay's action, so I think going with that would be a sane move :)

> Note: the failed actions are because I am running it in a fork

Actually, I think that you forgot to add the `checkout` action back in in the two jobs that are failing as both fail because of a missing `Cargo.toml` file :)

About using latest os, I was wondering if we would maybe want to stay with the older versions so that we don't miss an incompatibility with for example windows 10, ubuntu 22.04 or so. @dbrgn what do you think about this?

kbdharun:
> Hi, I think it is still not correct: You have added the line back in that starts more CI runs with different values for the `rust` variable, but this variable is never used (note the use of `${{ matrix.rust }}` before when installing the toolchain). Currently, it is still the preinstalled version of Rust that is used, which is not what we want for the reasons I wrote before.

Noted

> I was unaware that actions-rs is not maintained anymore, so switching to something else seems like a good idea. From what I can tell from [actions-rs/toolchain#216](https://github.com/actions-rs/toolchain/issues/216) and for example [jonhoo/rust-ci-conf@362696a](https://github.com/jonhoo/rust-ci-conf/commit/362696ab8007ef1a4779885a398286856cacf555) the world is moving to dtolnay's action, so I think going with that would be a sane move :)

Agreed, I will try to update this PR to https://github.com/dtolnay/rust-toolchain as even with matrix jobs it is possible for some old versions of packages to be dropped from runner images i.e https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md.

> > Note: the failed actions are because I am running it in a fork
> 
> Actually, I think that you forgot to add the `checkout` action back in in the two jobs that are failing as both fail because of a missing `Cargo.toml` file :)

Oops, just now noticed it will fix it.

> About using latest os, I was wondering if we would maybe want to stay with the older versions so that we don't miss an incompatibility with for example windows 10, ubuntu 22.04 or so. @dbrgn what do you think about this?

Regarding this, the [official GitHub runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) doesn't support Windows 10 runners while the `ubuntu-latest` runner is the latest LTS Ubuntu version which is 22.04 now. Do note if you need I could add Ubuntu 20.04 to tests but the runner will get removed next year when 24.04 gets released as GitHub only keeps 2 LTS versions of Ubuntu runners at a time. An alternative is running the job in a container (as [GitHub self-hosted runners aren't advisable for public projects](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners))

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
     name: run tests
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
```
niklasmohrin, line 16: Hm, running two versions of Ubuntu seems overkill. I've thought about it some more and think that your initial suggestion is correct and we should use `-latest` everywhere.

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-        rust: ['1.62', stable]
+        platform: [ubuntu-latest, macos-latest, windows-latest, ubuntu-20.04]
+        toolchain: [stable, 1.62.1]
     runs-on: ${{ matrix.platform }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
```
niklasmohrin, line 21: I think the intended way is to use `@master` when specifying `toolchain:` below

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-          args: --features logging,webpki-roots --no-default-features
+        run: cargo build --features logging,webpki-roots --no-default-features
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
+        run: cargo test
 
   clippy:
-    name: run clippy lints
```
niklasmohrin, line 41: Let's keep this line :)

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-          default: true
-          components: clippy
-      - uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --features logging
+       - uses: actions/checkout@v3
+       - uses: dtolnay/rust-toolchain@stable
+         with:
+           toolchain: stable
```
niklasmohrin, line 37: Since we already specified `@stable` above:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
+           components: clippy
+       - name: run clippy lints
+         run: cargo clippy --features logging
 
   fmt:
-    name: run rustfmt
```
niklasmohrin, line 57: let's also keep this

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-        default: true
-        components: rustfmt
-    - uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
```
niklasmohrin, line 48:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
+      - uses: actions/checkout@v3
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
           mdbook-version: '0.4.4'
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          default: true
```
niklasmohrin, line 85: There is no corresponding use of `dtolnay/rust-toolchain` for this use of `actions-rs/toolchain`

.github/workflows/release.yml
```diff
@@ -2,12 +2,13 @@ on:
   push:
     tags:
       - "v*" # push events to matching v*, i.e. v1.0, v20.15.10
+  workflow_dispatch:
```
niklasmohrin, line 5: What is this needed for?

.github/workflows/release.yml
```diff
@@ -63,42 +64,36 @@ jobs:
...
           path: "target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}/release/tldr"
 
   build-macos:
-    runs-on: macos-11
+    runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Build
-        uses: actions-rs/cargo@v1
```
niklasmohrin, line 83: Also no `dtolnay/rust-toolchain` before this usage of `cargo build`

.github/workflows/release.yml
```diff
@@ -63,42 +64,36 @@ jobs:
...
-        with:
-          command: build
-          args: --release --target x86_64-pc-windows-msvc
-      - uses: actions/upload-artifact@v2
+        run: cargo build --release --target x86_64-pc-windows-msvc
```
niklasmohrin, line 95: and here too

niklasmohrin (changes requested):
Alright, some more comments before we can merge this :)

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
     name: run tests
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
-        rust: ['1.62', stable]
+        platform: [ubuntu-latest, macos-latest, windows-latest, ubuntu-20.04]
```
kbdharun, line 16:
```suggestion
        platform: [ubuntu-latest, macos-latest, windows-latest]
```

.github/workflows/ci.yml
kbdharun:
```suggestion
      - uses: dtolnay/rust-toolchain@master
```
Will update it to `@master` above. Regarding the toolchain below, the detail is specified in the top as a matrix job instead of specifying it multiple times in the file.

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-          toolchain: stable
-          profile: minimal
-          default: true
-          components: clippy
-      - uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --features logging
+       - uses: actions/checkout@v3
+       - uses: dtolnay/rust-toolchain@stable
```
kbdharun, line 35:
```suggestion
       - uses: dtolnay/rust-toolchain@master
```

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
+        run: cargo build --features logging,webpki-roots --no-default-features
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
+        run: cargo test
 
   clippy:
-    name: run clippy lints
     runs-on: ubuntu-latest
```
kbdharun, line 32:
```suggestion
    name: run clippy lints
    runs-on: ubuntu-latest
```

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
+       - uses: dtolnay/rust-toolchain@stable
+         with:
+           toolchain: stable
+           components: clippy
+       - name: run clippy lints
+         run: cargo clippy --features logging
 
   fmt:
-    name: run rustfmt
     runs-on: ubuntu-latest
```
kbdharun, line 43:
```suggestion
    name: run rustfmt
    runs-on: ubuntu-latest
```

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-        toolchain: stable
-        profile: minimal
-        default: true
-        components: rustfmt
-    - uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
```
kbdharun, line 46:
```suggestion
      - uses: dtolnay/rust-toolchain@master
```

.github/workflows/release.yml
kbdharun: Workflow dispatch allows collaborators (like you) to manually run the workflow. Now that I go through again it is already set to run on tags. So this won't be necessary, so I will remove it.

```suggestion
```

.github/workflows/release.yml
```diff
@@ -63,42 +64,36 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
-      - uses: actions/upload-artifact@v2
+        run: cargo build --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
```
kbdharun, line 84:
```suggestion
           uses: dtolnay/rust-toolchain@master
           with:
              toolchain: stable
           run: cargo build --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
```

.github/workflows/release.yml
kbdharun:
```suggestion
       uses: dtolnay/rust-toolchain@master
        with:
            toolchain: stable
        run: cargo build --release --target x86_64-pc-windows-msvc
```

.github/workflows/ci.yml
kbdharun: I think the general one is fine in this case as it doesn't impact the time taken by the CI.

.github/workflows/ci.yml
```diff
@@ -13,80 +13,54 @@ jobs:
...
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          default: true
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
+        run: cargo build
```
kbdharun, line 63:
```suggestion
        uses: dtolnay/rust-toolchain@master
        with:
            toolchain: stable
        run: cargo build
```

kbdharun (commented):
GitHub actions runners have `cargo` and `rustfmt` builtin. As per your request, I will add the toolchain action for those use cases too.

.github/workflows/release.yml
```diff
@@ -63,42 +63,42 @@ jobs:
...
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
-      - uses: actions/upload-artifact@v2
+           uses: dtolnay/rust-toolchain@master
+           with:
+              toolchain: stable
+           run: cargo build --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
```
kbdharun, line 86:
```suggestion
              command: cargo build --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
```

.github/workflows/release.yml
```diff
@@ -63,42 +63,42 @@ jobs:
...
+      - uses: actions/checkout@v3
       - name: Build
-        uses: actions-rs/cargo@v1
+       uses: dtolnay/rust-toolchain@master
         with:
-          command: build
-          args: --release --target x86_64-pc-windows-msvc
-      - uses: actions/upload-artifact@v2
+            toolchain: stable
+        run: cargo build --release --target x86_64-pc-windows-msvc
```
kbdharun, line 100:
```suggestion
           command: cargo build --release --target x86_64-pc-windows-msvc
```

kbdharun:
Done, I think it is GTG now.

dbrgn:
This looks good to me, let's see if CI runs through! If yes, I think we can merge this.

Thanks for your work and your patience, @kbdharun and @niklasmohrin 🙂

PR #327, opened 2023-06-05
added checks for wrong `tldr --list` output in completions

iliya-malecki:
when tldr-pages doesnt exist (yet, as in, new installation), `tldr --list` will output something that makes no sense for the completions. I am not qualified to do a real fix, but the completions output did bug me enough to patch it for zsh and bash. I tried to patch it for fish as well but couldnt get the original completion to work so im leaving the problem for actual fish users. 

TL;DR:
Purpose: reduce annoyance during first use, which is important for user-friendliness
Method: patch completions with a simple if-no-error check, for zsh and bash.

### DIFF
--- a/completion/bash_tealdeer
+++ b/completion/bash_tealdeer
@@ -27,8 +27,9 @@ _tealdeer()
 		COMPREPLY=( $( compgen -W '$( _parse_help "$1" )' -- "$cur" ) )
 		return
 	fi
-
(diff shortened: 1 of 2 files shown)


niklasmohrin (approved):
As fish is addressed in #331, I am fine with merging this as is

dbrgn (approved):
As with Fish, I don't know much about completion files, but I trust your judgement (and otherwise we can still fix it, if there are edge cases that aren't yet covered) 🙂

PR #331, opened 2023-06-22
Chore Guard Fish Completion When Cache Is Empty

adamazing:
See: #327 

I had fish installed and so implemented @niklasmohrin's suggestion.
Tested locally with:
 - empty cache (it does the right thing and doesn't output the error text)
 - `update`d cache (it does the right thing and still auto completes)


niklasmohrin (approved):
Thanks!

dbrgn (approved):
I don't know much about Fish completions, but this doesn't look wrong! Thanks.

PR #333, opened 2023-07-10
Include `completion/*` in published crate

### LINKED ISSUE #332: Include shell completion in crates.io releases?
Hi, I packaged and maintain tealdeer [in Debian](https://tracker.debian.org/pkg/rust-tealdeer). Our current packaging approach, in the Debian Rust Team, uses crates.io release tarballs. It unfortunately doesn't include the completions, thus I have to use [a patch to add them back](https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/tealdeer/debian/patches/provide-completion.patch).

Is it feasible to include them? Otherwise a manual copy may become out of sync.

niklasmohrin:
Closes #332

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 repository = "https://github.com/dbrgn/tealdeer/"
 documentation = "https://dbrgn.github.io/tealdeer/"
 version = "1.6.1"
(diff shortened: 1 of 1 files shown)


bnoctis:
Thanks ;)

Issue #335, opened 2023-07-16
Add a language option when updating the cache


9mdv:
Updating the cache downloads pages in all locales which is not needed. Being able to specify a language would be ideal, like this:

```bash
tldr --update en
```

So we would only need to fetch 15 MB (for english) instead of all pages which is over 50 MB, saving disk space as well as bandwidth for those with metered connections. This would also go well as a config option as discussed in #251.

![2023-07-14_103116](https://github.com/dbrgn/tealdeer/assets/12945959/0ba0ba19-7283-4aef-9e72-0b58a26bb4ad)

dbrgn:
We currently download the archive from tldr.sh (https://tldr.sh/assets/tldr.zip) which is generated through GitHub AFAIK. This archive contains all languages and cannot be filtered.

As long as the TLDR pages project doesn't provide separate URLs/archives per language for downloading, I doubt that this can be implemented without a bigger effort. So I'll close this feature request for now as "not planned". If you think this is worth pursueing, you could open an issue at the TLDR project itself, and ask them to provide automatically generated download paths per language.

adamazing:
FYI Have started a discussion upstream about this issue. I've created a preliminary upstream PR that would enable this but it's rebuilding all zips on every push. I'll make a subsequent PR to zip only what's necessary. Not sure if the former will be merged or closed in deference to the latter.

kbdharun:
I have merged the PR upstream and the language assets are available at <https://github.com/tldr-pages/tldr-pages.github.io/tree/main/assets>; so IG this issue can be reopened, I will update the client specification to highlight this change in https://github.com/tldr-pages/tldr/pull/10148

adamazing:
> I have merged the PR upstream and the language assets are available at https://github.com/tldr-pages/tldr-pages.github.io/tree/main/assets; so IG this issue can be reopened, I will update the client specification to highlight this addition in [tldr-pages/tldr#10148](https://github.com/tldr-pages/tldr/pull/10148).

Thanks @kbdharun! 

@dbrgn, I'll focus on the second-phase PR upstream in tldr-pages (to avoid building/rebuilding assets unless necessary) but I'm happy to look at adding the language option in tealdeer once that's done?

niklasmohrin:
Sounds good, it isn't everyday that issues here actually make it into the spec so quickly. I am happy to review and accept a PR for this (and I would be surprised if @dbrgn disagrees)

niklasmohrin:
[...]

PR #337, opened 2023-08-27
Add configuration option for `archive_source`

mipedja:
This allows for specifying a custom archive URL
as an alternative approach to Custom Pages.


niklasmohrin:
Thank you for submitting this PR! I can see from your example code that the motivation for this change seems to be to host your own internal pages - cool idea, I agree that it could be useful.

We need to think about #335 here though. A new version of the client spec is apparently about to be released (see https://github.com/tldr-pages/tldr/pull/10148) which will specify where clients (including tealdeer) will look for language specific archives. So I think we should wait until that is all cleared up and then take note in our documentation accordingly. I will hence mark this as draft for now :)

docs/src/config_updates.md
```diff
@@ -24,3 +24,13 @@ is set to `false`.
     auto_update = true
     auto_update_interval_hours = 24
 
+### archive_url
+
+URL for the location of tldr pages archive. Default is the main `thdr.sh`
+archive location.
```
niklasmohrin, line 30: The new client specification has changed the recommended source for pages again, instead of using `tldr.sh`, the Github releases are preferred now. I would thus change this to
```suggestion
URL for the location of tldr pages archive. By default the pages are fetched from the latest `tldr-pages/tldr` GitHub release.
```

docs/src/config_updates.md
```diff
@@ -24,3 +24,13 @@ is set to `false`.
...
+
+    [updates]
+    auto_update = true
+    auto_update_interval_hours = 24
```
niklasmohrin, line 34: I think we should omit these here, because they are independent of the `archive_url` option
```suggestion
```

docs/src/config_updates.md
```diff
@@ -24,3 +24,13 @@ is set to `false`.
...
+    archive_url = https://tldr.infra.local/assets/tldr.zip
```
niklasmohrin, line 35: Let's go with something more neutral like
```suggestion
    archive_url = https://my-company.example.com/tldr.zip
```

src/config.rs
```diff
@@ -16,6 +16,7 @@ use crate::types::PathSource;
 pub const CONFIG_FILE_NAME: &str = "config.toml";
 pub const MAX_CACHE_AGE: Duration = Duration::from_secs(2_592_000); // 30 days
 const DEFAULT_UPDATE_INTERVAL_HOURS: u64 = MAX_CACHE_AGE.as_secs() / 3600; // 30 days
+const ARCHIVE_URL: &str = "https://tldr.sh/assets/tldr.zip";
```
niklasmohrin, line 19: Since this is now only used in `default_archive_url`, I think we can just inline it there (and also update it to the new recommended URL, which is https://github.com/tldr-pages/tldr/releases/latest/download/tldr.zip)

niklasmohrin (commented):
Hey @mipedja,

we have been letting this PR sit for a while now, but I think now is a good time to merge this!

I made some small suggestions, if you want we can discuss them. If you (understandably so) don't want to pick up this PR right now, I will add the final updates myself in a couple of days or so :)

As far as the new client spec discussion goes: The main reason for waiting (language specific archives) has been included in the specification. Since we want to support this, we should change our config option here to not point to `some.site/tldr/pages.zip` but instead just to `some.site/tldr/`, and then require that the zip files are present as `pages.zip` / `pages.en.zip` and so on. You don't have to add much documentation for this now, as the exact wording has to be changed anyways once we switch to downloading language specific archives :)

docs/src/config_updates.md
```diff
@@ -24,3 +24,13 @@ is set to `false`.
     auto_update = true
     auto_update_interval_hours = 24
 
+### archive_url
```
niklasmohrin, line 27: Given that we will soon switch to using a URL where we append `pages.en.zip`, `pages.de.zip` etc. ourselves, maybe we should prefer something more neutral as the name. How do you feel about "archive_source"?

mipedja:
Thanks for the review. I agree with all your comments.

I started switching from `archive_url` to `archive_source` and then realized you have #345 with `archives_url`.
Should I just make this to be `archives_url` instead and then it will fit with #345 well?

For the time being I would hard-code update() to always append `tldr.zip`, and #345 would make the appropriate language pattern changes.

docs/src/config_updates.md
```diff
@@ -24,3 +24,11 @@ is set to `false`.
     auto_update = true
     auto_update_interval_hours = 24
 
+### archive_source
+
+URL for the location of tldr pages archive. By default is the pages are
+fetched from the latest `tldr-pages/thdr` GitHub release.
```
niklasmohrin, line 30:
```suggestion
URL for the location of the tldr pages archive. By default the pages are
fetched from the latest `tldr-pages/tldr` GitHub release.
```

niklasmohrin (commented):
One more thing I spotted :)

src/cache.rs
```diff
@@ -163,11 +163,13 @@ impl Cache {
     }
 
     /// Update the pages cache from the specified URL.
-    pub fn update(&self, archive_url: &str) -> Result<()> {
+    pub fn update(&self, archive_source: &str) -> Result<()> {
         self.ensure_cache_dir_exists()?;
 
+        let archive_url = format!("{}{}", archive_source, "tldr.zip");
```
niklasmohrin, line 169: I think we can be more cautious here and add the `/`. If is not present in `archive_source`, the `archive_url` will be unexpected, and if it was already present, having `//` does not change the behavior 

```suggestion
        let archive_url = format!("{}/tldr.zip", archive_source);
```

docs/src/config_updates.md
```diff
@@ -24,3 +24,11 @@ is set to `false`.
     auto_update = true
     auto_update_interval_hours = 24
 
+### archive_source
+
+URL for the location of tldr pages archive. By default the pages are
```
niklasmohrin, line 29:
```suggestion
URL for the location of the tldr pages archive. By default the pages are
```

niklasmohrin (approved):
Very cool, thanks!

Issue #340, opened 2023-09-15
Add a random flag like node client


MustCodeAl:
it would be nice to see a random tldr every time terminal is opened to help with command line learning and configurable number of commands to show for that tldr page.

`tldr --random` 

```
  page: tar

  Archiving utility.
  Often combined with a compression method, such as gzip or bzip2.
  More information: https://www.gnu.org/software/tar.


  - [c]reate an archive and write it to a [f]ile:
    tar cf path/to/target.tar path/to/file1 path/to/file2 ...

  - [c]reate a g[z]ipped archive and write it to a [f]ile:
    tar czf path/to/target.tar.gz path/to/file1 path/to/file2 ...
```

MustCodeAl:
now i have added [changes](https://github.com/dbrgn/tealdeer/compare/main...MustCodeAl:tealdeer:main)

niklasmohrin:
I am not sure if I like this being its own flag of tealdeer, when the same behavior can already be achieved in one shell command, for example in bash: `tldr -q $(tldr -q --list | shuf -n1)`. We could however write this down as an example somewhere in the documentation as a nice idea. What do you think?

MustCodeAl:
i had no idea of that command existing 😅, that would be cool

dbrgn:
I agree, I don't think a "random" command belongs in tealdeer, since this can be achieved with external tools.

I think the use case is a bit obscure, but it wouldn't hurt to add a "Tips & Tricks" section to the docs.

PR #341, opened 2023-09-16
Add "random" command to fetch random tldr page

MustCodeAl:
Added a new `--random` command that allows users to get a random page from tldr(`tldr --random`). Logic for picking a random page were implemented in the `Cache` class. The `rand` crate was also included to handle random selection of pages. This new feature gives users a way to discover new commands they may not be aware of.


niklasmohrin:
See #340

PR #346, opened 2023-10-21
Update Github actions badge

niklasmohrin:
The badge currently says "failing", although the latest commit on main passed CI. I changed the image to what Github generates if you click on the "Create status badge" button and the URL is slightly different, so maybe this is the reason?


dbrgn:
Thx :)

PR #352, opened 2024-02-04
Bump MSRV to 1.75

dbrgn:
For clap v4 (#298), we need a higher MSRV, currently at least Rust 1.64.

Our MSRV policy is:

> When publishing a tealdeer release, the Rust version required to build it should be stable for at least a month.

I'd suggest to jump to 1.75. It's the current stable, released on December 28, and contains the "async fn in traits" change which makes it a potential synchronization point for libraries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable, 1.62.1]
+        toolchain: [stable, 1.64.0]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


niklasmohrin:
I think we should stay at 1.64 until we need a higher version. Are there any benefits to upgrading beyond that?

I am just thinking about something like the Debian package which maybe still has an old Rust version

dbrgn:
> Are there any benefits to upgrading beyond that?

As long as we can bump the required Rust version with any new release, probably not. (Debian would be free to use an older version though, as long as it works.)

PR updated to 1.64.

niklasmohrin (approved):
Yes, I think we can bump msrv freely (within reason and while adhering to our policy) between releases in PRs that need it.

If we would want to be strict about it, we could make semver major releases every time, but I feel that for us as a binary crate it is not as important as for libraries - maybe thats foolish though

Issue #349, opened 2023-11-30
Support Client Specification 2.1 and 2.2


dbrgn:
See https://github.com/tldr-pages/tldr/releases/tag/v2.1

kbdharun:
Hi, we have released Client Specification 2.2 a few days ago (announcing a major change with cache URLs). Feel free to check it out at https://github.com/tldr-pages/tldr/releases/tag/v2.2. Maybe, this issue's title and scope can be extended for the newer client specification too.

niklasmohrin:
I made several issues tagged with the "spec" label, which supersede this issue.

Issue #323, opened 2023-05-10
Allow passing `common` as platform


beatbrot:
In some cases, it makes sense to explicitly request tldr docs for the common platform.

When running `tldr curl` on Windows, for example, I only get an explanation that I should pick the common platform if I were interested in the unix tool curl:

![grafik](https://github.com/dbrgn/tealdeer/assets/1830332/087ec694-b7a4-4608-9a7b-8433a5ce6e86)

Unfortunately, if I try using the suggested command, I get the following error:

```
> tldr curl -p common
error: "common" isn't a valid value for '--platform <PLATFORM>'
        [possible values: linux, macos, windows, sunos, osx, android]

For more information try --help
```

In my opinion, common should be an option here.

RuiNtD:
Bump. The option to use `common` as the platform is also mentioned as a supported option [within the `tldr` page itself](https://github.com/tldr-pages/tldr/blob/30d1a8f0fc5df91f1bb8637e329a2fdd4611de4e/pages/common/tldr.md?plain=1#L21) and is supported by the official Python and Rust clients, but currently broken in the Node client (https://github.com/tldr-pages/tldr-node-client/issues/484).

PR #354, opened 2024-02-20
feat: add BSD platform support; chore: update workflows

kbdharun:
## Changes

- This PR adds support for FreeBSD, NetBSD and OpenBSD platforms that we recently added upstream (I announced it as part of the 2.1 client specification release last year).
- ~This PR also adds support for querying the common platform via a dedicated flag for #323.~ (This didn't work during testing so will revert it)
- Bump `actions` from `v3` (Node 16) to `v4` (Node 20).
- Added GitHub dependabot configuration to automate these dependency updates for GitHub actions as well as Cargo (dependabot will create PRs monthly to update these dependencies).
- Fix spacing in the changelog.

For #349

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ jobs:
   create-release:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Create release for tag
         if: startsWith(github.ref, 'refs/tags/')
         run: |
@@ -23,7 +23,7 @@ jobs:
       matrix:
         target: ["bash", "fish", "zsh"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Upload completion
         if: startsWith(github.ref, 'refs/tags/')
         run: |
@@ -39,7 +39,7 @@ jobs:
       matrix:
         target: ["MIT", "APACHE"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Upload license
         if: startsWith(github.ref, 'refs/tags/')
         run: |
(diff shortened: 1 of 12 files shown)


niklasmohrin (approved):
Thanks!

Not sure how I feel about dependabot for cargo, especially in regard to keeping a minimum supported Rust version (MSRV), but we can add the change and see how it goes. What do you think? @dbrgn

dbrgn:
I agree with @niklasmohrin. The "github-actions" dependabot check is great, but for Cargo we already get dependabot PRs for security issues. For regular updates, I'd prefer to do those manually, so it would be nice if you could remove the `package-ecosystem: "cargo"` config.

CHANGELOG.md
```diff
@@ -13,7 +13,6 @@ Possible log types:
 - `[docs]` for documentation changes.
 - `[chore]` for maintenance work.
 
-
```
dbrgn, line 16: Could you please revert this reformatting of an otherwise unchanged file?

.github/workflows/release.yml
```diff
@@ -65,44 +65,44 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
       - name: Build
         run: cargo build --release --target x86_64-apple-darwin --no-default-features --features webpki-roots
-      - uses: actions/upload-artifact@v3
```
dbrgn, line 90: Did you check the changelog of these actions (this one, and also `download-artifact` below), to ensure that there were no breaking changes?

(Since these actions are part of the release workflow, they don't run on regular testing CI runs.)

kbdharun: > Did you check the changelog of these actions (this one, and also download-artifact below), to ensure that there were no breaking changes?

Yes, I checked both the action's changelog's there seem to be no breaking changes.

CHANGELOG.md
kbdharun: Done

adamazing:
If this PR is close to merging and it doesn't hold it back, maybe the change to run the gh-pages action only on release tags could be done here?

If you want to keep it separate I can do a quick PR to do that if it would help?

Re: https://github.com/dbrgn/tealdeer/issues/358#issuecomment-2045967635

niklasmohrin:
Lets do it separately

niklasmohrin:
If you want to avoid the conflicts, I can merge this in the next couple of days

src/types.rs
```diff
@@ -14,6 +14,9 @@ pub enum PlatformType {
     SunOs,
     Windows,
     Android,
+    FreeBSD,
+    NetBSD,
+    OpenBSD,
```
niklasmohrin, line 19: One final thing: As per https://github.com/rust-lang/api-guidelines/blob/385ba009b19672a373cc65c98368527d2de82482/src/naming.md#casing-conforms-to-rfc-430-c-case , these should be `FreeBsd`, `NetBsd`, and `OpenBsd` instead. Other than that, we are good to go here

src/types.rs
```diff
@@ -73,6 +76,21 @@ impl PlatformType {
...
+    pub fn current() -> Self {
+        Self::FreeBSD
+    }
+
+    #[cfg(any(target_os = "netbsd",))]
+    pub fn current() -> Self {
+        Self::NetBSD
+    }
+
+    #[cfg(any(target_os = "openbsd",))]
```
niklasmohrin, line 89:
```suggestion
    #[cfg(any(target_os = "freebsd"))]
    pub fn current() -> Self {
        Self::FreeBSD
    }

    #[cfg(any(target_os = "netbsd"))]
    pub fn current() -> Self {
        Self::NetBSD
    }

    #[cfg(any(target_os = "openbsd"))]
```

src/types.rs
kbdharun: Noted, will fix it in a couple of minutes.

niklasmohrin (approved):

niklasmohrin:
Thanks!

PR #355, opened 2024-02-22
performance optimize with cargo flags

Walker-00:
```
Benchmark 1: tldr ip
  Time (mean ± σ):       2.9 ms ±   0.2 ms    [User: 1.0 ms, System: 1.6 ms]
  Range (min … max):     2.0 ms …   3.6 ms    500 runs

Benchmark 2: ./target/release/tldr ip
  Time (mean ± σ):       2.6 ms ±   0.2 ms    [User: 1.2 ms, System: 1.1 ms]
  Range (min … max):     1.8 ms …   3.4 ms    500 runs

Summary
  ./target/release/tldr ip ran
    1.15 ± 0.13 times faster than tldr ip
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,4 +62,7 @@ webpki-roots = ["reqwest/rustls-tls-webpki-roots"]
 native-tls = ["reqwest/native-tls"]
 
 [profile.release]
+strip = true
+opt-level = 3
 lto = true
+codegen-units = 1


niklasmohrin (approved):
Thanks!

Do we still need https://github.com/dbrgn/tealdeer/blob/17c08f026180204a47d7a1914c36de61e94ff991/.github/workflows/release.yml#L73 ?

Walker-00:
> Thanks!
> 
> Do we still need
> 
> https://github.com/dbrgn/tealdeer/blob/17c08f026180204a47d7a1914c36de61e94ff991/.github/workflows/release.yml#L73
> ?

I don't know but the performance really improved, right?

Walker-00:
and also sorry for late reply, reinstalling arch in slow internet connection is a nightmare.

dbrgn (approved):

PR #357, opened 2024-03-05
Updated Benchmark and Fixed Code For Purity and Readable

Walker-00:
Since The benchmark is made at 2021, almost everything has changed not just for tealdeer almost every tldr clients has been changed so I made a re benchmarking and rewrite some code for readable purpose.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable, 1.76.0]
+        toolchain: [stable, 1.65.0]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable, 1.64.0]
+        toolchain: [stable, 1.76.0]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
@@ -60,7 +60,7 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: '0.4.4'
+          mdbook-version: '0.4.37'
       - name: Setup toolchain
         uses: dtolnay/rust-toolchain@master
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,8 +66,3 @@ strip = true
 opt-level = 3
 lto = true
 codegen-units = 1
-
-[profile.release.package."*"]
-strip = true
-opt-level = 3
-codegen-units = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,8 @@ strip = true
 opt-level = 3
 lto = true
 codegen-units = 1
+
+[profile.release.package."*"]
+strip = true
+opt-level = 3
+codegen-units = 1
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,10 +28,9 @@
         feature = "native-tls"
     )),
 ))]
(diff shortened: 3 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable, 1.64.0]
+        toolchain: [stable, 1.76.0]
```
niklasmohrin, line 17: What is the motivation for this change? Your commit message suggests that you intended to update the version that is used for the release build, but we are already using the latest stable version for that. This number here is our Minimum Supported Rust Version (MSRV) and it should be kept as low as possible with the current code. Is there anything that we need from 1.76?

Cargo.toml
```diff
@@ -66,3 +66,8 @@ strip = true
 opt-level = 3
 lto = true
 codegen-units = 1
+
+[profile.release.package."*"]
```
niklasmohrin, line 70: What does this do? Dont we only have one package in our repo anyways? Or does it apply to our dependencies? (Although I would expect that the other config block above this one covers everything)

src/main.rs
```diff
@@ -200,7 +204,7 @@ fn init_log() {
 }
 
 #[cfg(not(feature = "logging"))]
-fn init_log() {}
+const fn init_log() {}
```
niklasmohrin, line 207: This makes the signature incompatible with the other `init_log` (when logging is enabled). Please revert

src/main.rs
```diff
@@ -209,7 +213,9 @@ fn get_languages(env_lang: Option<&str>, env_language: Option<&str>) -> Vec<Stri
     if env_lang.is_none() {
         return vec!["en".to_string()];
     }
-    let env_lang = env_lang.unwrap();
+
+    let env_lang = env_lang
+        .expect("Sorry It Shouldn't Be Happened We've checked and the env lang is not None!");
```
niklasmohrin, line 218: I think the version with `unwrap` is better here. If we really want to get rid of unwrap, we should use a let-else statement (which was not stable when this code was written)

src/main.rs
```diff
@@ -26,9 +28,10 @@
         feature = "native-tls"
     )),
 ))]
-compile_error!(
+/*compile_error!(
     "exactly one of the features \"native-roots\", \"webpki-roots\" or \"native-tls\" must be enabled"
-);
+);*/
+print!("hi");
```
niklasmohrin, line 34: Please revert

niklasmohrin (changes requested):
Haven't looked into the outfieldr stuff yet, although I think that if we dont get it compile we should just remove it. 

Was the impact of all the `const` significant?

.github/workflows/ci.yml
Walker-00: Yea, my mistake when I update the dep crates `regex-syntax v0.8.2` need MSRV of 1.65.0, Sorry I'll fix it.

src/main.rs
Walker-00: Yea I forget to delete that and expect code, I used them to ignore the errors msg from clippy and lsp, Sorry again I'll fix it.

Cargo.toml
Walker-00: Think it would cool to add that :3

niklasmohrin:
@Walker-00 I take it you do not want to continue working on this (which is totally fine)? Otherwise, I would suggest that we split the code changes and the benchmark changes into two separate PRs to make reviewing easier

Walker-00:
> @Walker-00 I take it you do not want to continue working on this (which is totally fine)? Otherwise, I would suggest that we split the code changes and the benchmark changes into two separate PRs to make reviewing easier

Nah bro, I really enjoy contributing to this pj. But I'm a bit busy lately with my startup. I'll keep contributing when I'm free.

niklasmohrin:
Okay, no worries

Walker-00:
yea bro, so should I reopen pull req

niklasmohrin:
Pleaae leave it closed until you want to resume working on it :)

Walker-00:
oki bro

PR #372, opened 2024-07-22
Replace app_dirs2 with directories

nc7s:
Main reason is to cut down number of little used dependencies in Debian. (app_dirs2 is only used by tealdeer, for example.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,18 +73,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
-name = "app_dirs2"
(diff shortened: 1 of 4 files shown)


dbrgn:
Potentially related: https://github.com/dbrgn/tealdeer/pull/319

nc7s:
Probably not? AFAIK app_dirs2 and directories both just follow OS conventions, while to respect XDG_* variables is an application level decision, as of now at least.

niklasmohrin:
To be honest, I would prefer sticking to app_dirs for the time being to avoid switching libraries more than once here. How much pain is app_dirs right now? If there is no urgent problem, I would favor keeping it around for a little longer until we figured out #319

nc7s:
Sure, you call ;)

niklasmohrin:
Okay I would pause this for now then :)

PR #369, opened 2024-06-17
Prepare release 1.7.0

dbrgn:
Did we forget anything?

Things that need to be done before the release:

- [x] Small tweaks and improved documentation for "multiple platforms" feature: https://github.com/dbrgn/tealdeer/pull/300#issuecomment-1925882814
- [x] Bump dependencies:  #374

...anything else?

PR #374, opened 2024-08-12
Update Dependencies

dbrgn:
...in preparation for the 1.7.0 release (#369).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,6 +94,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arbitrary"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d5a26814d8dcb93b0e5a0ff3c6d80a8843bafb21b39e8e18a6f05471870e110"
+dependencies = [
+ "derive_arbitrary",
+]
+
+[[package]]
 name = "assert_cmd"
 version = "2.0.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


dbrgn:
Release binary size increased from 3.66 MB to 3.80 MB. Largest 10 crate-level contributors after the updates:

```
$ cargo bloat --release --crates -n 10
    Finished `release` profile [optimized] target(s) in 0.06s
    Analyzing target/release/tldr

 File  .text     Size Crate
 8.8%  27.3% 704.1KiB std
 4.2%  13.0% 336.3KiB rustls
 3.1%   9.5% 244.5KiB clap_builder
 2.6%   8.0% 207.0KiB ring
 1.9%   5.9% 151.4KiB toml_edit
 1.6%   4.9% 127.3KiB tldr
 1.5%   4.5% 116.9KiB [Unknown]
 1.2%   3.6%  92.8KiB reqwest
 1.0%   3.2%  82.3KiB hyper
 0.8%   2.4%  61.9KiB hyper_util
 5.4%  16.7% 430.4KiB And 51 more crates. Use -n N to show more.
32.5% 100.0%   2.5MiB .text section size, the file size is 7.8MiB

Note: numbers above are a result of guesswork. They are not 100% correct and never will be.
```

RusTLS is quite significant, but I think having it bundled is worth it because it gets rid of many platform-specific issues.

A release build with `--no-default-features --features native-tls` has 2.61 MB, which is significantly smaller. If some folks want to have a small binary and have a proper TLS provider on their system, that is a good option.

niklasmohrin (approved):

PR #380, opened 2024-10-02
Replace `atty` with `std::io::IsTerminal`

niklasmohrin:
The `atty` crate is unmaintained, `std::io::IsTerminal` is stable since Rust 1.70.


dbrgn (approved):
Very nice 🙂

PR #382, opened 2024-10-04
Fix help message

se-jaeger:
Add the correct documentation URL.


kbdharun (commented):
Hi, are you sure you have pushed your changes, there seems to be no lines modified in your commit.

dbrgn:
As @kbdharun wrote, your branch does not contain any differences compared to `main`. Additionally, I checked the `tldr --help` output of the current release (1.7.0) and the documentation URL is correct there.

se-jaeger:
Ok wow, that's weird...  I edited and committed via the web UI and maybe the changes got lost somewhere.

Sorry for the annoyance.

PR #386, opened 2024-10-27
Allows configuring TLS backend

### LINKED ISSUE #384: Configure TLS at runtime
We keep running into problems where one TLS configuration or another breaks (see #244, #328, #330). We then usually have to tell people to try compiling with another compile feature.

The library `reqwests`, which is used for making the requests, can be configured at runtime: https://docs.rs/reqwest/0.12.8/reqwest/struct.ClientBuilder.html#method.tls_built_in_root_certs

It should be possible to use these methods, so that we can tell users to change configuration options instead of recompiling. We will still keep the features, so that smaller binaries can be created. However, I think our default feature-set should be to enable all options.

To get going, we only need support in the config file, not via CLI flags or environment variables. The option should be in the `updates` section. I am not settled on any particular name for the option, but something like "tls-backend" or so should work to get started. (It is possible that it will turn out nicer with two options). We will figure out the specifics once we have some code to talk about.

erickguan:
Fixes #384.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
 env_logger = { version = "0.11", optional = true }
 log = "0.4"
-reqwest = { version = "0.12.5", features = ["blocking"], default-features = false }
+reqwest = { version = "0.12.5", features = ["blocking", "native-tls", "rustls-tls", "rustls-tls-native-roots", "rustls-tls-webpki-roots"], default-features = false }
 serde = "1.0.21"
 serde_derive = "1.0.21"
 toml = "0.8.19"
@@ -56,6 +56,7 @@ logging = ["env_logger"]
 #
 # Exactly one of the three variants must be selected. By default, Rustls with
 # native roots is enabled.
+# When you compile the crate, you should also set the `tls_backend` config with the default feature.
 native-roots = ["reqwest/rustls-tls-native-roots"]
 webpki-roots = ["reqwest/rustls-tls-webpki-roots"]
 native-tls = ["reqwest/native-tls"]
--- a/docs/src/config_updates.md
+++ b/docs/src/config_updates.md
@@ -24,3 +24,17 @@ is set to `false`.
     auto_update = true
     auto_update_interval_hours = 24
 
+### `tls_backend`
+
+An advance option. Specifies which TLS backend to use. Only modify this if you encounter certificate errors.
+
+Available options:
+- `native-roots` - Rustls with native roots
+- `webpki-roots` - Rustls with WebPK roots
+- `native-tls` - Native TLS
+  - SChannel on Windows
+  - Secure Transport on macOS
+  - OpenSSL on other platforms
+
+    [updates]
+    tls_backend = "native-tls"
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -19,9 +19,10 @@ pub static TLDR_PAGES_DIR: &str = "tldr-pages";
 static TLDR_OLD_PAGES_DIR: &str = "tldr-master";
 
 #[derive(Debug)]
-pub struct Cache {
+pub struct Cache<'a> {
     cache_dir: PathBuf,
     enable_styles: bool,
+    tls_backend: &'a str, // for setting up reqwest client
 }
 
 #[derive(Debug)]
@@ -85,14 +86,15 @@ pub enum CacheFreshness {
     Missing,
 }
 
-impl Cache {
-    pub fn new<P>(cache_dir: P, enable_styles: bool) -> Self
+impl<'a> Cache<'a> {
+    pub fn new<P>(cache_dir: P, enable_styles: bool, tls_backend: &'a str) -> Self
     where
         P: Into<PathBuf>,
     {
         Self {
             cache_dir: cache_dir.into(),
             enable_styles,
+            tls_backend,
         }
     }
 
@@ -135,9 +137,17 @@ impl Cache {
         self.cache_dir.join(TLDR_PAGES_DIR)
     }
 
-    /// Download the archive from the specified URL.
-    fn download(archive_url: &str) -> Result<Vec<u8>> {
+    /// Builds HTTPS client based on configuration.
+    ///
+    /// Note that `Cargo.toml` also defines default feature .
+    fn build_client(&self) -> Result<Client> {
         let mut builder = Client::builder();
+        builder = match self.tls_backend {
+            "native-roots" => builder.use_rustls_tls().tls_built_in_native_certs(true),
+            "webpki-roots" => builder.use_rustls_tls().tls_built_in_webpki_certs(true),
+            "native-tls" => builder.use_native_tls().tls_built_in_native_certs(true),
+            _ => builder, // TLS backend defaults to the compiled feature
+        };
         if let Ok(ref host) = env::var("HTTP_PROXY") {
             if let Ok(proxy) = Proxy::http(host) {
                 builder = builder.proxy(proxy);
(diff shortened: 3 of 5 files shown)


dbrgn:
Hm, I wonder if we shouldn't just switch to native roots and call it a day... What's your opinion @niklasmohrin?

erickguan:
FYI, the default build uses the native roots.

I didn't change the build default in this PR. I can see some adjustments required and your inputs.

niklasmohrin:
@dbrgn I have some thoughts written down in the issue, I think that having the functionality is nice and people can still build it without the bloat by disabling features

src/cache.rs
```diff
@@ -504,4 +517,30 @@ mod tests {
...
+            $(
+                #[test]
+                fn $name() {
+                    let dir = tempfile::tempdir().unwrap();
+
+                    let _ = Cache::build_client(&Cache {
+                        cache_dir: dir.into_path(),
+                        enable_styles: false,
+                        tls_backend: $backend,
+                    });
```
niklasmohrin, line 532: We are not checking whether the returned `Result` of `build_client` is `Ok` here - let's do so

Additionally, I think we can make `build_client` not take the whole `Cache` but just the `TlsBackend` parameter. Then we won't need the tempdir and we can then remove the macro too and write out the three tests

niklasmohrin (changes requested):
Thanks! I have some requests regarding the code first, once we have this ironed out I will take a closer look at the documentation. Please feel free to reach out if you have any questions or thoughts :)

Cargo.toml
```diff
@@ -25,7 +25,7 @@ app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
 env_logger = { version = "0.11", optional = true }
 log = "0.4"
-reqwest = { version = "0.12.5", features = ["blocking"], default-features = false }
+reqwest = { version = "0.12.5", features = ["blocking", "native-tls", "rustls-tls", "rustls-tls-native-roots", "rustls-tls-webpki-roots"], default-features = false }
```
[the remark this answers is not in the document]
erickguan: I see. Fixed. I saw the reqwest 0.12.9 supports additional roots. So I also extended it. What do you think about it?

src/cache.rs
```diff
@@ -19,9 +19,10 @@ pub static TLDR_PAGES_DIR: &str = "tldr-pages";
 static TLDR_OLD_PAGES_DIR: &str = "tldr-master";
 
 #[derive(Debug)]
-pub struct Cache {
+pub struct Cache<'a> {
     cache_dir: PathBuf,
     enable_styles: bool,
+    tls_backend: &'a str, // for setting up reqwest client
```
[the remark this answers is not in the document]
erickguan: Of course!

niklasmohrin (changes requested):
I finally found some time to take another look :)

erickguan (commented):
Thanks for the review!

I implemented `tls_backend` as a somewhat special config to bubble up config conversation error based on your idea.

src/config.rs
```diff
@@ -329,6 +388,21 @@ impl Config {
...
+            RawTlsBackend::Rustls => TlsBackend::Rustls,
+            #[cfg(feature = "rustls-with-native-roots")]
+            RawTlsBackend::RustlsWithNativeRoots => TlsBackend::RustlsWithNativeRoots,
+            // when compiling without all TLS backend features, we want to handle config error.
+            #[allow(unreachable_patterns)]
+            _ => return Err(anyhow!(
+                    "Unsupported TLS backend: {}. This tealdeer build has support for the following options: {}",
+                 raw_config.updates.tls_backend,
+                 SUPPORTED_TLS_BACKENDS.map(|b| b.to_string()).join(", ")
+                ))
```
niklasmohrin, line 404: Very nice error message! Let's add a test that checks the output if this case occurs. To do so, you can use for example `let testenv = TestEnv::new().no_default_features().with_feature("rustls");` in our integration tests at `tests/lib.rs`.

I discovered that our implementation of `TestEnv::command` is not correct at the moment, sorry about that. You should change line 123 of `tests/lib.rs` from
```rust
            build = build.arg(format!("--feature {}", self.features.join(",")));
```
to
```rust
            build = build.arg("--features").arg(self.features.join(","));
```

niklasmohrin (changes requested):

erickguan (commented):
I rebased and pushed the PR to catch up with your fix on the master branch.

Updated feature names and added test cases along with other comments.

src/config.rs
```diff
@@ -291,11 +303,49 @@ pub struct DirectoriesConfig {
...
+    #[cfg(feature = "rustls-with-native-roots")]
+    RustlsWithNativeRoots,
+}
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Config {
     pub style: StyleConfig,
     pub display: DisplayConfig,
     pub updates: UpdatesConfig,
+    pub tls_backend: TlsBackend,
```
niklasmohrin, line 348: I think we can move `tls_backend` into the `UpdatesConfig` instead of having it directly in `Config`

docs/src/config_updates.md
```diff
@@ -32,3 +32,21 @@ fetched from the latest `tldr-pages/tldr` GitHub release.
     [updates]
     archive_source = https://my-company.example.com/tldr/
 
+### `tls_backend`
+
+An advance option. Specifies which TLS backend to use. Only modify this if you encounter certificate errors.
+
+Available options:
+- `rusttls-with-native-roots` - [Rustls][rustls] (a TLS library in Rust) with native roots
+- `rusttls-with-webpki-roots` - Rustls with [WebPKI][rustls-webpki] roots
```
niklasmohrin, line 41: Typos:
```suggestion
- `rustls-with-native-roots` - [Rustls][rustls] (a TLS library in Rust) with native roots
- `rustls-with-webpki-roots` - Rustls with [WebPKI][rustls-webpki] roots
```

docs/src/config_updates.md
```diff
@@ -32,3 +32,21 @@ fetched from the latest `tldr-pages/tldr` GitHub release.
     [updates]
     archive_source = https://my-company.example.com/tldr/
 
+### `tls_backend`
+
+An advance option. Specifies which TLS backend to use. Only modify this if you encounter certificate errors.
```
niklasmohrin, line 37: What do you think about the following?
```suggestion
Specifies which TLS backend to use. Try changing this setting if you encounter certificate errors.
```

Cargo.toml
```diff
@@ -25,7 +25,7 @@ app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
 env_logger = { version = "0.11", optional = true }
 log = "0.4"
-reqwest = { version = "0.12.5", features = ["blocking"], default-features = false }
+reqwest = { version = "0.12.9", features = ["blocking"], default-features = false }
```
niklasmohrin, line 28: Since the last time, I have found the following [ureq](https://crates.io/crates/ureq). It's a blocking HTTP request library which also supports all the TLS features we need and is probably not as heavy as reqwest. Since you are already caught up on the topic, what do you think about it?

niklasmohrin (commented):
Sorry for the long wait again, but here are my final comments - afterwards, I think we are ready to merge :)

In my last comment I mention a possible replacement for reqwest - no worries, I think it is still best to merge your PR for reqwest first, and then think about using the other library. In fact, if you would want to work on porting tealdeer to use ureq instead of reqwest, feel more than welcome, as you are already informed about the topic!

Cargo.toml
erickguan: `ureq` Looks promising. I want to see the build size with `reqwest` difference first. The compiler might be too good to remove the unused code to see a difference. We will see from the data.

erickguan (commented):
I understand.

Certainly, I am interested in `ureq` work. More details in the inline comment!

niklasmohrin (approved):
Very cool, thanks!

Cargo.toml
niklasmohrin: Cool, see #415

PR #388, opened 2024-11-02
Add args: `--edit-page` and `--edit-patch`

lengyijun:
Fix https://github.com/tealdeer-rs/tealdeer/issues/383

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -32,6 +32,14 @@ pub(crate) struct Cli {
     #[arg(short = 'l', long = "list")]
     pub list: bool,
 
+    /// Edit custom page with `EDITOR`
+    #[arg(long, requires = "command")]
+    pub edit_page: bool,
+
+    /// Edit custom patch with `EDITOR`
+    #[arg(long, requires = "command", conflicts_with = "edit_page")]
+    pub edit_patch: bool,
+
     /// Render a specific markdown file
     #[arg(
         short = 'f',
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,10 +32,15 @@ compile_error!(
 
 use std::{
     env,
+    fs::create_dir_all,
     io::{self, IsTerminal},
+    path::Path,
     process,
+    process::Command,
 };
 
+use anyhow::anyhow;
+use anyhow::Context;
 use app_dirs::AppInfo;
 use clap::Parser;
 
@@ -240,6 +245,32 @@ fn get_languages_from_env() -> Vec<String> {
     )
 }
 
+fn spawn_editor(custom_pages_dir: &Path, file_name: &str) -> anyhow::Result<()> {
+    let res = create_dir_all(custom_pages_dir)
+        .context(format!("Fail to create dir {:?}", custom_pages_dir));
+    if !custom_pages_dir.is_dir() {
+        res?;
+        return Err(anyhow!("Fail to create dir {:?}", custom_pages_dir));
+    }
+
+    let custom_page_path = custom_pages_dir.join(file_name);
+    let Some(custom_page_path) = custom_page_path.to_str() else {
+        return Err(anyhow!("`custom_page_path.to_str()` failed"));
+    };
+    let Ok(editor) = env::var("EDITOR") else {
+        return Err(anyhow!(
+            "To edit a custom page, please set the `EDITOR` environment variable."
+        ));
+    };
+    println!("Editing {custom_page_path:?}");
+
+    let status = Command::new(&editor).arg(custom_page_path).status()?;
+    if !status.success() {
+        return Err(anyhow!("{editor} exit with code {:?}", status.code()));
+    }
+    Ok(())
+}
+
 fn main() {
     // Initialize logger
     init_log();
@@ -270,6 +301,34 @@ fn main() {
             process::exit(1);
         }
     };
+
+    let custom_pages_dir = config
+        .directories
+        .custom_pages_dir
+        .as_ref()
+        .map(PathWithSource::path);
+
+    // Note: According to the TLDR client spec, page names must be transparently
+    // lowercased before lookup:
+    // https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#page-names
+    let command = args.command.join("-").to_lowercase();
+
+    if args.edit_patch || args.edit_page {
+        let file_name = if args.edit_patch {
+            format!("{command}.patch.md")
+        } else {
+            format!("{command}.page.md")
+        };
+
+        custom_pages_dir
+            .context("To edit custom pages/patches, please specify a custom pages directory.")
+            .and_then(|custom_pages_dir| spawn_editor(custom_pages_dir, &file_name))
+            .unwrap_or_else(|err| {
+                print_error(enable_styles, &err);
+                process::exit(1);
+            });
+        return;
+    }
 
     // Show various paths
     if args.show_paths {
@@ -313,7 +372,7 @@ fn main() {
 
     // Check cache presence and freshness
     if !cache_updated
-        && (args.list || !args.command.is_empty())
+        && (args.list || !command.is_empty())
         && check_cache(&cache, &args, enable_styles) == CheckCacheResult::CacheMissing
     {
         process::exit(1);
@@ -321,11 +380,6 @@ fn main() {
 
     // List cached commands and exit
     if args.list {
-        let custom_pages_dir = config
-            .directories
-            .custom_pages_dir
-            .as_ref()
-            .map(PathWithSource::path);
         println!(
             "{}",
             cache.list_pages(custom_pages_dir, &platforms).join("\n")
@@ -334,12 +388,7 @@ fn main() {
     }
 
     // Show command from cache
-    if !args.command.is_empty() {
-        // Note: According to the TLDR client spec, page names must be transparently
-        // lowercased before lookup:
-        // https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#page-names
-        let command = args.command.join("-").to_lowercase();
-
+    if !command.is_empty() {
         // Collect languages
         let languages = args
             .language
--- a/src/main.rs
+++ b/src/main.rs
@@ -246,12 +246,7 @@ fn get_languages_from_env() -> Vec<String> {
 }
 
 fn spawn_editor(custom_pages_dir: &Path, file_name: &str) -> anyhow::Result<()> {
-    let res = create_dir_all(custom_pages_dir)
-        .context(format!("Fail to create dir {:?}", custom_pages_dir));
-    if !custom_pages_dir.is_dir() {
-        res?;
-        return Err(anyhow!("Fail to create dir {:?}", custom_pages_dir));
-    }
+    create_dir_all(custom_pages_dir).context("Failed to create custom pages directory")?;
 
     let custom_page_path = custom_pages_dir.join(file_name);
     let Some(custom_page_path) = custom_page_path.to_str() else {
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -990,3 +990,67 @@ fn test_raw_render_file() {
         .success()
         .stdout(diff(include_str!("cache/pages/common/inkscape-v1.md")));
 }
+
+fn touch_custom_page(testenv: &TestEnv) {
+    let args = vec!["--edit-page", "foo"];
+
+    testenv
+        .command()
+        .args(&args)
+        .env("EDITOR", "touch")
+        .assert()
+        .success();
+    assert!(testenv.custom_pages_dir.path().join("foo.page.md").exists());
+}
+
+fn touch_custom_patch(testenv: &TestEnv) {
+    let args = vec!["--edit-patch", "foo"];
+
+    testenv
+        .command()
+        .args(&args)
+        .env("EDITOR", "touch")
+        .assert()
+        .success();
+    assert!(testenv
+        .custom_pages_dir
+        .path()
+        .join("foo.patch.md")
+        .exists());
+}
+
+#[test]
+fn test_edit_page() {
+    let testenv = TestEnv::new().write_custom_pages_config();
+    touch_custom_page(&testenv);
+}
+
+#[test]
+fn test_edit_patch() {
+    let testenv = TestEnv::new().write_custom_pages_config();
+    touch_custom_patch(&testenv);
+}
+
+#[test]
+fn test_recreate_dir() {
+    let testenv = TestEnv::new().write_custom_pages_config();
+    touch_custom_patch(&testenv);
+    touch_custom_page(&testenv);
+}
+
+#[test]
+fn test_custom_pages_dir_is_not_dir() {
+    let testenv = TestEnv::new().write_custom_pages_config();
+    let _ = std::fs::remove_dir_all(testenv.custom_pages_dir.path());
+    let _ = File::create(testenv.custom_pages_dir.path()).unwrap();
+    assert!(testenv.custom_pages_dir.path().is_file());
+
+    let args = vec!["--edit-patch", "foo"];
+
+    testenv
+        .command()
+        .args(&args)
+        .env("EDITOR", "touch")
+        .assert()
+        .failure();
+}
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -8,6 +8,8 @@ Arguments:
 
 Options:
   -l, --list                 List all commands in the cache
+      --edit-page            Edit custom page with `EDITOR`
+      --edit-patch           Edit custom patch with `EDITOR`
   -f, --render <FILE>        Render a specific markdown file
   -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
                              of preference [possible values: linux, macos, sunos, windows, android,


src/cli.rs
```diff
@@ -32,6 +32,14 @@ pub(crate) struct Cli {
     #[arg(short = 'l', long = "list")]
     pub list: bool,
 
+    /// Edit custom patch with `EDITOR`
+    #[arg(long = "edit-patch")]
+    pub edit_patch: bool,
+
+    /// Edit custom page with `EDITOR`
+    #[arg(long = "edit-page")]
+    pub edit_page: bool,
```
niklasmohrin, line 41: The defaults from clap already replace `_` with `-`, so this is equivalent:
```suggestion
    /// Edit custom patch with `EDITOR`
    #[arg(long)]
    pub edit_patch: bool,

    /// Edit custom page with `EDITOR`
    #[arg(long)]
    pub edit_page: bool,
```

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
...
+    if args.edit_page && args.edit_patch {
+        print_error(
+            enable_styles,
+            &anyhow!("Don't edit-patch and edit-page at the same time"),
+        );
+        process::exit(1);
+    }
+
+    if args.edit_patch || args.edit_page {
+        if args.command.is_empty() {
```
niklasmohrin, line 290: This can be achieved with clap using "requires", for example
```rust
#[arg(long, requires = "command")]
pub edit_page: bool,
```

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
         }
     };
 
+    if args.edit_page && args.edit_patch {
+        print_error(
+            enable_styles,
+            &anyhow!("Don't edit-patch and edit-page at the same time"),
+        );
+        process::exit(1);
+    }
```
niklasmohrin, line 287: This can done using clap's `conflicts_with` on one of the two options, for example
```rust
#[arg(long, conflicts_with = "edit_page")]
pub edit_patch: bool,
```

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
...
+                );
+            }
+            process::exit(1);
+        }
+        let command = args.command.join("-").to_lowercase();
+        let custom_page_dir = config
+            .directories
+            .custom_pages_dir
+            .as_ref()
+            .map(PathWithSource::path);
```
niklasmohrin, line 322: This is now duplicated with the code at the end of `main`. Can you make it so that both places use the same variable?

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
...
+            println!("Editing {custom_page_path:?}");
+            let editor = env::var("EDITOR").context("env `EDITOR` not set").unwrap();
+            let _ = std::process::Command::new(editor)
+                .arg(custom_page_path.to_str().unwrap())
+                .status();
+            return;
+        } else {
+            print_error(enable_styles, &anyhow!("Fail to get custom page dir"));
+            process::exit(1);
+        }
```
niklasmohrin, line 341: I think this would be nicer if the "else" wouldn't be so far away. How do you feel about refactoring it to be
```rust
let Some(custom_page_dir) = custom_page_dir else {
    // fail
};
```

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
...
+
+            let file_name = if args.edit_patch {
+                format!("{command}.patch.md")
+            } else {
+                format!("{command}.page.md")
+            };
+            let custom_page_path = custom_page_dir.join(file_name);
+            println!("Editing {custom_page_path:?}");
+            let editor = env::var("EDITOR").context("env `EDITOR` not set").unwrap();
+            let _ = std::process::Command::new(editor)
```
niklasmohrin, line 334: Right now, this does not work if people set EDITOR to something that includes arguments, for example `EDITOR="nvim --clean" tldr --edit-page foo`. I checked a couple of programs for how they handle this. Both `git config --edit` and `sudoedit` just spawn nvim without arguments, omitting the `--clean`. I suppose that having arguments in EDITOR is not a common thing then.

I think that crashing is better than omitting the arguments, so the current handling of EDITOR is fine. However, we should handle and print errors returned here.

Additionally, we should think about adding some documentation for this behavior, but let's postpone this until the code is ready.

src/main.rs
```diff
@@ -276,6 +278,69 @@ fn main() {
...
+            let _ = create_dir_all(custom_page_dir);
+
+            let file_name = if args.edit_patch {
+                format!("{command}.patch.md")
+            } else {
+                format!("{command}.page.md")
+            };
+            let custom_page_path = custom_page_dir.join(file_name);
+            println!("Editing {custom_page_path:?}");
+            let editor = env::var("EDITOR").context("env `EDITOR` not set").unwrap();
```
niklasmohrin, line 333: I think we should handle the absence of EDITOR more gracefully, maybe printing some advice and then exiting with exit code 1 or something similar? What do you think?

niklasmohrin (changes requested):
Thanks! In addition to the comments, the added functionality is not tested by our automated tests yet. I think the tests could run `tldr --edit-page foo` with `EDITOR=touch` and the custom pages dir set to some temporary directory. Then we can observe that the correct command `touch /some/tmp/dir/foo.page.md` is run by checking whether the file exists or not.

Please feel free to reach out for questions, thoughts, or help :)

src/main.rs
lengyijun: we can use shlex to parse

src/cli.rs
```diff
@@ -32,6 +32,14 @@ pub(crate) struct Cli {
...
     pub list: bool,
 
+    /// Edit custom patch with `EDITOR`
+    #[arg(long, requires = "command", conflicts_with = "edit_page")]
+    pub edit_patch: bool,
+
+    /// Edit custom page with `EDITOR`
+    #[arg(long, requires = "command")]
+    pub edit_page: bool,
+
```
niklasmohrin, line 42: I think I would switch the ordering of these two, because we also display patches after pages - it's always page & patch

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+        .map(PathWithSource::path);
+
+    // Note: According to the TLDR client spec, page names must be transparently
+    // lowercased before lookup:
+    // https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#page-names
+    let command = args.command.join("-").to_lowercase();
+
+    if args.edit_patch || args.edit_page {
+        let Some(custom_pages_dir) = custom_pages_dir else {
+            print_error(enable_styles, &anyhow!("Fail to get custom page dir"));
```
niklasmohrin, line 290: How about "To edit custom pages/patches, please specify a custom pages directory."?

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+            process::exit(1);
+        };
+        let _ = create_dir_all(custom_pages_dir);
```
niklasmohrin, line 293: Why are we ignoring errors here? If this fails, the editing with `EDITOR` might not work afterwards, or am I not getting something?

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+        let file_name = if args.edit_patch {
+            format!("{command}.patch.md")
+        } else {
+            format!("{command}.page.md")
+        };
+        let custom_page_path = custom_pages_dir.join(file_name);
+        let Ok(editor) = env::var("EDITOR") else {
+            print_error(
+                enable_styles,
+                &anyhow!("`EDITOR` not set. Please set `EDITOR` by `export EDITOR=vim`."),
```
niklasmohrin, line 304: Maybe "To edit a custom page, please set the \`EDITOR\` environment variable."?

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+            );
+            process::exit(1);
+        };
+        println!("Editing {custom_page_path:?}");
+        match std::process::Command::new(&editor)
```
niklasmohrin, line 309: Let's `use std::process::Command`

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+            .arg(custom_page_path.to_str().unwrap())
```
niklasmohrin, line 310: The `unwrap` should also print a nice error, that is, use `print_error`

src/main.rs
```diff
@@ -272,6 +274,57 @@ fn main() {
...
+                if !status.success() {
+                    print_error(
+                        enable_styles,
+                        &anyhow!("{editor} exit with code {:?}", status.code()),
+                    );
+                }
+            }
+            Err(e) => {
+                print_error(enable_styles, &e.into());
+            }
```
niklasmohrin, line 323: In general, we use `print_error` quite often in the added code segment. Maybe it would be better if we made a function that returns `anyhow::Result<()>` and call it from `main`. What do you think?

niklasmohrin (changes requested):
Tests look good and the flags seem to work, only some code style things left :)

src/main.rs
niklasmohrin: Nah, as I said, the current behavior is probably exactly what we want to do. As long as the behavior is documented, I think we are good.

You could add a section at the end of `docs/src/usage_custom_pages.md` to highlight the feature and we can phrase it so that the behavior is clear

src/main.rs
lengyijun: Maybe custom_pages_dir exists already

src/main.rs
lengyijun: No idea

src/main.rs
niklasmohrin: I thought that this wouldn't error - let's add a test for this and then see if checking for errors here still works

src/main.rs
niklasmohrin: Okay, so let's try it with a function and see which version we like better :)

src/main.rs
lengyijun: https://doc.rust-lang.org/std/fs/fn.create_dir_all.html

niklasmohrin: Yes, I read that too, but it doesn't support your point. In fact, running a fresh project with just
```rust
fn main() {
    std::fs::create_dir_all("test/dir").unwrap();
    std::fs::create_dir_all("test/dir").unwrap();
}
```
still works - the function will not error if (part of) the path already exists.

lengyijun: Maybe on some platforms it will return err.
Anyway, if an io error happens, `EDITOR` will catch it too

niklasmohrin: No, we shouldn't rely on the editor behavior. In fact, for example neovim happily opens non-existent and impossible files, for example I can run `nvim Cargo.toml/foo` and will get an editor as usual.

I stand by my point, we should have tests to ensure that tealdeer still works if the directories exist. Once we have those, I think we will see that using `create_dir_all(...)?` should do what we want

src/main.rs
niklasmohrin: In particular, I think it is a lot more readable if we add a function `spawn_editor(path: &Path) -> anyhow::Result<()>` that creates the directory of `path` and so on, so that we can call it in some way like this from `main`:
```rust
use anyhow::Context;

// ...

fn spawn_editor(path: &Path) -> anyhow::Result<()> {
    // ...
}

fn main() {
    // ...

    if args.edit_patch || args.edit_page {
        let filename = if args.edit_patch {
            format!("{command}.patch.md")
        } else {
            format!("{command}.page.md")
        };

        custom_pages_dir
            .context("To edit custom pages/patches, please specify a custom pages directory.")
            .and_then(|custom_pages_dir| spawn_editor(&custom_pages_dir.join(filename)))
            .unwrap_or_else(|err| {
                print_error(enable_styles, &err);
                process::exit(1);
            });
        return;
    }

    // ...
}
```

niklasmohrin (changes requested):
I left some more comments on the previous discussions, please let me know what you think :)

src/main.rs
lengyijun: > nvim Cargo.toml/foo

 Then you can't save it

lengyijun: There is a check at L298-305

niklasmohrin: Yes, and by this, the experience is more frustrating than it has to be - let's not spawn the editor in this case 

Please add a test for when the directory already exists, and one for when there is a file where the directory should be, then we can check which implementation fulfills the requirements

niklasmohrin:
@lengyijun you marked https://github.com/tealdeer-rs/tealdeer/pull/388#discussion_r1900185647 as resolved now, but the most crucial test I brought up is still missing: when we run tealdeer a second time, we should not get an error because of trying to create the directory.

In general, I am a bit unsure on when I should take a look at the PR, please write a short comment or re-request my review on Github when you are ready :)

lengyijun:
Review requested

lengyijun:
Review requested

src/main.rs
```diff
@@ -240,6 +245,30 @@ fn get_languages_from_env() -> Vec<String> {
     )
 }
 
+fn spawn_editor(custom_pages_dir: &Path, file_name: &str) -> anyhow::Result<()> {
+    let _ = create_dir_all(custom_pages_dir);
+    if !custom_pages_dir.is_dir() {
+        return Err(anyhow!("Fail to create dir {:?}", custom_pages_dir));
+    }
```
niklasmohrin, line 252: Now that we have `test_recreate_dir`, we can see that doing the following change:
```suggestion
    create_dir_all(custom_pages_dir).context("Failed to create custom pages directory")?;
```
... does _not_ break `test_recreate_dir`, so we can use this after all :)

tests/lib.rs
```diff
@@ -990,3 +990,67 @@ fn test_raw_render_file() {
...
+    let testenv = TestEnv::new().write_custom_pages_config();
+    touch_custom_patch(&testenv);
+    touch_custom_page(&testenv);
+}
+
+#[test]
+fn test_custom_pages_dir_is_not_dir() {
+    let testenv = TestEnv::new().write_custom_pages_config();
+    let _ = std::fs::remove_dir_all(testenv.custom_pages_dir.path());
+    let _ = File::create(testenv.custom_pages_dir.path()).unwrap();
```
niklasmohrin, line 1045: I think this might be a problem, because the created file is not automatically deleted after the tests are finished. However, I think we should in general change our test environment to place everything in unified temporary directory instead of creating multiple ones. I will make a follow up issue, so we can leave it as is for now

niklasmohrin (commented):
Alright, I have one last comment to finish the `create_dir_all` discussion before we can merge, everything else looks good!

src/main.rs
```diff
@@ -240,6 +245,34 @@ fn get_languages_from_env() -> Vec<String> {
...
 
+fn spawn_editor(custom_pages_dir: &Path, file_name: &str) -> anyhow::Result<()> {
+    let res = create_dir_all(custom_pages_dir);
+    if !custom_pages_dir.is_dir() {
+        if res.is_err() {
+            res.context(format!("Fail to create dir {:?}", custom_pages_dir))?;
+        } else {
+            return Err(anyhow!("Fail to create dir {:?}", custom_pages_dir));
+        }
+    }
```
niklasmohrin, line 256: @lengyijun Can you please explain why you are avoiding to just do `create_dir_all(...).context(...)?`? I feel like we are going on and on about this change and as long as we don't have clear evidence for why we _shouldn't_ use the simple version, I am not willing to merge anything more complicated. Your initial argument was that `create_dir_all` may return `Err` if the directory already exists, but we added a test that checks that this is not the case. What is there left to discuss?

lengyijun:
a simpler version now, review requested

src/main.rs
```diff
@@ -240,6 +245,32 @@ fn get_languages_from_env() -> Vec<String> {
     )
 }
 
+fn spawn_editor(custom_pages_dir: &Path, file_name: &str) -> anyhow::Result<()> {
+    let res = create_dir_all(custom_pages_dir)
+        .context(format!("Fail to create dir {:?}", custom_pages_dir));
+    if !custom_pages_dir.is_dir() {
+        res?;
+        return Err(anyhow!("Fail to create dir {:?}", custom_pages_dir));
+    }
```
niklasmohrin, line 254: See https://github.com/tealdeer-rs/tealdeer/pull/388#discussion_r1921624380

```suggestion
    create_dir_all(custom_pages_dir).context("Failed to create custom pages directory")?;
```

niklasmohrin (changes requested):

lengyijun:
We still can't deal with permission problem

niklasmohrin:
> We still can't deal with permission problem: if the config dir is created by sudo

Aaaah, now I finally get what you mean. It was very hard to understand what scenario exactly you meant.

Okay, so like I suggested the time before, if we want that tealdeer can handle this situation, we should have a test that ensures this feature still works. If I understand correctly, we wouldn't need to have the directory created by root, but using `chmod 0 ~/.local/share/tealdeer/pages/` should have the same effect.

However, I also again think that once we have this test, we will see that using the code I suggest in https://github.com/tealdeer-rs/tealdeer/pull/388#discussion_r1938000089 should still work. I tested with the change applied:
```shell
$ rm -rf ~/.local/share/tealdeer/
$ cargo run -- --edit-page foo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `/home/niklas/.cache/cargo-target-dir/debug/tldr --edit-page foo`
Editing "/home/niklas/.local/share/tealdeer/pages/foo.page.md"
$ rm -rf ~/.local/share/tealdeer/pages/
$ sudo mkdir ~/.local/share/tealdeer/pages/
$ cargo run -- --edit-page foo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `/home/niklas/.cache/cargo-target-dir/debug/tldr --edit-page foo`
Editing "/home/niklas/.local/share/tealdeer/pages/foo.page.md"
```
From how I understand your comment, you would expect that in the last run, `create_dir_all` would fail, right?

niklasmohrin:
Alright, given that this PR is, in my opinion, ready for release, and I want to avoid merge conflicts on this branch, let's proceed as follows: I will commit my suggested change and merge the PR. If we then want to add another test case and handle any permission situation more carefully, let's do it in a follow up PR

niklasmohrin (approved):
Very cool feature! I am looking forward to discussing how we can make the permission situation better :)

PR #389, opened 2024-11-05
Upgrade yansi: 0.5.1 -> 1.0.1

nc7s:
Release note: https://github.com/SergioBenitez/yansi/releases/tag/v1.0.0

It'd be nice to have a version bump for this (patch level suffices), so I can straight update it in Debian rather than having to carry a patch ;)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -251,7 +251,10 @@ fn main() {
     // Determine the usage of styles
     let enable_styles = match args.color.unwrap_or_default() {
         // Attempt to use styling if instructed
-        ColorOptions::Always => true,
+        ColorOptions::Always => {
+            yansi::enable(); // disable yansi's automatic detection for ANSI support on Windows
+            true
+        }
         // Enable styling if:
         // * NO_COLOR env var isn't set: https://no-color.org/
         // * The output stream is stdout (not being piped)
--- a/src/main.rs
+++ b/src/main.rs
@@ -249,20 +249,13 @@ fn main() {
     let args = Cli::parse();
 
     // Determine the usage of styles
-    #[cfg(target_os = "windows")]
-    let ansi_support = yansi::Paint::enable_windows_ascii();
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -249,20 +249,13 @@ fn main() {
     let args = Cli::parse();
 
     // Determine the usage of styles
-    #[cfg(target_os = "windows")]
-    let ansi_support = yansi::Paint::enable_windows_ascii();
```
niklasmohrin, line 253: I suppose by https://docs.rs/yansi/1.0.1/yansi/index.html#windows we can omit this. Fair enough

niklasmohrin (approved):

niklasmohrin:
Cool! Do you know of any other dependencies updates we can do that would declutter the package index only requiring a patch-version bump?

src/main.rs
nc7s: Yeah the commit message explained ;)

nc7s:
Not really, as least in Debian. Thanks for asking ;)

src/main.rs
niklasmohrin: Ah, I didn't see that, I'll make sure to include it in the merged commit

src/main.rs
```diff
@@ -249,20 +249,16 @@ fn main() {
...
-    #[cfg(not(target_os = "windows"))]
-    let ansi_support = true;
     let enable_styles = match args.color.unwrap_or_default() {
         // Attempt to use styling if instructed
-        ColorOptions::Always => true,
+        ColorOptions::Always => {
+            yansi::enable(); // disable yansi's automatic detection for ANSI support on Windows
```
niklasmohrin, line 255: @nc7s Sorry to bother you again, but how do you feel about this change I added? It should now be exactly the previous behavior right?

nc7s: I don't really know enough to judge this, from a glance LGTM?

Issue #306, opened 2022-12-23
Allow overriding via env vars custom pages/patches (just like config/cache)


eugenesvk:
Currently you can override via Environment Variables
  - [Override Config Directory](https://dbrgn.github.io/tealdeer/config.html#override-config-directory) and 
  - [Override Cache Directory](https://dbrgn.github.io/tealdeer/config.html#override-cache-directory)
Similarly, I'd like to use env vars to override:
  - [Custom Pages](https://dbrgn.github.io/tealdeer/usage_custom_pages.html#custom-pages)
  - [Custom Patches](https://dbrgn.github.io/tealdeer/usage_custom_pages.html#custom-patches)

To allow using the same config file on various machines/users, since config only accepts absolute paths while it's easier to set relative paths in a shell config

dbrgn:
Hmm, when setting relative paths, what would they be relative to?

Actually the `TEALDEER_CACHE_DIR` is something that is a leftover from when IIRC we didn't even have a config file. Since the config now contains a [`directories` section](https://dbrgn.github.io/tealdeer/config_directories.html), that's the place where I'd like to put all configuration values going forward. Adding additional env variables just makes things more complex. (I'd even like to deprecate the `TEALDEER_CACHE_DIR` variable and recommend using the config file instead, so that `TEALDEER_CONFIG_DIR` is the only env variable we need to check. What do you think, @niklasmohrin? We could still keep it working for quite some time in the code, but update docs.)

If absolute paths are a problem, then maybe that's where improvements could be made. Maybe supporting `~` expansion? (Not sure how that feels on Windows though, I think on Windows we'd need to support `%HOMEPATH%` instead 😕)

eugenesvk:
> Hmm, when setting relative paths, what would they be relative to?

Relative to the global config path variable I've set up

> If absolute paths are a problem, then maybe that's where improvements could be made

Sure, and the least complicated way to improve it is so support env vars, where all of the issues like `~` and `%HOMEPATH` are already resolved in your shell config and only require a simple 
`$TEALDEER_CONFIG_DIR = f"{cfgH}/tldr"`
instead of forcing to use your resolver

eugenesvk:
> cannot have all settings accessible only through env vars 

That's fine because env vars don't **replace** any configuration files, they only allow **overriding** it, so if users don't want to "clutter" their envs, then they can just... not set any env vars and use the configs

[...]

PR #395, opened 2024-12-08
Resolve paths in config [directories] relative to the config directory (#306)

MatejKafka:
See #306. The joined path is not cleaned in any way, so the resulting paths will not be very pretty, but since Rust stdlib afaik does not give us anything built-in for cleaning paths without touching the filesystem, I leave the paths as they are.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -295,7 +295,7 @@ impl Config {
     ///
     /// For this, some values need to be converted to other types and some
     /// defaults need to be set (sometimes based on env variables).
-    fn from_raw(raw_config: RawConfig) -> Result<Self> {
+    fn from_raw(raw_config: RawConfig, relative_path_root: &Path) -> Result<Self> {
         let style = raw_config.style.into();
         let display = raw_config.display.into();
         let updates = raw_config.updates.into();
@@ -316,7 +316,9 @@ impl Config {
         } else if let Some(config_value) = raw_config.directories.cache_dir {
             // If the user explicitly configured a cache directory, use that.
             PathWithSource {
-                path: config_value,
+                // Resolve possible relative path. It would be nicer to clean up the path, but Rust stdlib
+                // does not give any method for that that does not need the paths to exist.
+                path: relative_path_root.join(config_value),
                 source: PathSource::ConfigFile,
             }
         } else if let Ok(default_dir) = get_app_root(AppDataType::UserCache, &crate::APP_INFO) {
@@ -333,7 +335,8 @@ impl Config {
             .directories
             .custom_pages_dir
             .map(|path| PathWithSource {
-                path,
+                // Resolve possible relative path.
+                path: relative_path_root.join(path),
                 source: PathSource::ConfigFile,
             })
             .or_else(|| {
@@ -382,8 +385,12 @@ impl Config {
             RawConfig::new()
         };
 
-        // Convert to config
-        let mut config = Self::from_raw(raw_config).context("Could not process raw config")?;
+        // Safe to unwrap, it's a file path, so it should have a directory component
+        let config_file_dir = config_file_path.parent().unwrap();
+
+        // Convert to config, resolve relative paths from the config file dir
+        let mut config =
+            Self::from_raw(raw_config, config_file_dir).context("Could not process raw config")?;
 
         // Potentially override styles
         if !enable_styles {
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -474,3 +480,22 @@ fn test_serialize_deserialize() {
...
+    let config = Config::from_raw(config_dir.join("config.toml"), raw_config).unwrap();
+
+    assert_eq!(
+        config.directories.cache_dir.path(),
+        config_dir.join("../cache")
+    );
+    assert_eq!(
+        config.directories.custom_pages_dir.unwrap().path(),
+        config_dir.join("../custom_pages")
+    );
```
niklasmohrin, line 500: I think we should replace the `config_dir.join("../something")` lines here with the absolute path we expect, for example `/path/to/cache` (or `/path/to/config/../cache`).

src/config.rs
```diff
@@ -295,7 +295,9 @@ impl Config {
     ///
     /// For this, some values need to be converted to other types and some
     /// defaults need to be set (sometimes based on env variables).
-    fn from_raw(raw_config: RawConfig) -> Result<Self> {
+    fn from_raw(config_file_path: PathBuf, raw_config: RawConfig) -> Result<Self> {
+        // Safe to unwrap since we managed to read the file
+        let config_dir_path = config_file_path.parent().unwrap();
```
niklasmohrin, line 300: I think this should be
```suggestion
    fn from_raw(raw_config: RawConfig, relative_path_root: &Path) -> Result<Self> {
```
where `relative_path_root` replaces `config_dir_path` - the caller of `from_raw` should perform the `config_file_path.parent()` call, because they are the ones who know that they got the `RawConfig` from this path and thus it exists. We here shouldn't really know how `config_file_path` and `raw_config` are related

This also eliminates the need for the added comments in this function, in my opinion

niklasmohrin (commented):
Thank you for your submission, very cool!

I wonder whether having tealdeer print "/path/to/config/../cache" is actually a feature, because it makes it clear how the full path came to be - then again, the user wrote `../` in their config, so they probably have an idea about what they were doing

src/config.rs
MatejKafka: Iirc that doesn't work because the path will contain `\` on Windows. We'd first need to unify the separators, which seems more fragile than doing this.

src/config.rs
MatejKafka: Fair point, but if we pass the base dir and not config path, someone reading the code might start wondering "Hmm, and where are the paths resolved from? Is it always relative to the config file? Do I need to check for that?". If we pass the config file path, it's more obvious what the function does, imo.

src/config.rs
niklasmohrin: Hm, I would think that as long as we compare `Path` objects we don't care about the separator.

In fact, I just found https://doc.rust-lang.org/std/fs/fn.canonicalize.html which we could use on the result before comparing the paths

src/config.rs
niklasmohrin: Hm, I see, this makes `from_raw` more abstract than it is right now. I think that this abstraction is correct though, because we already moved "reading the raw config from the file" to another place, so we would reintroduce this knowledge here. This was what tripped me up, because the comment also says "we know we read the file, so we can unwrap", but in the context of just this method, we don't actually know that, we are just lifting the raw config to a proper Config object. Additionally, there isn't really a reason to limit what the path argument should be - we only use it as a root for the relative paths

What do you think?

src/config.rs
MatejKafka: `canonicalize` won't help us, because it needs an existing path, while the tested paths do not exist on disk.

MatejKafka: Good point with path comparison, did not realize that Rust correctly ignores separators. Updated to compare to the full path.

src/config.rs
MatejKafka: Hmm, thinking about it a bit more, the comment should be more like "if unwrap fails, it's the caller's fault, since they should pass a file path".

Your point about abstraction makes sense – if someone wanted to e.g. support reading config from stdin and resolve relative to the working directory, we'd want to pass in a directory, and technically, relative paths are always resolved relative to a directory, not to a file (although it's a little bit confusing for the config file scenario). Do not have a strong preference, so let's go with your suggestion.

MatejKafka: Updated.

src/config.rs
```diff
@@ -382,8 +385,12 @@ impl Config {
             RawConfig::new()
         };
 
-        // Convert to config
-        let mut config = Self::from_raw(raw_config).context("Could not process raw config")?;
+        // Safe to unwrap since we managed to read the file
+        let config_file_dir = config_file_path.parent().unwrap();
```
niklasmohrin, line 389: The comment is still a bit off here, we are not always reading the file. The proper argument is that we know that this path points to a file, which necessarily has a directory, as you pointed out in the other thread. This all builds on the implementation detail of `get_config_path`. If we would inline `get_config_path`, then we would not need the `.parent` here at all, so maybe we should just do that? I suppose we could get rid of `get_config_path` altogether and also replace the second occurrence with a combination of `get_config_dir` and `CONFIG_FILE_NAME`.

niklasmohrin (commented):
Great, I have one last thing before we can merge :)

src/config.rs
MatejKafka: Personally would not inline the method, since that gives you the freedom to e.g. move the config file into a subdirectory. Agree that reconstructing the parent is not very pretty, imo it's more explicit than inlining the method ("we're resolving paths relative to the parent directory of the config file, whichever it is").

MatejKafka: Updated the comment.

niklasmohrin (approved):
Thanks for contributing and discussion! :)

PR #396, opened 2024-12-18
Add `--no-pager` flag

RomanPro100:
In my `tealdeer` config I have `use_pager` set to `true`, but if there's a case I want it not to be paged, there is no easy way to do this. The only way I found is to change the `use_pager` option in the config to `false`.
So I decided it would be nice to have a `--no-pager` flag that disables paging whether it's enabled in the config or by the `--pager` flag.
I also added some tests and edited the shell completion scripts, I hope I made everything right.

### DIFF
--- a/completion/bash_tealdeer
+++ b/completion/bash_tealdeer
@@ -6,7 +6,7 @@ _tealdeer()
 	_init_completion || return
 
 	case $prev in
-		-h|--help|-v|--version|-l|--list|-u|--update|--no-auto-update|-c|--clear-cache|--pager|-r|--raw|--show-paths|--seed-config|-q|--quiet)
+		-h|--help|-v|--version|-l|--list|-u|--update|--no-auto-update|-c|--clear-cache|--pager|--no-pager|-r|--raw|--show-paths|--seed-config|-q|--quiet)
 			return
 			;;
 		-f|--render)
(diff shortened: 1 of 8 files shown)


niklasmohrin:
Hey @RomanPro100,

this is definitely a good motivation to implement this flag - thank you for submitting the PR! 

I would have thought that clap would be able to generate the `--no-` version of a flag on its own, but https://github.com/clap-rs/clap/issues/815 is still open. For the time being, I think we should include some more tweaks to the interaction of `--[no-]pager`, in particular `tldr --pager --no-pager` and `tldr --no-pager --pager` should do what the latest argument says. I believe this is done using clap's `overrides_with` attribute.

A good example on how to do this cleanly is `uv`:
- https://github.com/astral-sh/uv/blob/44664022148e92c00a0b534c6d17aedc41ad0488/crates/uv/src/settings.rs#L190
- https://github.com/astral-sh/uv/blob/dcd96a83aa5254d6b5068d2c68b074a0376d4100/crates/uv-cli/src/lib.rs#L2525-L2542

In our case, we use `flag(args.pager, args.no_pager).unwrap_or(config.display.pager)`, so that the command line parameters always take precedence. All of this "deciding whether or not we spawn a pager" business can be done in `main` instead of in `print_page`, because it really is more CLI parsing logic than printing logic.

The tests should cover these cases, i.e. test that CLI takes precedence and that using both `--pager` and `--no-pager` together still works as expected. This should optimally include an actual check on whether or not the pager was spawned, for example by using `tee tealdeer-pager-log` as the pager and then checking for the presence of the `tealdeer-pager-log` file (we can set the environment variable `PAGER` when we run the tests).

---

All of these tasks might be a bit overwhelming, in particular the testing part, so if you have any troubles or questions please reach out :)

niklasmohrin:
@RomanPro100 do you want to continue working on this, or should we close the PR for now?

RomanPro100:
Nah you can close this

PR #400, opened 2025-01-07
Move most of `main` to `try_main`

niklasmohrin:
Currently, `main` uses `print_error` and `process::exit` in several places. These could be unified by propagating errors via `Result`s. There is one place left that manually calls `print_error`, because the error should be followed by help instructions printed without any highlighting.


beatbrot (approved):
Nice improvement! Should make testability easier and centralized error handling

niklasmohrin:
For reference, I am waiting for #388 to be merged before merging this, as the conflicts will be easier to fix here

PR #401, opened 2025-01-08
Add `common` platform to CLI

### LINKED ISSUE #323: Allow passing `common` as platform

beatbrot:
Fixes #323

### DIFF
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -9,9 +9,9 @@ Arguments:
 Options:
   -l, --list                 List all commands in the cache
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference
-                             [possible values: linux, macos, sunos, windows, android, freebsd, netbsd, openbsd,
-                             common]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
+                             of preference [possible values: linux, macos, sunos, windows, android,
+                             freebsd, netbsd, openbsd, common]
   -L, --language <LANGUAGE>  Override the language
   -u, --update               Update the local cache
       --no-auto-update       If auto update is configured, disable it for this run
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -9,9 +9,9 @@ Arguments:
 Options:
   -l, --list                 List all commands in the cache
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
-                             of preference [possible values: linux, macos, sunos, windows, android,
-                             freebsd, netbsd, openbsd]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference
+                             [possible values: linux, macos, sunos, windows, android, freebsd, netbsd, openbsd,
+                             common]
   -L, --language <LANGUAGE>  Override the language
   -u, --update               Update the local cache
       --no-auto-update       If auto update is configured, disable it for this run
--- a/src/main.rs
+++ b/src/main.rs
@@ -282,7 +282,7 @@ fn main() {
         create_config_and_exit(enable_styles);
     }
 
-    let fallback_platforms: &[PlatformType] = &[PlatformType::current()];
+    let fallback_platforms: &[PlatformType] = &[PlatformType::current(), PlatformType::Common];
     let platforms = args
         .platforms
         .as_ref()
--- a/src/main.rs
+++ b/src/main.rs
@@ -282,11 +282,7 @@ fn main() {
         create_config_and_exit(enable_styles);
     }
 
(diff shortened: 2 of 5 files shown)


niklasmohrin:
Thanks for the PR! As "common" is the fallback choice for other platforms, can you change it so that we don't have that behavior hardcoded in the code of `Cache` and instead include the new `Platform::Common` in the list of searched platforms in `main.rs`?

beatbrot:
@niklasmohrin Done :)

src/main.rs
```diff
@@ -282,7 +282,7 @@ fn main() {
         create_config_and_exit(enable_styles);
     }
 
-    let fallback_platforms: &[PlatformType] = &[PlatformType::current()];
+    let fallback_platforms: &[PlatformType] = &[PlatformType::current(), PlatformType::Common];
     let platforms = args
```
niklasmohrin, line 286: We still have an occurrence of `"common"` when listing all available pages which should be removed. Once this is done, one of our tests will fail that asserts that using `tldr --platform linux` actually checks both "linux" and "common" (for printing pages, as well as for listing them). So we should make sure to append "common" to the list of considered platforms here also in the case that `args.platforms` is `Some(_)`. It's fine if we clone `args.platforms` for this. 

(the comments for listing pages also include mentions of the "common" platform, which should then be removed)

docs/src/usage.txt
```diff
@@ -9,9 +9,9 @@ Arguments:
 Options:
   -l, --list                 List all commands in the cache
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
-                             of preference [possible values: linux, macos, sunos, windows, android,
-                             freebsd, netbsd, openbsd]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference
+                             [possible values: linux, macos, sunos, windows, android, freebsd, netbsd, openbsd,
+                             common]
```
niklasmohrin, line 14: CI fails here because it executes `tldr --help` on a terminal with a different width. This is kind of annoying, and I don't know if one can easily emulate this programmatically. To fix CI, you can just manually do the adjustments that are suggested in the CI run

niklasmohrin (changes requested):
As Rust 1.84 was released today, the CI now also has a failure with clippy on code that is unrelated to your PR. Feel free to ignore the failure for that check

src/main.rs
beatbrot: Fair point :) I hope I found all special handlings now!

docs/src/usage.txt
niklasmohrin: Worked!

src/main.rs
niklasmohrin: The behavior for `--list` is correct now :+1: I had another implementation in mind though, sorry for not making this clear immediately!

I think that now that we have the `Common` platform, we can move the "we will also check the common directory for pages" logic out of `Cache` entirely - this is nice, because the cache shouldn't care about _the choice_ of platforms, only about how to find pages for the given name, platform, etc.

Thus, no method of `Cache` should make an exception for "common", it should just be treated like any other platform. The fact that we are still checking the "common" platform is business logic, it fits a lot better in `main`. I am imagining to replace the code around line 285 with something like
```rust
let fallback_platforms = ...;
let chosen_platforms = args.platforms.clone();
let platforms = chosen_platforms.something();
```
such that in the end `platforms` is either a referencing `fallback_platforms` or `chosen_platforms` and `chosen_platforms` has `Common` appended, if it was not already present.

---

This would also change the behavior for `tldr --platform linux tar` - right now, this would check only "linux", but the behavior of tealdeer 1.7 is to check both "linux" and "common". It seems that we are missing a test case for this, but it is definitely the expected behavior for the current version. If you are eager, you could also add a testcase for this in `lib/tests.rs` around the other tests checking the `--platform` lookup behavior (like `test_multiple_platform_command_search_not_found` and `test_multiple_platform_command_search`).

Sorry that I am blowing this PR up so much, if you want to hand it off at any point please let me know. Otherwise, I am very happy to keep working on this together! :)

beatbrot: Nah its fine as long as you can live with a few days interruption here and there :) 

Thank you very much for your thorough review! I hope I got it this time :)

niklasmohrin (approved):
Everything looks good to me, thanks!

src/main.rs
niklasmohrin: Oh yes, I might also not be able to do thorough reviews for longer times if other work is piling up, so no worries

PR #407, opened 2025-01-11
Add tests for osx/macos alias

### LINKED ISSUE #406: Test `macos` alias behavior
In #215, we added support for the "macos" platform which works as an alias for "osx". This was later included in [version 2.0 of the client specification](https://github.com/tldr-pages/tldr/blob/v2.0/CLIENT-SPECIFICATION.md). However, we don't really have tests for this, as "osx" does not appear in our test suite at all and "macos" only appears once in another context.

Tests should be added that ensure that both "macos" and "osx" map to "osx" directory in the page cache. To do so, a test in `tests/lib.rs` should create a cache with a page for the "osx" platform and check that it is found in a normal `tldr` invocation and in `tldr --list`.

beatbrot:
Fixes #406

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -653,6 +653,35 @@ fn test_multiple_platform_command_search_not_found() {
 }
 
 #[test]
+fn test_osx_is_alias_for_macos() {
+    let testenv = TestEnv::new();
+    testenv.add_os_entry("osx", "maconly", "this command only exists on mac");
+
+    testenv
+        .command()
+        .args(["--platform", "macos", "maconly"])
+        .assert()
+        .success();
+    testenv
+        .command()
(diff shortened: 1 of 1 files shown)


tests/lib.rs
```diff
@@ -652,6 +652,35 @@ fn test_multiple_platform_command_search_not_found() {
         .stderr(contains("Page `windows-only` not found in cache."));
 }
 
+#[test]
+fn test_osx_is_alias_for_macos() {
```
niklasmohrin, line 656: I think we usually phrase it the other way around:
```suggestion
fn test_macos_is_alias_for_osx() {
```

niklasmohrin (approved):
Thanks!

Issue #325, opened 2023-05-17
Create an auto man page


tjapro:
I know this seems to be irrelevant and/or redundant, but sometimes we need more info about tldr. In this case, would be better to have more information in man tldr or man tealdeer.
Currently, there is no man page.

In the man page could be all auto-parsed documentation from [github pages](https://dbrgn.github.io/tealdeer/).

Any thoughts?

> Posted originally in https://github.com/tldr-pages/tldr/discussions/10214.  
> Related: https://github.com/dbrgn/tealdeer/issues/218

niklasmohrin:
Sounds like a cool idea, but I don't think it will be very easy. For starters, not all chapters from the online documentation would need to be included, for example "installation" or "usage" (instead of the current "usage", a more thorough explanation could be used or maybe just "examples"). Then, we need to figure out how the generation would actually work - maybe mdbook can do it on its own already? So, if anyone comes up with something working that is not too complicated, I would not oppose including it :)

PR #410, opened 2025-01-31
chore: Updated banchmark README.md

Walker-00:
Hey there!, I'm back.  

Here’s what I’ve updated in this PR:  

1. **Fixed the `tldr-bash` link:**  
   The old link (`https://4e4.win/tldr`) wasn’t working for me, so I updated it to point to the GitHub repo: [`https://github.com/raylee/tldr-sh-client`](https://github.com/raylee/tldr-sh-client). This should make things clearer and easier for anyone trying to find the right repo.  

2. **Updated the benchmark section in the README:**  
   The current benchmarks were last updated in 2021, and since it’s now 2025, I thought it was time for a refresh! I ran new benchmarks (50 runs on 31.01.2025) and replaced the old data with the latest results.  

3. **Added a note about `outfieldr`:**  
   I couldn’t get `outfieldr` to compile on my Arch Linux machine unless I downgraded `clang`, `llvm`, `zig`, etc., which feels like a lot of hassle. I added a note in the README about its lack of maintenance so users know what to expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,23 +45,24 @@ measures the execution times on a cold disk cache. The benchmarking is run in a
 Docker container using sharkdp's [`hyperfine`][hyperfine-gh]
 ([Dockerfile][benchmark-dockerfile]).
 
-| Client (50 runs, 17.10.2021)      | Programming Language | Mean in ms | Deviation in ms | Comments                |
-| :---:                             | :---:                | :---:      | :---:           | :---:                   |
-| [`outfieldr`][outfieldr-gh]       | Zig                  | 9.1        | 0.5             | no user configuration   |
-| `tealdeer`                        | Rust                 | 13.2       | 0.5             |                         |
-| [`fast-tldr`][fast-tldr-gh]       | Haskell              | 17.0       | 0.6             | no example highlighting |
-| [`tldr-hs`][hs-gh]                | Haskell              | 25.1       | 0.5             | no example highlighting |
-| [`tldr-bash`][bash-gh]            | Bash                 | 30.0       | 0.8             |                         |
-| [`tldr-c`][c-gh]                  | C                    | 38.4       | 1.0             |                         |
-| [`tldr-python-client`][python-gh] | Python               | 87.0       | 2.4             |                         |
-| [`tldr-node-client`][node-gh]     | JavaScript / NodeJS  | 407.1      | 12.9            |                         |
+| Client (50 runs, 31.01.2025)      | Programming Language | Mean in ms | Deviation in ms | Comments                         |
+| :---:                             | :---:                | :---:      | :---:           | :---:                            |
+| [`outfieldr`][outfieldr-gh]       | Zig                  | ???        | ???             | lacks maintenance and features   |
+| `tealdeer`                        | Rust                 | 2.0        | 0.4             |                                  |
+| [`fast-tldr`][fast-tldr-gh]       | Haskell              | 2.9        | 0.4             | no example highlighting          |
+| [`tldr-c`][c-gh]                  | C                    | 4.9        | 0.6             |                                  |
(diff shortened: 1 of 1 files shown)


Walker-00:
Also added a man page for https://github.com/tealdeer-rs/tealdeer/issues/325

README.md
```diff
@@ -45,23 +45,24 @@ measures the execution times on a cold disk cache. The benchmarking is run in a
...
+| :---:                             | :---:                | :---:      | :---:           | :---:                            |
+| [`outfieldr`][outfieldr-gh]       | Zig                  | ???        | ???             | lacks maintenance and features   |
+| `tealdeer`                        | Rust                 | 2.0        | 0.4             |                                  |
+| [`fast-tldr`][fast-tldr-gh]       | Haskell              | 2.9        | 0.4             | no example highlighting          |
+| [`tldr-c`][c-gh]                  | C                    | 4.9        | 0.6             |                                  |
+| [`tldr-hs`][hs-gh]                | Haskell              | 12.4       | 0.2             | no example highlighting          |
+| [`tldr-python-client`][python-gh] | Python               | 59.3       | 2.5             |                                  |
+| [`tldr-bash`][bash-gh]            | Bash                 | 100.5      | 7.1             |                                  |
+| [`tldr-node-client`][node-gh]     | JavaScript / NodeJS  | 282.8      | 10.7            |                                  |
```
niklasmohrin, line 57: How did you generate these numbers? The main reason we haven't updated the benchmarks already is that no one felt like updating the `Dockerfile` mentioned in the paragraph above the table. As described in the README, the benchmark should be easily reproducible, otherwise the numbers have little meaning

docs/tealdeer.1
```diff
@@ -0,0 +1,132 @@
+TEALDEER(1)                     User Commands                     TEALDEER(1)
```
niklasmohrin, line 1: In #325 we concluded that we want to have a man page that is _automatically generated_ from the documentation. This looks like it is created by hand, which I would prefer to not add, as it adds more places that need to be updated when, for example, command line parameters change

README.md
```diff
@@ -116,7 +117,7 @@ Thanks to @severen for coming up with the name "tealdeer"!
 [c-gh]: https://github.com/tldr-pages/tldr-c-client
 [hs-gh]: https://github.com/psibi/tldr-hs
 [fast-tldr-gh]: https://github.com/gutjuri/fast-tldr
-[bash-gh]: https://4e4.win/tldr
+[bash-gh]: https://github.com/raylee/tldr-sh-client
```
niklasmohrin, line 120: For context, it seems like the URL we used previously was changed some time ago: https://github.com/pepa65/tldr-bash-client/commit/6321f7c6df35aa90f3bfb03477f64206cd824d83

However, the client there seems to be unmaintained anyways, so using the one from `raylee` seems sensible. Still, it would also have to be changed in the `Dockerfile`

niklasmohrin (changes requested):
Hey @Walker-00,

in contrast to #357, this PR only updates the table of benchmark results, while leaving the benchmark itself untouched. However, I would not want to put performance numbers into our README unless they are backed by a reproducible benchmark setup.

In addition to the comment regarding the man page, I would also prefer if you could work on these two concerns in two separate PRs, so that the discussions don't overlap.

README.md
Walker-00: I didn't run `Dockerfile`, I manually downloaded them and ran them on my machine. I will fix the Docker and rewrite the benchmarks.

docs/tealdeer.1
Walker-00: You mean like auto generated from the cargo-doc?

niklasmohrin: No, I meant automatically generated from the `docs/` directory of the repository. Please see #325

niklasmohrin:
@Walker-00 do you want to continue working on this?

PR #417, opened 2025-03-01
Replace reqwest with ureq

### LINKED ISSUE #415: Consider replacing `reqwest` with `ureq`
I would hope that this would decrease binary size, as `reqwest` pulls in `tokio`. In contrast, `ureq` is designed to be blocking.

erickguan:
ureq is a blocking HTTP client. ureq is simpler than reqwest. This brings:
- smaller binary size
- cleaner configuration interfaces

Fixes #415

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,9 +25,9 @@ app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
 env_logger = { version = "0.11", optional = true }
 log = "0.4"
-reqwest = { version = "0.12.9", features = ["blocking"], default-features = false }
 serde = "1.0.21"
 serde_derive = "1.0.21"
+ureq = { version = "3.0.8", default-features = false, features = ["gzip"] }
 toml = "0.8.19"
 walkdir = "2.0.1"
 yansi = "1"
@@ -47,20 +47,20 @@ filetime = "0.2.10"
 default = ["native-tls", "rustls-with-webpki-roots", "rustls-with-native-roots"]
 logging = ["env_logger"]
 
-# Reqwest (the HTTP client library) can handle TLS connections in four
+# ureq (the HTTP client library) can handle TLS connections in four
 # different modes:
 #
 # - Rustls with:
 #   - native roots
-#   - WebPK roots
-# - Native TLS (SChannel on Windows, Secure Transport on macOS and OpenSSL otherwise) with:
+#   - WebPKI roots
+# - Native TLS (by `native_tls` crate) with:
 #   - native roots
-#   - WebPK roots (not implemented in tealdeer)
+#   - WebPKI roots (not implemented in tealdeer)
 #
-# At least one of variants must be selected. By default, uses native TLS and native roots.
-native-tls = ["reqwest/rustls-tls-native-roots-no-provider", "reqwest/rustls-tls-webpki-roots-no-provider", "reqwest/native-tls"]
-rustls-with-webpki-roots = ["reqwest/rustls-tls-native-roots-no-provider", "reqwest/rustls-tls-webpki-roots"]
-rustls-with-native-roots = ["reqwest/rustls-tls-webpki-roots-no-provider", "reqwest/rustls-tls-native-roots"]
+# At least one of variants must be selected.
+native-tls = ["ureq/native-tls", "ureq/platform-verifier"]
+rustls-with-webpki-roots = ["ureq/rustls"] # ureq uses WebPKI roots by default
+rustls-with-native-roots = ["ureq/rustls", "ureq/platform-verifier"]
 
 ignore-online-tests = []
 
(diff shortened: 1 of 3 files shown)


src/cache.rs
```diff
@@ -475,6 +432,60 @@ impl Cache {
...
 
+fn build_proxy() -> Result<Option<Proxy>> {
+    if let Ok(ref host) = env::var("HTTP_PROXY") {
+        return Ok(Some(Proxy::new(host)?));
+    }
+    if let Ok(ref host) = env::var("HTTPS_PROXY") {
+        return Ok(Some(Proxy::new(host)?));
+    }
+
+    Ok(None)
```
niklasmohrin, line 443: This is different from what we did before. The code before uses `HTTP_PROXY` as a `Proxy::http` and `HTTPS_PROXY` as a `Proxy::https`, which means that they will be used for http and https traffic respectively (see https://docs.rs/reqwest/latest/reqwest/struct.Proxy.html). Now, this uses `HTTP_PROXY` also for https traffic if present and ignores `HTTP_PROXY`.

If we want to be conservative, we could _only_ check `HTTPS_PROXY`. Alternatively, there is https://docs.rs/ureq/latest/ureq/struct.Proxy.html#method.try_from_env, which would prefer `HTTPS_PROXY`, but still use `HTTP_PROXY` if `HTTPS_PROXY` is unset. I am not sure which approach is best

src/cache.rs
```diff
@@ -475,6 +432,60 @@ impl Cache {
...
+}
+
+impl Cache {
+    fn build_client(tls_backend: TlsBackend) -> Result<Agent> {
+        let proxy = build_proxy()?;
+        let mut tls_builder = TlsConfig::builder();
+        tls_builder = match tls_backend {
```
niklasmohrin, line 450: Can't we omit the first line and use the following?
```rust
let tls_builder = match tls_backend {    
    TlsBackend::NativeTls => TlsConfig::builder()
        .provider(...)
    // ...
};
```

src/cache.rs
```diff
@@ -475,6 +432,60 @@ impl Cache {
...
+            TlsBackend::RustlsWithWebpkiRoots => tls_builder
+                .provider(TlsProvider::Rustls)
+                .root_certs(RootCerts::WebPki),
+            #[cfg(feature = "rustls-with-native-roots")]
+            TlsBackend::RustlsWithNativeRoots => tls_builder
+                .provider(TlsProvider::Rustls)
+                .root_certs(RootCerts::PlatformVerifier),
+        };
+        let config = Agent::config_builder()
+            .user_agent(USER_AGENT)
```
niklasmohrin, line 465: Is there a reason to not use the default user agent?

src/cache.rs
```diff
@@ -475,6 +432,60 @@ impl Cache {
...
+            .timeout_global(Some(Duration::from_secs(GLOBAL_DOWNLOAD_TIMEOUT_SECONDS)))
```
niklasmohrin, line 466: Same thing, why is this better than the default? I would expect that it just waits until the download is complete, which seems preferable for users with a very slow Internet connection

src/cache.rs
```diff
@@ -475,6 +432,60 @@ impl Cache {
...
+    /// Download the archive from the specified URL.
+    fn download(client: &Agent, archive_url: &str) -> Result<Vec<u8>> {
+        let mut response = client
+            .get(archive_url)
+            .call()
+            .with_context(|| format!("Could not download tldr pages from {archive_url}"))?;
+        let buf = response
+            .body_mut()
+            .with_config() // to use unlimited (default) output buffer
+            .read_to_vec()?;
```
niklasmohrin, line 483: Shouldn't `response.into_body().into_reader().read_to_end(&mut buf)` also work?

niklasmohrin (commented):
Nice!

Cargo.toml
```diff
@@ -47,20 +47,20 @@ filetime = "0.2.10"
...
 logging = ["env_logger"]
 
-# Reqwest (the HTTP client library) can handle TLS connections in four
+# ureq (the HTTP client library) can handle TLS connections in four
 # different modes:
 #
 # - Rustls with:
 #   - native roots
-#   - WebPK roots
-# - Native TLS (SChannel on Windows, Secure Transport on macOS and OpenSSL otherwise) with:
```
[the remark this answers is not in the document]
erickguan: Truly!

src/cache.rs
erickguan: My experience is to set both `HTTP_PROXY`/`HTTPS_PROXY`. Otherwise, I will go with `SOCKS_PROXY`. The default preference looks okay. `HTTP_PROXY` is also not going to affect tldr given the source is usually GitHub. And GitHub uses https.

Let's go with the default which will use `Proxy::try_from_env`.

src/cache.rs
erickguan: Sure.

src/cache.rs
erickguan: Removed to use the default.

src/cache.rs
erickguan: Alright, removed to use the default.

src/cache.rs
erickguan: Indeed! A better choice.

niklasmohrin (approved):
Very cool, thanks!

src/cache.rs
niklasmohrin: Alright. We should add a test at some point, see #418

PR #420, opened 2025-03-23
CI: use ARM Linux runner for tests and releases

theofficialgman:
test release: https://github.com/theofficialgman/tealdeer/actions/runs/14022597794
test CI: https://github.com/theofficialgman/tealdeer/actions/runs/14022594689

you will note that currently all the linux release runners are failing (unrelated to this PR) due to not have openssl in the musl docker container

in the future I hope that the tests could be expanded (to include all the same cross compiles as the release targets) to catch these sort of issues https://github.com/tealdeer-rs/tealdeer/issues/419

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,9 +55,27 @@ jobs:
         include:
           - arch: "x86_64"
             libc: "musl"
-          - arch: "aarch64"
-            libc: "musl"
           - arch: "i686"
+            libc: "musl"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Pull Docker image
+        run: docker pull messense/rust-musl-cross:${{ matrix.arch }}-${{ matrix.libc }}
+      - name: Build in Docker
+        run: docker run --rm -i -v "$(pwd)":/home/rust/src messense/rust-musl-cross:${{ matrix.arch }}-${{ matrix.libc }} cargo build --release
+      - name: Strip binary
+        run: docker run --rm -i -v "$(pwd)":/home/rust/src messense/rust-musl-cross:${{ matrix.arch }}-${{ matrix.libc }} musl-strip -s /home/rust/src/target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}/release/tldr
+      - uses: actions/upload-artifact@v4
+        with:
+          name: "tealdeer-linux-${{ matrix.arch }}-${{ matrix.libc }}"
+          path: "target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}/release/tldr"
+
+  build-linux-arm:
+    runs-on: ubuntu-24.04-arm
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -55,10 +55,28 @@ jobs:
         include:
           - arch: "x86_64"
             libc: "musl"
-          - arch: "aarch64"
-            libc: "musl"
           - arch: "i686"
             libc: "musl"
+    steps:
```
niklasmohrin, line 60: To avoid repeating the steps, we can include another value in the matrix list above like
```
- runner: ubuntu-latest
  arch: "x86_64"
  libc: "musl"
```
and then use
```
runs-on: ${{ matrix.runner }}
```

niklasmohrin:
Yes, I also kept #260 open until we build / test for arm. This github image does not cover all the architectures we are targeting, right?

In general, what is the advantage of running the cross compilation for the release on arm? If we are running docker anyways, it shouldn't make a difference right?

theofficialgman:
> Yes, I also kept #260 open until we build / test for arm. This github image does not cover all the architectures we are targeting, right?
> 
> In general, what is the advantage of running the cross compilation for the release on arm? If we are running docker anyways, it shouldn't make a difference right?

The point was to build on a host that can actually run the arm binaries. The GitHub action arm runners can run all of the arm binaries. I figured if the unit test CI was building and then running the arm binaries so should the release CI. Otherwise you might runinto a scenario where you find during the release process that something doesn't work that same as the unit test. That's a reason why I was also going to suggest that unit tests be added for all of the release matrix options. It doesn't really make sense to run glibc targets in the CI unit tests and then only publish musl releases.

niklasmohrin:
> The GitHub action arm runners can run all of the arm binaries.

Does that mean that we can get rid of the musl-cross docker images and just use rustup targets to compile for arm? I would prefer that change very much!

> It doesn't really make sense to run glibc targets in the CI unit tests and then only publish musl releases.

Yes, I don't remember how this came to be. We should change that (not necessarily in this PR).

---

For this PR, I would be fine with the proposed additions, I would just like for the workflow file to stay short / not have the duplication. Hopefully I will get around to look at it more quickly, sorry about that

.github/workflows/release.yml
```diff
@@ -76,7 +94,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: "tealdeer-linux-${{ matrix.arch }}-${{ matrix.libc }}"
-          path: "target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}/release/tldr"
+          path: "target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}/release/tldr"
```
niklasmohrin, line 97: Please also make sure to remove this trailing whitespace :)

niklasmohrin:
@theofficialgman Do you want to continue working on this?

PR #422, opened 2025-05-10
Add an option to specify a custom config file to be used

### LINKED ISSUE #107: Override config file

MHS-0:
Inspired by #141
Closes #107

Note: In the original PR, some modifications were made so that the path specified by the config file flag can be used as the output path of the seed-config option. I chose not to do that here for now, because I think the meaning of this usage wouldn't be visible and clear to the user. Let me know if you think otherwise.

### DIFF
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -11,13 +11,14 @@ Options:
       --edit-page            Edit custom page with `EDITOR`
       --edit-patch           Edit custom patch with `EDITOR`
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference [possible values: linux, macos, sunos, windows, android, freebsd,
-                             netbsd, openbsd, common]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
+                             of preference [possible values: linux, macos, sunos, windows, android,
+                             freebsd, netbsd, openbsd, common]
   -L, --language <LANGUAGE>  Override the language
   -u, --update               Update the local cache
       --no-auto-update       If auto update is configured, disable it for this run
   -c, --clear-cache          Clear the local cache
-      --config <FILE>        Override config file location
+      --config-path <FILE>   Override config file location
       --pager                Use a pager to page output
   -r, --raw                  Display the raw markdown instead of rendering it
   -q, --quiet                Suppress informational messages
--- a/docs/src/usage.txt
+++ b/docs/src/usage.txt
@@ -11,13 +11,13 @@ Options:
       --edit-page            Edit custom page with `EDITOR`
       --edit-patch           Edit custom patch with `EDITOR`
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
-                             of preference [possible values: linux, macos, sunos, windows, android,
-                             freebsd, netbsd, openbsd, common]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference [possible values: linux, macos, sunos, windows, android, freebsd,
+                             netbsd, openbsd, common]
   -L, --language <LANGUAGE>  Override the language
   -u, --update               Update the local cache
       --no-auto-update       If auto update is configured, disable it for this run
   -c, --clear-cache          Clear the local cache
+      --config <FILE>        Override config file location
       --pager                Use a pager to page output
   -r, --raw                  Display the raw markdown instead of rendering it
   -q, --quiet                Suppress informational messages
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -74,8 +74,8 @@ pub(crate) struct Cli {
     #[arg(short = 'c', long = "clear-cache")]
     pub clear_cache: bool,
 
-    /// Use the specified file as config instead of the default
-    #[arg(long = "config")]
+    /// Override config file location
+    #[arg(long = "config", value_name = "FILE")]
     pub config: Option<PathBuf>,
 
     /// Use a pager to page output
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -75,8 +75,8 @@ pub(crate) struct Cli {
     pub clear_cache: bool,
 
     /// Override config file location
-    #[arg(long = "config", value_name = "FILE")]
-    pub config: Option<PathBuf>,
+    #[arg(long = "config-path", value_name = "FILE")]
+    pub config_path: Option<PathBuf>,
 
     /// Use a pager to page output
     #[arg(long = "pager", requires = "command_or_file")]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -74,6 +74,10 @@ pub(crate) struct Cli {
     #[arg(short = 'c', long = "clear-cache")]
     pub clear_cache: bool,
 
+    /// Use the specified file as config instead of the default
+    #[arg(long = "config")]
+    pub config: Option<PathBuf>,
+
     /// Use a pager to page output
     #[arg(long = "pager", requires = "command_or_file")]
     pub pager: bool,
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,7 @@
 use std::{
-    env, fmt, fs,
-    io::Write,
+    env, fmt,
+    fs::{self, File},
+    io::{ErrorKind, Read, Write},
     path::{Path, PathBuf},
     time::Duration,
 };
@@ -254,22 +255,12 @@ impl RawConfig {
         Self::default()
     }
 
-    fn load(config_file_path: &Path) -> Result<RawConfig> {
-        if config_file_path.is_file() {
-            let contents = fs::read_to_string(config_file_path).with_context(|| {
-                format!(
-                    "Failed to read from config file at {}",
-                    config_file_path.display()
-                )
-            })?;
-            return toml::from_str(&contents).with_context(|| {
-                format!(
-                    "Failed to parse TOML config file at {}",
-                    config_file_path.display()
-                )
-            });
-        }
-        bail!("The given path doesn't point to a file")
+    fn load(mut config: impl Read) -> Result<RawConfig> {
+        let mut content = String::new();
+        config
+            .read_to_string(&mut content)
+            .context("Failed to read from config file")?;
+        toml::from_str(&content).context("Failed to parse TOML config file")
     }
 }
 
@@ -469,8 +460,7 @@ impl Config {
     ///
     /// path: The path to the config file.
     pub fn load(path: &Path) -> Result<Self> {
-        // Load raw config, return with error if it fails.
-        let raw_config = RawConfig::load(path)?;
+        let raw_config = RawConfig::load(File::open(path)?)?;
 
         let config = Self::from_raw(
             raw_config,
@@ -491,11 +481,18 @@ impl Config {
         let config_file_path =
             get_default_config_path().context("Could not determine config path")?;
 
-        // Load raw config if possible, or load default
-        let raw_config = if config_file_path.path().is_file() {
-            RawConfig::load(config_file_path.path())?
-        } else {
-            RawConfig::default()
+        let raw_config = match File::open(config_file_path.path()) {
+            Ok(file) => RawConfig::load(file)?,
+            Err(e) => {
+                if e.kind().eq(&ErrorKind::NotFound) {
+                    RawConfig::default()
+                } else {
+                    bail!(
+                        "Failed to open config file at {}",
+                        config_file_path.path().display()
+                    );
+                }
+            }
         };
         let config =
             Self::from_raw(raw_config, config_file_path).context("Could not process raw config")?;
@@ -573,7 +570,7 @@ pub fn make_default_config(path: Option<&Path>) -> Result<PathBuf> {
 
     // Write default config
     let mut config_file =
-        fs::File::create(&config_file_path).context("Could not create config file")?;
+        File::create(&config_file_path).context("Could not create config file")?;
     let _wc = config_file
         .write(serialized_config.as_bytes())
         .context("Could not write to config file")?;
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,6 @@
 use std::{
     env, fmt, fs,
-    io::{Read, Write},
+    io::Write,
     path::{Path, PathBuf},
     time::Duration,
 };
@@ -254,21 +254,22 @@ impl RawConfig {
         Self::default()
     }
 
-    fn from_config_path(config_file_path: &Path) -> Result<Option<RawConfig>> {
-        if config_file_path.exists() && config_file_path.is_file() {
-            let mut config_file = fs::File::open(&config_file_path).with_context(|| {
-                format!("Failed to open config file path at {:?}", &config_file_path)
+    fn load(config_file_path: &Path) -> Result<RawConfig> {
+        if config_file_path.is_file() {
+            let contents = fs::read_to_string(config_file_path).with_context(|| {
+                format!(
+                    "Failed to read from config file at {}",
+                    config_file_path.display()
+                )
             })?;
-            let mut contents = String::new();
-            config_file.read_to_string(&mut contents).with_context(|| {
-                format!("Failed to read from config file at {:?}", &config_file_path)
-            })?;
-            return Ok(Some(toml::from_str(&contents).with_context(|| {
-                format!("Failed to parse TOML config file at {config_file_path:?}")
-            })?));
-        } else {
-            Ok(None)
+            return toml::from_str(&contents).with_context(|| {
+                format!(
+                    "Failed to parse TOML config file at {}",
+                    config_file_path.display()
+                )
+            });
         }
+        bail!("The given path doesn't point to a file")
     }
 }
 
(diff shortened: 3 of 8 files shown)


src/types.rs
```diff
@@ -216,6 +218,7 @@ impl fmt::Display for PathSource {
                 Self::OsConvention => "OS convention",
                 Self::EnvVar => "env variable",
                 Self::ConfigFile => "config file",
+                Self::Override => "override by flag",
```
niklasmohrin, line 221: and here we can then say "command line argument" or so

src/cli.rs
```diff
@@ -74,6 +74,10 @@ pub(crate) struct Cli {
     #[arg(short = 'c', long = "clear-cache")]
     pub clear_cache: bool,
 
+    /// Override config file location
+    #[arg(long = "config", value_name = "FILE")]
+    pub config: Option<PathBuf>,
```
niklasmohrin, line 79: In the issue, I was worried that we might want to have `--config` for something like `tldr --config 'cache.auto_update = true'`, that is, have `--config <some setting>` at some point. For example, git has a `-c` flag which works this way. Thus, `--config-path` would make more sense for the flag we are implementing here. However, I am not sure whether we will ever have such a flag ourselves, and `--config` is more convenient in the meantime. Still, I suspect that `--config-path` is not _much_ longer and is clearer to understand either way, so I am still learning towards that.

What do you think?

cc @dbrgn

docs/src/usage.txt
```diff
@@ -11,13 +11,13 @@ Options:
       --edit-page            Edit custom page with `EDITOR`
       --edit-patch           Edit custom patch with `EDITOR`
   -f, --render <FILE>        Render a specific markdown file
-  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order
-                             of preference [possible values: linux, macos, sunos, windows, android,
-                             freebsd, netbsd, openbsd, common]
+  -p, --platform <PLATFORM>  Override the operating system, can be specified multiple times in order of preference [possible values: linux, macos, sunos, windows, android, freebsd,
```
niklasmohrin, line 14: Here CI fails because we are comparing with the usage printed when the terminal is (I think) 80 chars wide. You can just edit this by hand according to the output of our CI output

src/main.rs
```diff
@@ -142,7 +142,7 @@ fn update_cache(cache: &Cache, archive_source: &str, quietly: bool) -> Result<()
 }
 
 /// Show file paths
-fn show_paths(config: &Config) {
+fn show_paths(custom_config_path: Option<&Path>, config: &Config) {
```
niklasmohrin, line 145: This interface does not really stay in one level of abstraction: The `Config` is a really high level object whereas the `custom_config_path` is just directly the value of one of the paths.

I think we can fix this by tracking the config file location in the `Config` option directly. I want to slowly refactor the config system so that `Config` is kind of "the ultimate config used at runtime", which is the combined values of CLI, env, and config file. In that sense, writing a value that doesn't directly come from the contents of the config file is still reasonable.

You can add a new field `config_file_location: PathBuf` to `Config` and set it as part of `Config::load` - I will propose another change to `Config::load` in the next comment

src/config.rs
```diff
@@ -441,11 +441,18 @@ impl Config {
...
     }
 
-    pub fn load(enable_styles: bool) -> Result<Self> {
+    /// Load and read the config file into a [Config] and return it.
+    ///
+    /// path: The path to the config file could be specified with the path variable.
+    /// if not specified, the default config file will be determined and used.
+    ///
+    /// enable_styles: Whether to use colors, etc.
+    pub fn load(path: Option<&Path>, enable_styles: bool) -> Result<Self> {
```
niklasmohrin, line 450: So I think we should just have `Config::load(path: &Path, enable_styles: bool) -> Result<Self>`. Then, we change `main` to do something like
```rust
let config = match args.config {
    Some(path) => Config::load(path, enable_styles)?,
    None => Config::load_default_location(enable_styles)?,
};
```
where `load_default_location` checks whether there exists a file at the default location (like right now), but `load` errors if the file does not exist. You might want to introduce another private method to avoid some duplication here

src/config.rs
```diff
@@ -509,10 +523,16 @@ pub fn get_config_dir() -> Result<(PathBuf, PathSource)> {
 ///
 /// Note that this function does not verify whether the file at that location
 /// exists, or is a file.
-pub fn get_config_path() -> Result<(PathBuf, PathSource)> {
-    let (config_dir, source) = get_config_dir()?;
-    let config_file_path = config_dir.join(CONFIG_FILE_NAME);
-    Ok((config_file_path, source))
+pub fn get_config_path(path: Option<&Path>) -> Result<(PathBuf, PathSource)> {
```
niklasmohrin, line 526: I think we should keep this function unchanged, except for a rename to `get_default_config_path`; the CLI override logic is not really a concern for the config.rs file

niklasmohrin (changes requested):
Very cool! I left a handful of comments, please feel free to respond if you have any questions and/or disagree with them.

Please also make sure to add tests to cover these changes - we don't want to accidentally break this feature in the future. You tests should probably go into the `tests/lib.rs` file.

Finally, regarding the `--seed-config` argument: I think that it would be weird if `tldr --seed-config --config-path my-config.toml` would create the files in the home directory instead of at my-config.toml. While you are right that it might not be clear that these flags can be used together, we should still do the sensible thing in the case that someone uses them this way. If you feel that this is not in scope for this PR, that is no problem though, just let me know and I will track it in another issue

MHS-0:
@niklasmohrin Thanks for your good suggestions! 
That should be all of them! :)

Let me know if there's anything else that needs to change

src/config.rs
```diff
@@ -254,6 +253,23 @@ impl RawConfig {
     fn new() -> Self {
         Self::default()
     }
+
+    fn from_config_path(config_file_path: &Path) -> Result<Option<RawConfig>> {
```
niklasmohrin, line 257:
```suggestion
    fn read_at(config_file_path: &Path) -> Result<Option<RawConfig>> {
```

src/config.rs
```diff
@@ -254,6 +253,23 @@ impl RawConfig {
...
+        if config_file_path.exists() && config_file_path.is_file() {
```
niklasmohrin, line 258: Okay, so `is_file` already implies `exists`, so to keep the current logic we would only need `is_file`. _However_, we should probably not return `Ok` in the case where `config_file_path` exists, but is not a file (for example when it is a directory).

You can either add this check here (and add a test!), or we can track it in another issue. Either way, please change the check accordingly :)

src/config.rs
```diff
@@ -254,6 +253,23 @@ impl RawConfig {
...
+            let mut config_file = fs::File::open(&config_file_path).with_context(|| {
+                format!("Failed to open config file path at {:?}", &config_file_path)
+            })?;
+            let mut contents = String::new();
+            config_file.read_to_string(&mut contents).with_context(|| {
+                format!("Failed to read from config file at {:?}", &config_file_path)
+            })?;
```
niklasmohrin, line 265: We can abbreviate this using `fs::read_to_string`

src/config.rs
```diff
@@ -366,17 +382,20 @@ pub struct Config {
...
     ///
     /// For this, some values need to be converted to other types and some
     /// defaults need to be set (sometimes based on env variables).
-    fn from_raw(raw_config: RawConfig, relative_path_root: &Path) -> Result<Self> {
+    fn from_raw(raw_config: RawConfig, config_file_path: PathWithSource) -> Result<Self> {
         let style = raw_config.style.into();
         let display = raw_config.display.into();
         let updates = raw_config.updates.try_into()?;
+        // Safe to unwrap, it's a file path, so it should have a directory component
+        let relative_path_root = config_file_path.path().parent().unwrap();
```
niklasmohrin, line 398: _should_! I am not so sure, let's not risk it and make this a `.context(...)?` instead (in particular for future-proofing)

src/config.rs
```diff
@@ -481,7 +473,50 @@ impl Config {
...
+
+    /// Load and read the config file from the given path into
+    /// a [Config] and return it.
+    ///
+    /// path: The path to the config file.
+    /// enable_styles: Whether to use colors, etc.
+    pub fn load(path: &Path, enable_styles: bool) -> Result<Self> {
+        // Load raw config, return with error if it fails.
+        let raw_config = RawConfig::from_config_path(path)?;
+        if raw_config.is_none() {
```
niklasmohrin, line 487: You can use
```rust
let Some(raw_config) = RawConfig::from_config(path)? else {
    bail!(...);
};
```
here to avoid the `unwrap` later

src/config.rs
```diff
@@ -481,7 +473,50 @@ impl Config {
...
+            bail!("Given path doesn't exist or doesn't point to a file");
+        }
+
+        // Convert to config, resolve relative paths from the config file dir
```
niklasmohrin, line 491:
```suggestion
```

src/config.rs
```diff
@@ -481,7 +473,50 @@ impl Config {
...
+    ///
+    /// enable_styles: Whether to use colors, etc.
+    pub fn load_default_path(enable_styles: bool) -> Result<Self> {
+        // Determine path
+        let config_file_path =
+            get_default_config_path().context("Could not determine config path")?;
+
+        // Load raw config if possible, or load default
+        let raw_config = RawConfig::from_config_path(config_file_path.path())?.unwrap_or_default();
+        // Convert to config, resolve relative paths from the config file dir
```
niklasmohrin, line 515:
```suggestion
```

src/main.rs
```diff
@@ -279,7 +277,14 @@ fn main() -> ExitCode {
...
     // Look up config file, if none is found fall back to default config.
-    let config = Config::load(enable_styles).context("Could not load config")?;
+    debug!("Loading config");
+    let config = if args.config_path.is_some() && !args.seed_config {
+        Config::load(args.config_path.as_deref().unwrap(), enable_styles)
+            .context("Could not load config from given path")?
+    } else {
+        Config::load_default_path(enable_styles)
+            .context("Could not load config from default path")?
+    };
```
niklasmohrin, line 287: How do you feel about the following? It avoids the unwrap, which makes it slightly easier in my opinion:
```suggestion
    let config = match &args.config_path {
        Some(path) if !args.seed_config => {
            Config::load(path, enable_styles).context("Could not load config from given path")?
        }
        _ => Config::load_default_path(enable_styles)
            .context("Could not load config from default path")?,
    };
```

src/config.rs
```diff
@@ -481,7 +473,50 @@ impl Config {
...
+        let config =
+            Self::from_raw(raw_config, config_file_path).context("Could not process raw config")?;
+
+        Ok(config.process_config_styles(enable_styles))
```
niklasmohrin, line 519: I wonder if it would be better to remove `enable_styles` completely from this part of the code and just change main to have
```rust
if !enable_styles {
    config.style = StyleConfig::default();
}
```
This way we wouldn't have this "are we printing colors" logic here in config.rs. What do you think?

niklasmohrin (commented):
Looks mostly good, only some small remarks left:

niklasmohrin:
Not all failing lints are from your changes I think, you can ignore those that are not reported inside the diff

src/config.rs
MHS-0: That's OK too!
I personally go for more explicit names usually, out of habit and for clarity, but load works nicely too ;)

src/config.rs
MHS-0: That's OK too!
I personally go for more explicit names usually, out of habit and for clarity, but load works nicely too ;)
I changed it to load for now

src/config.rs
MHS-0: Yeah I added a check for it in main.rs for now.
It's a good move for decreasing repetition I think 🤔
It simplifies the load functions too.

MHS-0:
I fixed some other simple clippy lints while I was at it too (unnecessary semicolons mostly).

An important part to discuss is I think the part where I chose to check if a file exists in the `load_default_path` function, which could be argued to be redundant when the `RawConfig::load` function also checks whether the file exists.

I decided to do it that way, because I want to think of `RawConfig::load` as an individual and independent unit that loads a RawConfig from a file and handles the case of an invalid input by itself. If we think of it that way, then it makes sense to keep the check in the function, especially for future proofing reasons in case we want to use the load function of RawConfig in other places too.

I thought of checking whether a file exists in the load functions, before calling the `RawConfig::load`, and removing the check in the function itself, but that would obviously add the need to check for this case before calling the function, which adds to code complexity in my opinion.

Let me know if this solution is fine, or you think we should do it another way 🤔

src/config.rs
```diff
@@ -254,6 +253,24 @@ impl RawConfig {
     fn new() -> Self {
         Self::default()
     }
+
+    fn load(config_file_path: &Path) -> Result<RawConfig> {
+        if config_file_path.is_file() {
+            let contents = fs::read_to_string(config_file_path).with_context(|| {
```
niklasmohrin, line 259: Regardless of whether or not we _should_ check whether the path is valid here, the check is extraneous anyways here, because read_to_string will return an error if the path is not a file anyways. In fact, the errors printed when removing the `is_file` check are really helpful:
image
image
So either way, we don't need to guard the "open or error" with a "will the open work" check. If we want to codify this further, we should change the signature again to take an already opened `File` instead - then we signal that the caller has already dealt with whether a file exists or not. This is a bit in the spirit of "parse, don't validate". (If you are interested, you can read more motivation for this here: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

One could argue that this function should then just take `String` to begin with - and this is a fair point I think. However, if we go for `File` (or `impl Read` even), then the point of the method becomes to not duplicate the code (and the error messages!). I think that's still a fair reason to have the method.

Thus, I think the cleanest way would be to change `RawConfig::load` here to take an already open file (or `impl Read`). Then, in the callers, we can use `File::open` and check what happened: In `Config::load`, we error if opening fails; in `Config::load_default_path`, we investigate what error we receive from `File::open`, and if its kind is [`ErrorKind::NotFound`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.NotFound), then we swallow the error and continue with the default config.

What do you think?

src/config.rs
```diff
@@ -438,48 +460,45 @@ impl Config {
...
     }
 
-    pub fn load(enable_styles: bool) -> Result<Self> {
-        debug!("Loading config");
+    /// Load and read the config file from the given path into
+    /// a [Config] and return it.
+    ///
+    /// path: The path to the config file.
+    pub fn load(path: &Path) -> Result<Self> {
+        // Load raw config, return with error if it fails.
```
niklasmohrin, line 472: I think the code is clear enough :)
```suggestion
```

src/config.rs
```diff
@@ -509,29 +528,39 @@ pub fn get_config_dir() -> Result<(PathBuf, PathSource)> {
...
+/// the default path.
+pub fn make_default_config(path: Option<&Path>) -> Result<PathBuf> {
+    let config_file_path = if let Some(p) = path {
+        p.into()
     } else {
-        fs::create_dir_all(&config_dir).context("Could not create config directory")?;
-    }
+        let (config_dir, _) = get_config_dir()?;
+
+        // Ensure that config directory exists
```
niklasmohrin, line 549: Ah, so here we create the parent directories only if the default path is used. I think I agree with the decision - can we document it by adding a test?

niklasmohrin (commented):
Great stuff, and good points! I think the PR is already in an acceptable state, so if you get tired of working on it let me know and we are good to go. Otherwise, I am happy to continue discussing, I think we are still improving in every iteration!

MHS-0:
I'm not tired at all! And thanks for working on this with me :)
I applied your suggestions and added some more tests. Let me know if anything catches your eye again 🤔

src/config.rs
```diff
@@ -438,48 +451,51 @@ impl Config {
...
+            Err(e) => {
+                if e.kind().eq(&ErrorKind::NotFound) {
+                    RawConfig::default()
+                } else {
+                    bail!(
+                        "Failed to open config file at {}",
+                        config_file_path.path().display()
+                    );
+                }
+            }
```
niklasmohrin, line 495: I think we can make this slightly nicer as follows:
```suggestion
            Ok(file) => RawConfig::load(file)?,
            Err(e) if e.kind() == ErrorKind::NotFound => RawConfig::default(),
            Err(e) => {
                return Err(e).context(format!(
                    "Failed to open config file at {}",
                    config_file_path.path().display()
                ));
            }
```

This way we also keep the information from the error in the case we return here

niklasmohrin (approved):
Awesome! I have one final thing we can change, everything else is really good :)

src/config.rs
MHS-0: That makes sense!
I committed your suggestion from here :)

niklasmohrin:
Okay, I checked and AccesDenied also appears outside of CI. So we can change the test and merge :)

MHS-0:
Thank you for testing it!
I was a little busy today, though I made a windows 11 VM ready to test it with, but you ended up testing it yourself in the mean time! xD

Alright then. I'll change the test and make a new commit ;)

niklasmohrin (approved):
Thanks!

PR #428, opened 2025-06-23
rename: enable_markdown => disable_markdown

lengyijun:


niklasmohrin:
I think that all kinds of boolean arguments are kind of meh here, but the code is not complex enough to require a change so far. I agree that `raw` is probably better. An enum would be even better, and actually a reasonable improvement.

So sure, we can change it. Just `disable_markdown` is not an improvement

PR #429, opened 2025-08-02
Allow references in `Config`

niklasmohrin:
For #251, I want to use the `Language<'_>` type inside `Config`. The languages will either refer to values read from the config file, or to static strings from `get_languages_from_env`, so just using `Language<'static>` is not an option.

Instead, some input for the `Config` needs to be persisted in the main function for the duration of the program so that the config can reference it. At first I was hoping that this input would be the `contents` string from `RawConfig::load`, but as it turns out you cannot (in general) deserialize strings from toml without having to alter them, for example when they contain escapes like `\n`. Thus, the toml parser seemingly doesn't even try and just throws an error when deserializing into a borrowed string (even if it could in theory just return the correct substring from the input). Given that `RawConfig` should stay static then, the raw config itself is the next best thing to keep alive and have the config reference into.

While this change might seem a bit drastic for little benefit, I am actually pretty happy with it because I want to unify the configuration anyways at some point so that the CLI arguments, environment variables, and the config file are merged at the beginning of the program and then only a single config is used for the everything (no more `enable_styles` everywhere!). At this time, the `Config` would have references into `Cli` anyways, and having the `ConfigLoader` as an entity for this merging also seems natural.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1007,9 +1007,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "0.6.8"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87607cb1398ed59d48732e575a4c28a7a8ebf2454b964fe3f224f2afc07909e1"
+checksum = "40734c41988f7306bb04f0ecf60ec0f3f1caa34290e4e8ea471dcd3346483b83"
 dependencies = [
  "serde",
 ]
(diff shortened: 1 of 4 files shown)


niklasmohrin:
@MatejKafka @beatbrot @mipedja @erickguan @MHS-0 For this and the next PR, I am looking for some reviews again, so if anyone of you would be interested that would be great! I am hoping to push for [version 1.8](https://github.com/tealdeer-rs/tealdeer/milestone/6) soon!

Please let me know via a reaction on this comment if you would like me to include you in future requests for such reviews (use :+1:), or if you want to not get pinged again by me (use :-1:).

Thanks!

niklasmohrin:
And of course, anyone else reading this is also welcome to leave a review!

erickguan (approved):
Clever solution.

tealdeer is very optimized so this PR looks okay. 
One thing is that lifetime is a bit more complicated compared to returning a smart pointer. But tealdeer solves a stable problem. Raw config to config is a stable architecture and no foreseeable changes are coming. I don't see borrowing being a problem.

beatbrot (approved):
Looks good to me :)

Afaik not too much changes except we save a few allocations, right?

Issue #390, opened 2024-11-14
Reduce `--update` invocations in tests


niklasmohrin:
We have a lot of tests that initially run `tldr --update`. This is unnecessary as we could just place the needed pages in the temporary cache directory ourselves. This would a) speed up the tests, because they don't need to download the whole archive from the internet, b) remove unneeded internet traffic and c) make the tests reproducible and runnable in offline environments (like nix builds).

All tests not directly testing the update procedure should be changed.

niklasmohrin:
Once #337 is merged, we can use it to set up our own update server when the tests are running. We should keep at least one test that actually updates from the official source, but it should be ignored by default

PR #430, opened 2025-08-05
 Add `search.languages` and `updates.download_languages` settings

### LINKED ISSUE #251: Add config file option to set prefered language

niklasmohrin:
Closes #251

Best reviewed commit by commit.

Test for the download_languages setting are just destined to be flaky if we update from the upstream repository. I would just test it by hand for now and defer automatic tests until we have #390.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
-use log::debug;
+use log::{debug, info};
 use ureq::{
     http::StatusCode,
     tls::{RootCerts, TlsConfig, TlsProvider},
@@ -27,7 +27,8 @@ pub struct CacheConfig<'a> {
     pub pages_directory: &'a Path,
     pub custom_pages_directory: Option<&'a Path>,
     pub platforms: &'a [PlatformType],
-    pub languages: &'a [Language<'a>],
+    pub search_languages: &'a [Language<'a>],
+    pub download_languages: &'a [Language<'a>],
 }
 
 /// The directory backing this cache is checked to be populated at construction.
@@ -110,7 +111,7 @@ impl<'a> Cache<'a> {
             .filter(|path| path.is_file());
 
         for &platform in self.config.platforms {
-            for language in self.config.languages {
+            for language in self.config.search_languages {
                 let mut search_path = self.config.pages_directory.to_path_buf();
                 search_path.push(language.directory_name());
                 search_path.push(platform.directory_name());
@@ -159,7 +160,7 @@ impl<'a> Cache<'a> {
         };
 
         let mut search_path = self.config.pages_directory.to_path_buf();
-        for language in self.config.languages {
+        for language in self.config.search_languages {
             search_path.push(language.directory_name());
             for platform in self.config.platforms {
                 search_path.push(platform.directory_name());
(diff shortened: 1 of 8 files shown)


tests/lib.rs
```diff
@@ -908,6 +927,64 @@ fn test_common_platform_is_used_as_fallback() {
         .success();
 }
 
+#[test]
+fn test_search_language_precedence() {
+    let testenv = TestEnv::new();
```
erickguan, line 932: This is nice.

src/cache.rs
```diff
@@ -212,7 +213,7 @@ impl<'a> Cache<'a> {
         // Download everything before deleting anything
         let archives = self
             .config
-            .languages
+            .download_languages
```
erickguan, line 216: Do we want to hide from users that their downloaded pages failed? This doesn't make sense to me anymore with `download_languages`. I would go with exposing the error to users. Because I don't want to see a search fail after the download later.

`Self::download` returns an Ok(None) when 404. Hence we don't report anything.

-----

Not relevant to this PR, `fs::rename` should work for these:

```
        // Clear cache directory
        // Note: This is not the best solution. Ideally we would download the
        // archive to a temporary directory and then swap the two directories.
        // But renaming a directory doesn't work across filesystems and Rust
        // does not yet offer a recursive directory copying function. So for
        // now, we'll use this approach.
        fs::remove_dir_all(self.config.pages_directory)?;
        fs::create_dir(self.config.pages_directory)?;
```

erickguan (commented):
Amazing, now I see why you did the previous refactor. A lot of work to make sure zero-copy. But I generally trust lifetime works when I see the CI works. This is great.

I have one doubt with error report that I want to see your response firstly.

src/cache.rs
niklasmohrin: About `fs::rename`: No, as the comment says this doesn't work across filesystems. Since it is common to have `/tmp` as a separate filesystem, we cannot use it and I would rather not extract the archives anywhere else.

Any error after re-creating the directory would be when extracting the zips downloaded from the upstream repository. I don't expect that those will ever be faulty to the degree we can't extract them. If they do, I will deal with it then

src/cache.rs
```diff
@@ -14,20 +14,21 @@ use ureq::{
...
-pub struct Language<'a>(pub &'a str);
-
 #[derive(Clone)]
 pub struct CacheConfig<'a> {
     pub pages_directory: &'a Path,
     pub custom_pages_directory: Option<&'a Path>,
     pub platforms: &'a [PlatformType],
-    pub languages: &'a [Language<'a>],
+    pub search_languages: &'a [Language<'a>],
+    pub download_languages: &'a [Language<'a>],
```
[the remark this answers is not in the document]
erickguan: cargo doc can generate a JSON format which works well. Though you want to update the website with versions when you release. This is more work.

src/cache.rs
erickguan: > So the thing is that in the default case where we use get_languages_from_env, the language list is something like ["en_US", "en"]. The download for en_US fails, but for en it works, which is expected.

True, no point to output errors in this case. However, this is more like a cascading fallback to the same language (not exactly language variants but good enough difference for developers).

> I don't think we should output warnings by default. However, we could return the list of languages that actually worked from this function and use them in main.rs to output something like "Successfully updated cache with the following languages: en, de.".
> I think I will also promote the debug! log to info! and make sure to mention the logging in the section of the docs.

Yeah, I will be able to find potentially what language is missing. I think this is good enough to start and not far away from implementing it.

If I am reading the code right, users could send a junk language tag and hope the download works. That was the reason I think it will be more useful to output an error. I also see your plan will work equally well.

> About fs::rename: No, as the comment says this doesn't work across filesystems. Since it is common to have /tmp as a separate filesystem, we cannot use it and I would rather not extract the archives anywhere else.

True when you don't control the OS. A workaround is to use another folder. But as your comment says, it's fine to wait until someone complains.

src/config.rs
```diff
@@ -611,3 +655,71 @@ fn test_relative_path_resolution() {
...
+mod test {
+    use super::*;
+
+    mod language {
+        use super::*;
+
+        #[test]
+        fn missing_lang_env() {
+            let lang_list = get_languages(None, Some("de:fr"));
+            assert_eq!(lang_list, [Language("en")]);
```
erickguan, line 701: A stupid question. Why are `de` and `fr` missing here? I know it's copy & paste code.

erickguan (approved):
LGTM.

Feel free to ping me again if you want to add logs in this PR or the next too!

MHS-0:
Hello!
Sorry for not being available for the previous PRs. Feel free to tag me :)
I'll come help if I can.

I reviewed this PR btw, and I think you did a good job! I'll comment on any part that catches my attention.

PR #435, opened 2025-09-21
Add config option to search all known platforms

### LINKED ISSUE #278: Search other Platforms when command isn't found

niklasmohrin:
Closes #278

The current version is implemented to also use extra platforms when passing `--platform`, so for example running `tldr --platform linux` would still show Windows pages with the default config.
We could also make it so that the setting is only taken if `--platform` is not used.
I actually suspect that the latter would be more pleasant.

What do you think? @k12ish @marchersimon @osalbahr @erickguan @MHS-0

### DIFF
--- a/docs/src/config_search.md
+++ b/docs/src/config_search.md
@@ -12,3 +12,13 @@ Either way, the language used can be overwritten using the `--language` command 
     [search]
     # Show pages in German if available, otherwise show in English
     languages = ["de", "en"]
+
+## `try_all_platforms`
+
+Whether or not all known platforms should be tried when searching for pages.
+For example, pages for Windows would also be found when running on Linux.
+By default, this setting is `true`.
+
+    [search]
+    # Don't use extra platforms in search
+    try_all_platforms = false
--- a/src/config.rs
+++ b/src/config.rs
@@ -217,6 +217,7 @@ struct RawDirectoriesConfig {
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 struct RawSearchConfig {
     pub languages: Option<Vec<String>>,
+    pub try_all_platforms: Option<bool>,
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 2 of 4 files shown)


docs/src/config_search.md
```diff
@@ -12,3 +12,13 @@ Either way, the language used can be overwritten using the `--language` command
...
+
+## `try_all_platforms`
+
+Whether or not all known platforms should be tried when searching for pages.
+For example, pages for Windows would also be found when running on Linux.
+By default, this setting is `true`.
+
+    [search]
+    # Don't use extra platforms in search
+    try_all_platforms = false
```
erickguan, line 24:
```suggestion
## `search_other_platforms`

If tealdeer can’t find a page for the user’s current platform, it will try looking on other platforms instead.
For instance, when running on Linux, if no Linux-specific page exists, tealdeer will fall back to searching for a Windows page.

This option is enabled by default (true).

    [search]
    # Restrict search to the current platform only
    search_other_platforms = false
```

This might read a bit better.

niklasmohrin: Fair enough! What do you think about `include_all_platforms` as a middle ground?

Or maybe I will just implement something like `platforms = ["all"]` and close the other issue as well. I guess this would make the most sense since we want to have some kind of search.platforms setting anyways

erickguan: I like this idea!

Issue #431, opened 2025-08-07
Add `display.show_title` option


techieshark:
Hi all, thanks for this awesome project.

Random small thought - adding the title of what was searched for may make the output make more sense.

Consider the example:

> [Showing a random page on shell start](https://tealdeer-rs.github.io/tealdeer/tips_and_tricks.html#showing-a-random-page-on-shell-start)
> ``` tldr --quiet $(tldr --quiet --list | shuf -n1) ```
(sidenote: on vanilla Mac that above command might instead be `tldr --quiet $(tldr --quiet --list | sort -R |head -1)` as `shuf` isn't installed)

And assume that the random command it picked was `tar`.

Observe that you need to read a bit before realizing that is the command it must have picked:

<img width="891" height="674" alt="Image" src="https://github.com/user-attachments/assets/76d5ee80-2af4-4492-ac57-36a05ee57f6a" />

Now compare with [`tlrc`](https://github.com/tldr-pages/tlrc?tab=readme-ov-file#tlrc) and observe that it includes 'tar' (the selected command) at the top:

<img width="830" height="511" alt="Image" src="https://github.com/user-attachments/assets/0c14eef3-06be-404b-b535-ee89be3ebe42" />

I didn't see a command line or config option to change this but perhaps I missed it?

niklasmohrin:
Currently we are explicitly omitting the title from the output, because for normal use it is just a waste of space to display the name when you just typed it to query the page (`tldr tar` - I know that I am looking at the page for "tar"). See https://github.com/tealdeer-rs/tealdeer/blob/630b7f442376232012b01f4fde5fcae6fc236aa2/src/formatter.rs#L47-L54

I don't have a strong opinion on whether or not there should be a config option to show the heading. I do think that the default output should not include the heading for the reasons above. So if someone wants to send a PR to add a `display.show_title` option, I am happy to include it.

For the time being, you can also do `PAGE=$(tldr --quiet --list | shuf -n1); echo $PAGE; tldr --quiet $PAGE` for a cheap variant of this feature.

niklasmohrin:
Note that in addition to adding the boolean value to show the title at all, we probably also want to add the option to customize the style of the title by adding another field to `RawStyleConfig` and propagating it like the other style options. This is easy though, the hardest thing will be choosing the default style :)

niklasmohrin:
Closed by #439

PR #439, opened 2025-10-02
Add show-title option to display command titles in output

hex1c:
This PR adds a new CLI option: --show-title.
#431

### DIFF
--- a/docs/src/config_display.md
+++ b/docs/src/config_display.md
@@ -23,6 +23,7 @@ Set this to enforce more compact output, where empty lines are stripped out
     compact = true
 
 ## `show_title`
+
 Display the command name at the top of the page output (default `false`).
 
     [display]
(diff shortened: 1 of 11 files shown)


niklasmohrin (commented):
Nice job! Can you add a quick integration test to tests/lib.rs and user documentation to the docs/ directory for the config setting (there is already a .md file about the "display" section)?

docs/src/config_display.md
```diff
@@ -21,3 +21,12 @@ Set this to enforce more compact output, where empty lines are stripped out
 
     [display]
     compact = true
+
+## `show_title`
```
niklasmohrin, line 25:
```suggestion
## `show_title`

```

niklasmohrin (changes requested):
Awesome! I have a small change request from looking at the expected test output. Other than that, our CI is complaining about the code formatting which can be fixed by running `cargo fmt`. Finally, I left two minor suggestions which you can directly commit in the Github UI, or copy locally. After that, we should be good to go :)

PR #458, opened 2026-01-11
Add `--override-config` flag

### LINKED ISSUE #449: Add `--extra-config` flag
To override config settings temporarily, a new flag `--extra-config` should be added. This is similar to git's `-c` flag. The flag should be used with a valid toml string specifying which settings are changed, for example
```bash
tldr --extra-config 'updates.tls_backend = "native-tls"' --update
```

The implementation could be somewhat difficult because we cannot just use our regular config parsing on the string, because the parsing will fill in default values for other settings and we won't be able to tell which of the config values were actually intended to be overridden. Thus, I would suggest that anyone looking to implement this should discuss possible approaches here first.

The exact naming of the flag is also not set in stone. In particular, `--config` has been suggested as a shorter alternative. I feel that since we already have `--config-path` where the default config file is not used at all anymore, it could be confusing that `--config` does still use the default config location. Thus, I prefer the `--extra-config` name right now. As usual, I would not add a short variant of the flag for now, since the single letters could be reserved for other things later by the tldr client specification, leading to an annoying migration. We can always add a short version later, if there is demand.

Please let me know what you think in this thread :)

Giftzwerg02:
Addresses #449, as discussed this is a simple implementation to merge two toml configs using a recursive approach.
As expected, this does not affect the runtime at all (at least on my machine)

(The tests are added are a little crude right now, but I didn't want to immediately add a lot of setup procedure)


niklasmohrin:
After some more thinking, I believe we should really do the other approach where we parse the argument as `name=value` ourselves and then don't do any recursing. Do you want to continue working on this, even if we are not doing the approach you initially suggested?

Giftzwerg02:
Yeah sure, I'll probably have time tomorrow to have a look at this again

src/config.rs
```diff
@@ -744,6 +820,63 @@ mod test {
         );
     }
 
+    mod override_config {
```
niklasmohrin: can we add a test that uses "dependent defaults"? what I mean is that when converting the raw config into a proper config we use, for example, the value of `search.languages` as the default for `updates.download_languages`. thus, if I overwrite search.languages, this should propagate to download_languages

this would also be good to have as a test that uses our actual Config type as the result

src/cli.rs
niklasmohrin: we have one failing CI job which is about the `docs/src/usage.txt` file: this file should contain the output of `tldr --help`. you can regenerate the file by running `cargo run -- --help > docs/src/usage.txt`, but sometimes this still does not make CI pass because the terminal in the CI runner might have a different width than what you have locally. the easiest way to make it pass is to just copy the output from the github actions job

niklasmohrin (commented):
Very cool, I like this feature and the code a lot, thank you! I have some comments on how I think we can improve the code and on what still needs to be done to make the user experience complete.

niklasmohrin:
@Giftzwerg02 do you want to continue working on this, or should I take over? I think this is a very cool feature and I want to include it in the next release :)

Giftzwerg02:
> @Giftzwerg02 do you want to continue working on this, or should I take over? I think this is a very cool feature and I want to include it in the next release :)

Sorry, but I've quite gotten busy with other things since my PR, and I'd appreciate it if you could take over since I am working on other projects right now.

src/config.rs
```diff
@@ -583,9 +584,29 @@ pub struct ConfigLoader {
...
 impl ConfigLoader {
-    fn read_internal(path: PathWithSource, allow_not_found: bool) -> Result<Self> {
-        match fs::read_to_string(&path.path) {
-            Ok(content) => Ok(Self {
+    fn read_internal(
+        path: PathWithSource,
+        allow_not_found: bool,
+        overrides: Option<Vec<String>>,
+    ) -> Result<Self> {
+        match (fs::read_to_string(&path.path), overrides) {
```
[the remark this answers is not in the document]
niklasmohrin: current commit does not yield any difference for this, so no optimization is needed here

PR #467, opened 2026-02-25
Support XDG directory spec on macOS by switching `app_dirs2` to `etcetera`

### LINKED ISSUE #311: Respect XDG_* variables on macos if set

niklasmohrin:
Closes #311

Similar to how `bat` does it, we now have a global struct which holds all the queried system directories. We prefer the XDG directories, but use the "native" directories as a fallback. This should ensure that when MacOs users upgrade from an older version, their existing config and cache are still used. After deleting the old directories, the new ones should be used by tealdeer automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable, 1.85.0]
+        toolchain: [stable, 1.87.0] # MSRV
         include:
           - platform: windows-latest
             exe_suffix: .exe
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,18 +74,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
-name = "app_dirs2"
-version = "2.5.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a7e7b35733e3a8c1ccb90385088dd5b6eaa61325cb4d1ad56e683b5224ff352e"
-dependencies = [
- "jni",
- "ndk-context",
- "winapi",
- "xdg",
-]
-
-[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


niklasmohrin:
I don't have a MacOs machine to test this, so it would be nice if someone else could check that the following works:
1. Use your current tealdeer version to update the cache, and set up a config file (`tldr --seed-config`)
2. Run your normal tealdeer commands, but using the tldr executable from this PR (see below for steps on how to get it).
3. Check the output of `tldr --show-paths` with the new executable, and afterwards delete (some of) them.
4. Run your tealdeer commands with the new executable again and check that the XDG paths are used.

To get an executable with the changes from this PR, either compile it yourself from source, or use the binaries built in CI here: https://github.com/tealdeer-rs/tealdeer/actions/runs/28721862364 (scroll to "artifacts" at the bottom).

Tagging some people from the issue as possible testers: @diktomat @uncenter @bartekpacia @baodrate 

Thanks to everyone who helps!

niklasmohrin:
So the paths on Windows are changed somewhat significantly. Before
```
Config dir:       C:\Users\runneradmin\AppData\Roaming\tealdeer\tealdeer\ (OS convention)
Config path:      C:\Users\runneradmin\AppData\Roaming\tealdeer\tealdeer\config.toml (OS convention)
Cache dir:        C:\Users\runneradmin\AppData\Local\tealdeer\tealdeer (OS convention)
Pages dir:        C:\Users\runneradmin\AppData\Local\tealdeer\tealdeer\tldr-pages\
Custom pages dir: C:\Users\runneradmin\AppData\Local\tealdeer\tealdeer\pages\ (OS convention)
```
and after
```
Config dir:       C:\Users\runneradmin\AppData\Roaming\tealdeer\config\ (OS convention)
Config path:      C:\Users\runneradmin\AppData\Roaming\tealdeer\config\config.toml (OS convention)
Cache dir:        C:\Users\runneradmin\AppData\Local\tealdeer\cache (OS convention)
Pages dir:        C:\Users\runneradmin\AppData\Local\tealdeer\cache\tldr-pages\
Custom pages dir: C:\Users\runneradmin\AppData\Roaming\tealdeer\data\pages\ (OS convention)
```

I think I don't want to introduce a builtin migration strategy like for Macos because a) the previous paths on Windows are strictly less nice than our old paths on Macos (the custom pages should really be in Roaming rather than Local), and b) this would require more code because there is no strategy in etcetera that derives these paths. I will just add a warning to the changelog.

PR #471, opened 2026-03-11
Add `display.indent` config option

### LINKED ISSUE #394: Remove indentation
Is it possible to remove the indents for displaying commands?

ptcodes:
This pull request introduces a new -C / --compact flag that displays output without indentation for commands, texts, and descriptions. This makes the output more compact and suitable for environments where minimal whitespace is preferred.

When the flag is enabled, indentation is removed from the generated output, resulting in a denser and slightly more concise format.

Fixes #394.


niklasmohrin:
Hi @ptcodes! This was not clear from the issue description, but I had a slightly different UX in mind for this feature. Instead of having a command line flag to switch between two different modes, I think it would be better if we added an option to our configuration file so that users can specify any number of spaces they want. I think it should work like this:
```toml
[display]
indent = 6 # changes indent to 6 spaces
```

Once we have #458 this option will also be available from the CLI. What do you think about this approach?

You can find the config loading logic by following the `let config = ...` line in `main.rs`, it leads into `config.rs` where all the options are defined.

ptcodes:
Hi @niklasmohrin,

That makes sense! However, that parameter would only affect the indentation of the commands, right?

In the current implementation in this pull request, the indentation is removed from texts (2 spaces), descriptions (2 spaces), and commands (6 spaces). See the screenshot for a clearer illustration:

Screenshot_20260311_184013

ptcodes:
Hi @niklasmohrin, I've introduced the indent option to the config file. Could you check it out please?

niklasmohrin:
I see, you are totally right about having indents in multiple places. Being able to just configure one kind of indent is a bit odd; from what the user wrote in the issue I suppose it would make sense to say that tealdeer always prints a base indent of 2 spaces and then some extra indent for code, and we make this extra indent configurable. However, that feels a bit strange.

What would be the more thorough UI is if we allowed configuring all the indents. I am not sure what the best way for that would be though, my intuition says it would be something like
```toml
[display]
indent = {
    description = 2,
    code = 4,
}
```
However, we need to ensure that the default logic makes sense then, in particular it should be possible to override only some of the indents and keep the defaults for the rest.

This is totally blowing the scope of the issue and we are leaving "good first issue" territory here, sorry about that. If you want, we can implement the "base indent of 2 spaces and config for additional code indent" version in this PR and I can think about extending it later; if you are up to the challenge of introducing this whole table and considering what the best UI for this is (which might mean we try out some ideas that ultimately turn out to be incorrect), then I am happy to do this in this PR too. Just let me know which option you prefer and I will review accordingly

ptcodes:
No worries. I like the new approach.

First, I think we might want to flatten the table so it looks like this, similar to other style-related config sections:

```
[style.example_variable]
foreground = "cyan"
underline = true
bold = false
italic = false
```

I’d also suggest renaming the variables: `description` → `base` (value `2`), and `code` → `command` (value `6`). I think these names are a bit more intuitive.

That would give us:

```
[display.indent]
base = 2
command = 6
```

What do you think?

ptcodes:
The code is added. @niklasmohrin please take a look when you have a chance. Thank you!

docs/src/config_display.md
```diff
@@ -37,3 +37,25 @@ show_title = true
 
 When enabled, the command name will be displayed at the top of the output,
 styled with the `command_name` style configuration.
+
+## `indent`
+
+Controls the indentation of the output via two sub-keys.
```
niklasmohrin, line 43: As far as I know, writing sub tables with sections like `display.indent` is semantically equivalent with writing a dictionary value like `display.indent = { base = 2, command = 6}`. I think I read that only somewhat recent versions of the toml crate support splitting such a dictionary value across multiple lines like
```toml
[display]
indent = {
  base = 2,
  command = 6,
}
```
because I think `cargo` itself added support for doing this in Cargo.toml, but I may be wrong.

Either way, I feel that we could mention this in one sentence here, like "You can also configure both subkeys in a single line like this: ..."

src/config.rs
```diff
@@ -171,6 +179,26 @@ struct RawDisplayConfig {
...
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+struct RawIndent {
+    #[serde(default = "default_base_indent")]
+    base: usize,
+    #[serde(default = "default_command_indent")]
+    command: usize,
+}
+
+#[allow(clippy::derivable_impls)] // Explicitly control defaults
```
niklasmohrin: Not needed because the lint is not triggered here; this would only trigger if we would set both values to 0, which is what the derived Default would have done
```suggestion
```

src/config.rs
```diff
@@ -312,6 +344,8 @@ impl Default for RawConfig {
         raw_config.style.example_code.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.foreground = Some(RawColor::Cyan);
         raw_config.style.example_variable.underline = true;
+        raw_config.display.indent.base = 2;
+        raw_config.display.indent.command = 6;
```
niklasmohrin: This is already what we get when using `RawDisplayConfig::default` above.

In general, this whole method here should not exist and this should be done in `RawStyleConfig::default`. But don't worry about it, this is an issue outside of your PR
```suggestion
```

src/output.rs
niklasmohrin, line 1: Hm, your changes here really show how poorly structured the code was previously with respect to those indents. What I would expect is that there is a single place that reads what is inside `config.display.indent` and then emits something further down the pipeline. Right now though, we handle the base indent when printing snippets but emit additional code snippets for command indents.

Since a PageSnippet does not really carry any meaning except what style we should select to print it, how do you feel about removing the indent logic from print_snippet and instead emitting additional snippets of the same Title, Text, and Description type in highlight_lines? Then we can just pass the entire indent config into highlight_lines and only use it right there

tests/lib.rs
```diff
@@ -811,6 +811,24 @@ fn test_rendering_color_never() {
...
 
+/// An end-to-end integration test for the indent config option
+#[test]
+fn test_rendering_with_indentation() {
+    let testenv = TestEnv::new().install_default_cache();
+    let expected_custom_indentation = include_str!("rendered/inkscape-compact-no-color.expected");
+
+    // Configure to set base and command indents
+    testenv.append_to_config("display.indent.base = 2\n");
+    testenv.append_to_config("display.indent.command = 2\n");
```
niklasmohrin: Let's do something funky with both values so that we know that both are used, something like
```suggestion
    testenv.append_to_config("display.indent.base = 3\n");
    testenv.append_to_config("display.indent.command = 1\n");
```

niklasmohrin (commented):
Very nice, I have a small refactor in mind that I would still like to see here, along with some minor changes, but then we are good to go!

tests/lib.rs
ptcodes: Hi @niklasmohrin, these are fair suggestions!  I'm pushing a commit with the update.

src/config.rs
niklasmohrin: I think you removed this line on the wrong struct, the new RawIndent struct still has this, but doesn't need it, whereas the other one actually needs this `allow` because all values are the default

niklasmohrin:
hi @ptcodes, sorry for the long wait. The changes look good, although I think you removed the `allow(...)` on the wrong struct (regarding https://github.com/tealdeer-rs/tealdeer/pull/471#discussion_r3069274183) and the change I proposed in https://github.com/tealdeer-rs/tealdeer/pull/471#discussion_r2996663606 is still missing. The idea is that there should be only one place in the code where we use what is stored in `config.display.indent`. This should be the single place in the codebase that "handles indents". Right now, the two values are read in two different places. I describe a possible approach in the comment above, but of course let me know if you think it should be done differently.

Also, let me know if you need more input or help with this, then I can try to make some more concrete suggestions. Finally, if you feel that you don't want to do this refactor then I can also take over at any point. I think it will not be too much work though and the Rust compiler will mostly guide you to where the code needs to be edited

ptcodes:
Hi @niklasmohrin you were totally right, I had deleted the wrong line (it's now back) and didn't implement what you had asked.

So now I consolidated all indent logic into highlight_lines so there is a single place in the codebase that reads and applies config.display.indent.

Now highlight_lines receives the full Indent config and handles both values. For Description, Text, and Title lines it emits an indent snippet of the same type followed by the content snippet and an explicit Linebreak consistent with how ExampleCode indentation already worked. print_snippet no longer knows about indentation at all.

Also why the snapshot files changed: 
Previously the indent spaces were written as **raw unstyled bytes** before calling .paint() on the text. Now the spaces are emitted as a snippet of the same type and go through .paint() as well. For lines whose style carries a color (e.g. example_text in green), the spaces are now wrapped in the same ANSI escape codes as the text content: \x1b[32m  \x1b[0m\x1b[32mtext\x1b[0m instead of  \x1b[32mtext\x1b[0m. 

Visually the output is identical, but the raw bytes differ, so the byte-exact snapshot files needed updating.

niklasmohrin (approved):
Awesome, thanks! About the updated test files: This is definitely what is correct with this change (this way is exactly what we did for code already), I wonder whether this makes sense in the bigger picture though. Like, if I would set the style to be underlined, then would the indent also be underlined? I will check what is the behavior and maybe open a follow up issue for this.

PR #472, opened 2026-03-11
Add builtin `tldr tealdeer` page

### LINKED ISSUE #218: Docs: Add better usage instructions to docs
In #108 the argument parsing was switched from docopt to clap. This resulted in the removal of the examples:

```
Examples:

    $ tldr tar
    $ tldr --list

To control the cache:

    $ tldr --update
    $ tldr --clear-cache

To render a local file (for testing):

```

The usage docs should probably be updated to include some examples like this. In general, we should explain to the user how tealdeer can be used. (This _can_ be done through examples, but other means are possible as well.)

ptcodes:
Adds examples section to the bottom:

```
To view the user documentation, please visit https://tealdeer-rs.github.io/tealdeer/.

EXAMPLES:
# View the TLDR page for ip:
tldr ip

# View a multi-word TLDR page:
tldr git rebase

# Update TLDR with pages for a specific language:
tldr --update --language es

# Specify the language for the page:
tldr tar --language es
```

Fixes #218.

### DIFF
--- a/pages/tealdeer.md
+++ b/pages/tealdeer.md
@@ -1,25 +1,9 @@
 # tldr
 
-> A fast tldr client written in Rust.
+> This is a builtin page that shows information for your installed tealdeer version.
 > More information: <https://tealdeer-rs.github.io/tealdeer/>.
 
-- Show the tldr page for a command:
-
-`tldr {{command}}`
-
-- Show the tldr page for a subcommand:
-
-`tldr {{command}} {{subcommand}}`
-
-- Show the tldr page for a specific platform:
-
-`tldr --platform {{android|common|freebsd|linux|netbsd|openbsd|osx|sunos|windows}} {{command}}`
-
-- Show the tldr page in a specific language:
-
-`tldr --language {{language_code}} {{command}}`
-
-- Download or update the local page cache:
+> This page shows tealdeer specific functionality. See tldr tldr for more examples.
 
 `tldr --update`
 
--- a/pages/tealdeer.md
+++ b/pages/tealdeer.md
@@ -5,12 +5,6 @@
 
 > This page shows tealdeer specific functionality. See tldr tldr for more examples.
 
-`tldr --update`
-
-- List all pages in the cache:
-
-`tldr --list`
-
 - Render a local markdown file as a tldr page:
 
 `tldr --render {{path/to/file.md}}`
@@ -27,6 +21,10 @@
 
 `tldr --seed-config`
 
+- Override config file location:
+
+`tldr --config-path <FILE>`
+
 - Open a custom page for a command in `$EDITOR` (creates it if it doesn't exist):
 
 `tldr --edit-page {{command}}`
@@ -38,3 +36,7 @@
 - Clear the local cache:
 
 `tldr --clear-cache`
+
+- If auto update is configured, disable it for this run:
+
+`tldr --no-auto-update`
--- a/pages/tldr.md
+++ b/pages/tldr.md
@@ -1,0 +1,37 @@
+# tldr
+
+> Display simple help pages for command-line tools from the tldr-pages project.
+> Note: The `--language` and `--list` options are not required by the client specification, but most clients implement them.
+> More information: <https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#command-line-interface>.
+
+- Print the tldr page for a specific command (hint: this is how you got here!):
+
+`tldr {{command}}`
+
+- Print the tldr page for a specific subcommand:
+
+`tldr {{command}} {{subcommand}}`
+
+- Print the tldr page for a command in the given language (if available, otherwise fall back to English):
+
+`tldr {{[-L|--language]}} {{language_code}} {{command}}`
+
+- Print the tldr page for a command from a specific platform:
+
+`tldr {{[-p|--platform]}} {{android|cisco-ios|common|dos|freebsd|linux|netbsd|openbsd|osx|sunos|windows}} {{command}}`
+
+- Update the local cache of tldr pages:
+
+`tldr {{[-u|--update]}}`
+
+- List all pages for the current platform and `common`:
+
+`tldr {{[-l|--list]}}`
+
+- Browse tldr pages in a terminal window (`fzf` must be available):
+
+`tldr {{[-l|--list]}} | fzf --preview "tldr {1} --color=always" --preview-window=right,70% | xargs tldr`
+
+- Print the tldr page for a random command:
+
+`tldr {{[-l|--list]}} | shuf {{[-n|--head-count]}} 1 | xargs tldr`
--- a/pages/tldr.md
+++ b/pages/tldr.md
@@ -1,37 +1,56 @@
 # tldr
 
-> Display simple help pages for command-line tools from the tldr-pages project.
-> Note: The `--language` and `--list` options are not required by the client specification, but most clients implement them.
-> More information: <https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md#command-line-interface>.
+> A fast tldr client written in Rust.
+> More information: <https://tealdeer-rs.github.io/tealdeer/>.
 
-- Print the tldr page for a specific command (hint: this is how you got here!):
+- Show the tldr page for a command:
 
 `tldr {{command}}`
 
-- Print the tldr page for a specific subcommand:
+- Show the tldr page for a subcommand:
 
 `tldr {{command}} {{subcommand}}`
 
-- Print the tldr page for a command in the given language (if available, otherwise fall back to English):
+- Show the tldr page for a specific platform:
 
-`tldr {{[-L|--language]}} {{language_code}} {{command}}`
+`tldr --platform {{android|common|freebsd|linux|netbsd|openbsd|osx|sunos|windows}} {{command}}`
 
-- Print the tldr page for a command from a specific platform:
+- Show the tldr page in a specific language:
 
-`tldr {{[-p|--platform]}} {{android|cisco-ios|common|dos|freebsd|linux|netbsd|openbsd|osx|sunos|windows}} {{command}}`
+`tldr --language {{language_code}} {{command}}`
 
-- Update the local cache of tldr pages:
+- Download or update the local page cache:
 
-`tldr {{[-u|--update]}}`
+`tldr --update`
 
-- List all pages for the current platform and `common`:
+- List all pages in the cache:
 
-`tldr {{[-l|--list]}}`
+`tldr --list`
 
-- Browse tldr pages in a terminal window (`fzf` must be available):
+- Render a local markdown file as a tldr page:
 
-`tldr {{[-l|--list]}} | fzf --preview "tldr {1} --color=always" --preview-window=right,70% | xargs tldr`
+`tldr --render {{path/to/file.md}}`
 
-- Print the tldr page for a random command:
+- Show the raw markdown source of a page instead of rendering it:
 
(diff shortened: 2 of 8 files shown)


src/cli.rs
```diff
@@ -6,6 +6,21 @@ use clap::{builder::ArgAction, ArgGroup, Parser};
 
 use crate::types::{ColorOptions, PlatformType};
 
+const AFTER_HELP: &str = r#"To view the user documentation, please visit https://tealdeer-rs.github.io/tealdeer/.
+
+EXAMPLES:
```
niklasmohrin: (I prefer the `tldr tealdeer` approach, but in case we stick to this one)

Right now, the output looks like this on my machine: 

Image

I think it would be cool if we could get the "Examples" heading to look the same as "Arguments", "Options", etc.; do you know of any way to make this work? I am not sure what clap is doing internally here, but I can observe is that they do not always print the decorations. In particular, when redirecting the output to a file they don't have them as we can see by running `cargo run 2> output.txt; cat output.txt`. If there is no good/easy way to do it, we can also consider leaving it as is

niklasmohrin (commented):
Thanks for picking this up!

I wonder if including these examples in `--help` is the best way to show examples, tealdeer is a program for showing examples on its own; maybe we could suggest running `tldr tldr` or `tldr tealdeer` in our help text and move the examples to those pages? I know there is already a page for `tldr` itself which includes the first two examples, but the language specific updates are tealdeer-specific.

What do you think? To me it seems that having `tldr tealdeer` with tealdeer specific things built in would be cool, we would just have to add an exception to our cache logic to show this page directly instead of looking for it on disk.

src/cli.rs
ptcodes: I like the `tldr tealdeer approach!

I’ve been using tldr for years, but for some reason it never occurred to me that you can run tldr on itself (e.g., `tldr tldr` or `tldr tealdeer`). That’s a really nice discovery pattern.

So maybe we could do something like this:

- Create a tealdeer page.
- Once the pull request is accepted and the page is live (since it’s in a different repository), add a short comment like:

```
To view the user documentation, please visit https://tealdeer-rs.github.io/tealdeer/.

To view usage examples, run tldr tealdeer.
```

So this way there is no need to think about the headings (and no, there was no easy way to make the the Examples heading look the same as "Arguments", etc).

What do you think @niklasmohrin?

niklasmohrin: Yes I also didn't know about `tldr tldr` until someone reported that an example from there does not work with tealdeer 😅

For `tldr tealdeer` I think we probably don't want to show a page from the tldr-pages repository because then also older versions of tealdeer will start to show this page with all the new features and that would be confusing. For example Debian still has tealdeer 1.5 which is super old.

I think we can have this page in our repo and use the normal tldr page format. Then we use something like `include_str! ` to include the page in the tealdeer binary and when someone queries for the tealdeer page we show that instead of using our `Cache` struct.

What do you think?

niklasmohrin: Of course this makes the issue more complicated than what it said originally, but I think that it should be doable even without experience in the tealdeer code. Feel free to ask questions if you get stuck

ptcodes: That's a great idea @niklasmohrin  I've followed your suggestion, please take a look.

We now have the new file pages/tldr.md which ~~for the moment is just a copy of https://raw.githubusercontent.com/tldr-pages/tldr/refs/heads/main/pages/common/tldr.md~~ is now updated with examples specific to tealtldr. 

Please let me know if you want to update the file with tealtldr's specific examples. I can put together a test or two once the file is finalized.

Thanks!

src/main.rs
```diff
@@ -68,6 +68,7 @@ const APP_INFO: AppInfo = AppInfo {
     name: NAME,
     author: NAME,
 };
+const SELF_PAGE: &str = include_str!("../pages/tldr.md");
```
niklasmohrin: Can we avoid `..` by doing something like here? https://stackoverflow.com/a/68865971/13679671

src/main.rs
```diff
@@ -407,20 +408,24 @@ fn try_main(args: Cli, enable_styles: bool) -> Result<ExitCode> {
...
-                         Try updating with `tldr --update`, or submit a pull request to:\n\
-                         https://github.com/tldr-pages/tldr",
-                        &command
-                    ),
-                );
-            }
-
-            return Ok(ExitCode::FAILURE);
+        let lookup_result = if command == "tldr" || command == "tealdeer" {
+            PageLookupResult::with_embedded(SELF_PAGE)
```
niklasmohrin: I think it would be okay to not change the definition of `PageLookupResult` and keep using it only for what is returned from `Cache`. As `print_page` does not do anything with the result except calling `.reader` on it, we can change the definition of `print_page` to take the reader (or `impl Read` or so) directly.

src/main.rs
```diff
@@ -407,20 +408,24 @@ fn try_main(args: Cli, enable_styles: bool) -> Result<ExitCode> {
...
-                        "Page `{}` not found in cache.\n\
-                         Try updating with `tldr --update`, or submit a pull request to:\n\
-                         https://github.com/tldr-pages/tldr",
-                        &command
-                    ),
-                );
-            }
-
-            return Ok(ExitCode::FAILURE);
+        let lookup_result = if command == "tldr" || command == "tealdeer" {
```
niklasmohrin: I think I would not interfere with `tldr tldr`, this page should still be the same as before; only `tldr tealdeer` should show our custom page.

pages/tldr.md
```diff
@@ -0,0 +1,56 @@
+# tldr
+
+> A fast tldr client written in Rust.
+> More information: <https://tealdeer-rs.github.io/tealdeer/>.
```
niklasmohrin: We should say something like the following here:

> This is a builtin page that shows information for your installed tealdeer version.

src/main.rs
niklasmohrin: and then we can also shrink our custom page to only show tealdeer specific behavior and it should say something like

> This page shows tealdeer specific functionality. See `tldr tldr` for more examples.

src/cli.rs
niklasmohrin: Thanks, good to hear that you are up for it! I left some comments in a new thread :)

src/main.rs
ptcodes: Yeah, I agree. I should've not touched PageLookupResult.

ptcodes:
Hi @niklasmohrin, thanks for the review, everything makes sense.

I've update the PR:
- Reverted my changes in PageLookupResult and re-implemented according to your suggestion.
- Truncated tealdeer.md. Should it be shrunk further? As there’s an overlap of commands between tldr and tealdeer.
- Some other minor changes according to your comments.

pages/tealdeer.md
```diff
@@ -0,0 +1,40 @@
...
+> This is a builtin page that shows information for your installed tealdeer version.
+> More information: <https://tealdeer-rs.github.io/tealdeer/>.
+
+> This page shows tealdeer specific functionality. See tldr tldr for more examples.
+
+`tldr --update`
+
+- List all pages in the cache:
+
+`tldr --list`
```
niklasmohrin: These two are already covered in tldr tldr.

pages/tealdeer.md
niklasmohrin, line 1: Very good selection! Once the two examples I mentioned are removed, can we add one that uses the `--config-path` and one that uses the `--no-auto-update` flag?

src/main.rs
```diff
@@ -407,23 +409,27 @@ fn try_main(args: Cli, enable_styles: bool) -> Result<ExitCode> {
...
+            Box::new(Cursor::new(TEALDEER_PAGE.as_bytes()))
+        } else {
+            let Some(result) = cache.find_page(&command) else {
+                if !args.quiet {
+                    print_warning(
+                        enable_styles,
+                        &format!(
+                            "Page `{}` not found in cache.\n\
+                       Try updating with `tldr --update`, or submit a pull request to:\n\
+                       https://github.com/tldr-pages/tldr",
```
niklasmohrin: This does not change the output, but the code looks slightly nicer like this:
```suggestion
                            "Page `{}` not found in cache.\n\
                             Try updating with `tldr --update`, or submit a pull request to:\n\
                             https://github.com/tldr-pages/tldr",
```

src/output.rs
```diff
@@ -30,14 +29,13 @@ fn configure_pager(enable_styles: bool) {
...
-    lookup_result: &PageLookupResult,
+    reader: impl Read,
     enable_markdown: bool,
     enable_styles: bool,
     use_pager: bool,
     config: &Config,
 ) -> Result<()> {
-    // Create reader from file(s)
-    let reader = lookup_result.reader()?;
+    let reader = BufReader::new(reader);
```
niklasmohrin, line 38: This means that we have a BufReader inside a BufReader, because the reader from a PageLookupResult is already in one: https://github.com/tealdeer-rs/tealdeer/blob/6f91c3a765513c1030723e2ab96a8c6d5ea1dca0/src/cache.rs#L285

We can either require `reader: impl BufRead` here or we can change the `.reader` function to just return `Box<dyn Read>` without wrapping it in a BufReader. I think the second option makes more sense because then we will end up with creating a BufReader only right here whereas with the first approach we would create them in `.reader` _and_ for the static page. But either version is fine with me as long as we avoid the nested BufReader

src/main.rs
niklasmohrin, line 1: Can you add a short test in `tests/lib.rs` that checks that the `tldr tealdeer` command works even when no cache has been created? We don't need to check what exactly is printed, just that the output contains something like "for your installed tealdeer version"

niklasmohrin (commented):
Awesome, I think we are very close to the optimum here! I have some small asks left, but then this feature is complete and ready!

ptcodes:
Hi @niklasmohrin, good catch on the nested BufReader! 

I've implemented all the suggestions and hope the PR looks good for the merge now. But let me know if you want to make any changes. Thanks!

src/main.rs
```diff
@@ -258,8 +260,20 @@ fn try_main(args: Cli, enable_styles: bool) -> Result<ExitCode> {
...
-        print_page(&path, args.raw, enable_styles, args.pager, &config)?;
+        let reader = PageLookupResult::with_page(file).reader()?;
+        print_page(reader, args.raw, enable_styles, args.pager, &config)?;
+        return Ok(ExitCode::SUCCESS);
+    }
+
+    // The tealdeer page is embedded in the binary, no cache needed
+    if command == "tealdeer" {
+        print_page(
+            Cursor::new(TEALDEER_PAGE.as_bytes()),
```
niklasmohrin: Since `&[u8]` implements `Read` already, we don't need the `Cursor` here:

```suggestion
            TEALDEER_PAGE.as_bytes(),
```

src/cli.rs
niklasmohrin, line 1: We have one CI failure because the file `docs/src/usage.txt` always needs to contain the current output of `tldr --help`. You can see that just the line you added to the help output is missing here: https://github.com/tealdeer-rs/tealdeer/actions/runs/23876108783/job/70961685019?pr=472

So the fix for this is to just add this line at the end of the file.

src/main.rs
niklasmohrin: Make sure to also remove `Cursor` from the `use` statement above, otherwise CI will fail

niklasmohrin (commented):
Yes, this looks very nice and is way better than what we initially planned in the issue! I have two more small comments, but they are just small nits. If you want, I can also fix these myself before merging, just let me know.

ptcodes:
@niklasmohrin thanks for the thorough review! I've learned a few things from your comments. I've made the fixes.

niklasmohrin (approved):
Thanks! This turned out into something way better than I imagined in the issue text :)

PR #476, opened 2026-04-19
Support tilde (`~`) expansion on config paths

ellsclytn:
Allows directories in the configuration file (cache & custom pages) to be relative to the user's home directory by expanding the common `~` path prefix notation. Works for the current user only (`~diferentUser/` syntax is not supported, and will cause an error if attempted).

Works for Linux/Unix (via `HOME` env var) and Windows (via `USERPROFILE` env var).

Examples (assuming a user called "foo" on a Linux system):

```
~/my/custom-pages # /home/foo/custom-pages
~ # /home/foo
~bar/cache # error
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ path = "src/main.rs"
 anyhow = "1"
 app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
+clean-path = "0.2"
+dirs = "6"
 env_logger = { version = "0.11", optional = true }
 log = "0.4"
 serde = "1.0.21"
--- a/src/config.rs
+++ b/src/config.rs
@@ -592,7 +592,7 @@ impl<'a> Config<'a> {
             }
         } else if let Some(config_value) = &raw_config.directories.cache_dir {
             // Resolve possible ~ prefixed path
-            let expanded_path = expand_path(config_value, home_path.as_ref())?;
+            let expanded_path = expand_home(config_value, home_path.as_ref())?;
             // Resolve possible relative path.
             let resolved_path = relative_path_root.join(expanded_path);
 
@@ -616,7 +616,7 @@ impl<'a> Config<'a> {
             .as_ref()
             .map(|path| -> Result<PathWithSource> {
                 // Resolve possible ~ prefixed path
-                let expanded_path = expand_path(path, home_path.as_ref())?;
+                let expanded_path = expand_home(path, home_path.as_ref())?;
                 // Resolve possible relative path.
                 let resolved_path = relative_path_root.join(expanded_path);
 
@@ -654,7 +654,7 @@ impl<'a> Config<'a> {
 }
 
 /// Expands tilde (~) prefixed directories into its absolute version
-fn expand_path<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Result<Cow<'a, Path>> {
+fn expand_home<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Result<Cow<'a, Path>> {
     if input_path.is_absolute() {
         return Ok(Cow::Borrowed(input_path));
     }
@@ -834,7 +834,7 @@ mod test {
         let path_to_expand = PathBuf::from("~/baz");
 
         assert_eq!(
-            *expand_path(&path_to_expand, home.as_ref()).unwrap(),
+            *expand_home(&path_to_expand, home.as_ref()).unwrap(),
             PathBuf::from("/foo/bar/baz")
         );
     }
@@ -845,7 +845,7 @@ mod test {
         let dir_to_expand = PathBuf::from("/one/two");
 
         assert_eq!(
-            *expand_path(&dir_to_expand, home.as_ref()).unwrap(),
+            *expand_home(&dir_to_expand, home.as_ref()).unwrap(),
             dir_to_expand
         );
     }
@@ -855,7 +855,7 @@ mod test {
         let home = Some(PathBuf::from("/foo/bar"));
         let dir_to_expand = PathBuf::from("~baz/foo");
 
-        assert!(expand_path(&dir_to_expand, home.as_ref()).is_err());
+        assert!(expand_home(&dir_to_expand, home.as_ref()).is_err());
     }
 
     #[test]
--- a/src/config.rs
+++ b/src/config.rs
@@ -659,7 +659,6 @@ fn expand_path<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Resu
         return Ok(Cow::Borrowed(input_path));
     }
 
-    let home_path = home_path.ok_or(anyhow!("Unable to find user home directory"))?;
     let mut components = input_path.components();
 
     if let Some(Component::Normal(first_component_raw)) = components.next() {
@@ -667,16 +666,14 @@ fn expand_path<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Resu
             .to_str()
             .ok_or(anyhow!("Path contains invalid UTF-8"))?;
 
-        if first_component.starts_with("~") {
-            match first_component.len() {
-                1 => {
-                    let rest: PathBuf = components.collect();
-                    let expanded = home_path.join(rest);
+        if first_component == "~" {
+            let home_path = home_path.ok_or(anyhow!("Unable to find user home directory"))?;
+            let rest: PathBuf = components.collect();
+            let expanded = home_path.join(rest);
 
-                    return Ok(Cow::Owned(expanded));
-                }
-                _ => return Err(anyhow!("Tilde expansion with a login name not supported")),
-            }
+            return Ok(Cow::Owned(expanded));
+        } else if first_component.starts_with('~') {
+            return Err(anyhow!("Tilde expansion with a login name not supported"));
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,8 +1,9 @@
 use std::{
+    borrow::Cow,
     env, fmt,
     fs::{self, File},
     io::{ErrorKind, Write},
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
     sync::LazyLock,
     time::Duration,
 };
@@ -589,11 +590,13 @@ impl<'a> Config<'a> {
                 source: PathSource::EnvVar,
             }
         } else if let Some(config_value) = &raw_config.directories.cache_dir {
-            // If the user explicitly configured a cache directory, use that.
+            // Resolve possible ~ prefixed path
+            let expanded_path = expand_path(config_value, env::home_dir().as_ref())?.into_owned();
+            // Resolve possible relative path.
+            let resolved_path = relative_path_root.join(expanded_path);
+
             PathWithSource {
-                // Resolve possible relative path. It would be nicer to clean up the path, but Rust stdlib
-                // does not give any method for that that does not need the paths to exist.
-                path: relative_path_root.join(config_value),
+                path: resolved_path,
                 source: PathSource::ConfigFile,
             }
         } else if let Ok(default_dir) = get_app_root(AppDataType::UserCache, &crate::APP_INFO) {
@@ -610,11 +613,18 @@ impl<'a> Config<'a> {
             .directories
             .custom_pages_dir
             .as_ref()
-            .map(|path| PathWithSource {
+            .map(|path| -> Result<PathWithSource> {
+                // Resolve possible ~ prefixed path
+                let expanded_path = expand_path(path, env::home_dir().as_ref())?.into_owned();
                 // Resolve possible relative path.
-                path: relative_path_root.join(path),
-                source: PathSource::ConfigFile,
+                let resolved_path = relative_path_root.join(expanded_path);
+
+                Ok(PathWithSource {
+                    path: resolved_path,
+                    source: PathSource::ConfigFile,
+                })
             })
+            .transpose()?
             .or_else(|| {
                 get_app_root(AppDataType::UserData, &crate::APP_INFO)
                     .map(|path| {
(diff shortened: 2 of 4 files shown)


src/types.rs
```diff
@@ -220,6 +228,56 @@ impl fmt::Display for PathSource {
     }
 }
 
+fn expand_tilde(s: &str) -> Option<PathBuf> {
```
ellsclytn, line 231: I suspect this isn't the right file for this logic. I left it here because it's easy enough to move, but consider it an open question.

Cargo.toml
```diff
@@ -23,6 +23,8 @@ path = "src/main.rs"
 anyhow = "1"
 app_dirs = { version = "2", package = "app_dirs2" }
 clap = { version = "4", features = ["std", "derive", "help", "usage", "cargo", "error-context", "color", "wrap_help"], default-features = false }
+clean-path = "0.2"
+dirs = "6"
```
niklasmohrin, line 27: I think we can make this work without either of these two dependencies.
- There isn't really a need to clean paths, if you open something like `../` it just works (or not? in that case this is a separate bug)
- To get the home directory we can just use `std::env::home_dir`

src/types.rs
```diff
@@ -220,6 +228,56 @@ impl fmt::Display for PathSource {
...
+    if let Some(rest) = s.strip_prefix("~/") {
```
niklasmohrin, line 232: We probably want to do this replacement using a Path instead of a string, for example Windows users might write `~\Documents\...` and expect it to work. On the other hand, just stripping `~` as a prefix also doesn't work because maybe users write `~joe/...` which usually means "joe's home directory".

That's another thing, there is this `~username` syntax that I don't really want to support, I am not sure how to communicate that though. I guess just not replacing it would work, but it might be frustrating if we also just print `~username` unexpanded and users don't know that we won't expand it

src/config.rs
```diff
@@ -815,6 +851,33 @@ mod test {
...
+        let config = Config::from_raw(
+            &raw_config,
+            PathWithSource {
+                path: PathBuf::from("/path/to/config/config.toml"),
+                source: PathSource::OsConvention,
+            },
+        )
+        .unwrap();
+
+        let home_dir = dirs::home_dir().unwrap();
```
niklasmohrin, line 869: I think the test would be slightly better if we would not query the home path but instead set it in the beginning by setting the HOME or USERPROFILE (Windows) environment variable

src/types.rs
```diff
@@ -220,6 +228,56 @@ impl fmt::Display for PathSource {
...
+}
+
+impl Deref for CleanPath {
+    type Target = PathBuf;
+    fn deref(&self) -> &PathBuf {
+        &self.0
+    }
+}
+
+impl<'de> DeserializeTrait<'de> for CleanPath {
```
niklasmohrin, line 257: I think it would be okay if we just did the expansion at the time we convert a RawConfig to a Config, this is a good place to put such "how to interpret raw values" stuff. I would also be fine if we then did not have the extra type and just used PathBuf/Path

niklasmohrin (commented):
Hi, thanks for submitting a PR :)

I have some comments on the implementation, in particular I think we can keep it a bit simpler than the current suggestion. Please let me know what you think

src/types.rs
niklasmohrin: I think that this is mostly config related logic, so `src/config.rs` seems like a good place

ellsclytn:
The suggestions all make a lot of sense, thanks :)

I'll probably be a few days before I get back into this properly, but I agree with everything you've mentioned.

src/types.rs
ellsclytn: Good points, and that `~username` syntax is particularly tricky. I agree, writing code to support it feels a bit excessive (I suspect the number of people who'd want to be able to use `~username` style for configuring tealdeer would be quite low).

Perhaps we just exit early with an error if that syntax is supplied? Otherwise it would end up making a directory literally called `<tilde><username>` which I think feels worse. What do you think?

niklasmohrin:
Of course, no worries. I can sometimes also take a bit to find time to review, so no rush

Cargo.toml
ellsclytn: Good call. I'd lost track of things a bit here. I'd originally had an idea that it might be nice to have `--show-paths` show the resolved paths, which is where `clean-path` helped. But I'd forgotten to really explain that in the PR. I've pulled it out now.

And wow, I'm amazed I never saw `std::env::home_dir` sooner.

src/types.rs
ellsclytn: I've taken an approach of making the application error if the `~username` syntax is supplied now.

src/config.rs
ellsclytn: I've ended up removing one of the tests in favour of more isolated tests of the `expand_path` function. Admittedly, this alone doesn't feel like a very comprehensive testing strategy, so I've also added a test (`expand_tilde_in_config_file`) that now uses `env::home_dir()` to test that a raw config file with tilde in paths loads properly. My thinking is that these two tests together provide reasonable coverage.

ellsclytn:
Hey @niklasmohrin, I've updated the branch in response to your feedback. Let me know if you have any questions :)

src/config.rs
```diff
@@ -589,11 +590,13 @@ impl<'a> Config<'a> {
                 source: PathSource::EnvVar,
             }
         } else if let Some(config_value) = &raw_config.directories.cache_dir {
-            // If the user explicitly configured a cache directory, use that.
+            // Resolve possible ~ prefixed path
+            let expanded_path = expand_path(config_value, env::home_dir().as_ref())?.into_owned();
```
niklasmohrin: We can remove the `to_owned` here and below, the `.join` creates a new PathBuf anyways; alternatively, we can keep the `to_owned` and use `push` instead of `join`

niklasmohrin: Can we pull out the call to `home_dir` so that we can reuse the result when we use it a second time below?

src/config.rs
```diff
@@ -712,7 +752,14 @@ pub fn get_config_dir() -> Result<(PathBuf, PathSource)> {
...
     if let Ok(value) = env::var("TEALDEER_CONFIG_DIR") {
-        return Ok((PathBuf::from(value), PathSource::EnvVar));
+        let path = PathBuf::from(value);
+
+        // Let this error bubble up: the user has supplied $TEALDEER_CONFIG_DIR, but we couldn't
+        // resolve it. We should exit early instead of loading config from a path that wasn't asked
+        // for.
+        let expanded_path = expand_path(&path, env::home_dir().as_ref())?;
+
+        return Ok((expanded_path.into_owned(), PathSource::EnvVar));
```
niklasmohrin: I wonder whether we should run our expansion on env variables at all. On the one hand it is convenient, on the other hand users might as well use `$HOME` when defining the env var.

I think it's okay to do it, but if we do then we should do it for all env vars, that is, we should also do it TEALDEER_CACHE_DIR

This would then mean that we covered all occurrences of `PathSource::EnvVar` and `PathSource::ConfigFile`; I think we don't need to cover `PathSource::Cli`, such paths should _really_ be expanded by the shell already and if not then users probably really meant to put a literal `~` in there

src/config.rs
```diff
@@ -712,7 +752,14 @@ pub fn get_config_dir() -> Result<(PathBuf, PathSource)> {
     // Allow overriding the config directory by setting the
     // $TEALDEER_CONFIG_DIR env variable.
     if let Ok(value) = env::var("TEALDEER_CONFIG_DIR") {
-        return Ok((PathBuf::from(value), PathSource::EnvVar));
+        let path = PathBuf::from(value);
+
+        // Let this error bubble up: the user has supplied $TEALDEER_CONFIG_DIR, but we couldn't
+        // resolve it. We should exit early instead of loading config from a path that wasn't asked
+        // for.
+        let expanded_path = expand_path(&path, env::home_dir().as_ref())?;
```
niklasmohrin: aaaand, if we also use `home_dir` here, we could consider just saving the result of this globally. we do something similar for the `LANG` and `LANGUAGE` env vars and the same approach should work here, just that for the result of `home_dir` we would use a global static variable

src/config.rs
```diff
@@ -642,6 +652,36 @@ impl<'a> Config<'a> {
     }
 }
 
+/// Expands tilde (~) prefixed directories into its absolute version
+fn expand_path<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Result<Cow<'a, Path>> {
```
niklasmohrin: I think that `expand_tilde` or `expand_home` would make clearer what is going on, otherwise it seems pretty generic when looking at the callsite

src/config.rs
```diff
@@ -642,6 +652,36 @@ impl<'a> Config<'a> {
...
+    if input_path.is_absolute() {
+        return Ok(Cow::Borrowed(input_path));
+    }
+
+    let home_path = home_path.ok_or(anyhow!("Unable to find user home directory"))?;
```
niklasmohrin: We don't need to return this error if no tilde appears in the path; I would move this line down to where the variable is used

src/config.rs
```diff
@@ -642,6 +652,36 @@ impl<'a> Config<'a> {
     }
 }
 
+/// Expands tilde (~) prefixed directories into its absolute version
+fn expand_path<'a>(input_path: &'a PathBuf, home_path: Option<&PathBuf>) -> Result<Cow<'a, Path>> {
+    if input_path.is_absolute() {
```
niklasmohrin: Hm, does this extra check do anything? Wouldn't an absolute path have some non-normal component as a first component anyways?

src/config.rs
```diff
@@ -642,6 +652,36 @@ impl<'a> Config<'a> {
...
+    let home_path = home_path.ok_or(anyhow!("Unable to find user home directory"))?;
+    let mut components = input_path.components();
+
+    if let Some(Component::Normal(first_component_raw)) = components.next() {
+        let first_component = first_component_raw
+            .to_str()
+            .ok_or(anyhow!("Path contains invalid UTF-8"))?;
+
+        if first_component.starts_with("~") {
+            match first_component.len() {
```
niklasmohrin: I think this would be slightly easier like this:
```rust
if first_component == "~" {
  // ...
} else if first_component.starts_with("~") {
  // ...
}
```

src/config.rs
```diff
@@ -712,7 +752,14 @@ pub fn get_config_dir() -> Result<(PathBuf, PathSource)> {
     // Allow overriding the config directory by setting the
     // $TEALDEER_CONFIG_DIR env variable.
     if let Ok(value) = env::var("TEALDEER_CONFIG_DIR") {
-        return Ok((PathBuf::from(value), PathSource::EnvVar));
+        let path = PathBuf::from(value);
```
niklasmohrin: (clippy suggests to change the definition of `expand_path` to take `&Path` instead of `&PathBuf`, so this conversion here would also not be needed anymore)

niklasmohrin (changes requested):
Very cool, I think this is shaping up to be a nice and maintainable feature! :) I have some more comments on details, but overall the approach is good and the places where we use the new helper also make sense

src/config.rs
ellsclytn: I'll drop env var support. It turns out I pursued that due to an error on my part: I was testing in my own shell (Fish) how tilde expansion behaves. I ran something like `export FOO=~/foo`. Turns out that in Fish, this will not expand (instead, Fish has a the [`set`](https://fishshell.com/docs/current/cmds/set.html) command which expands as expected).

I was pretty confused myself, because I was pretty certain ~ gets expanded in shells. And sure enough, it does, if you use the right commands :man_facepalming:.

`export FOO=~/foo` expands as expected in bash, so I now think expanding the env vars within tealdeer is redundant.

src/config.rs
ellsclytn: [Removed env var support](https://github.com/tealdeer-rs/tealdeer/pull/476#discussion_r3166734863) (and fixed up my editor so that Clippy works there now :slightly_smiling_face:)

src/config.rs
ellsclytn: Good call, renamed to `expand_home`.

src/config.rs
ellsclytn: Dropped the `.to_owned`, thanks.

src/config.rs
ellsclytn: It's not strictly necessary. It's just a short circuit because if a path is absolute, we immediately know there's no need to consider any further logic related to expanding `~` and joining paths. Happy to drop it if you prefer.

src/config.rs
ellsclytn: Now that I've [dropped env var expansion support](https://github.com/tealdeer-rs/tealdeer/pull/476#discussion_r3166734863) I'm thinking this isn't so necessary. I could still move it up to be globally reusable if you think it might get used in more places in the future?

ellsclytn:
Hey @niklasmohrin, thanks again! I've fixed up the things you pointed out and left the threads unresolved where I'd like your opinion.

I've added a commit for each piece of feedback, but I'll squash them down before any merging.

src/config.rs
niklasmohrin: I have to say I was also a bit confused because in fish `FOO=~ echo $FOO` expands whereas `FOO="~" echo $FOO` does not; normally expansion is controlled by using single quotes instead of double, but here double quotes also disable it :thinking:

niklasmohrin: Okay, but I agree that not expanding in env vars makes sense :+1:

src/config.rs
niklasmohrin: I agree, if we only have it here then there is no need for a global :+1:

src/config.rs
niklasmohrin: I don't think that this saves any work for the computer, the call to `is_absolute` also does `.components()` which does some parsing etc.; I don't have a strong opinion on whether to keep it though, either way is fine

src/config.rs
```diff
@@ -642,6 +653,33 @@ impl<'a> Config<'a> {
     }
 }
 
+/// Expands tilde (~) prefixed directories into its absolute version
+fn expand_home<'a>(input_path: &'a Path, home_path: Option<&PathBuf>) -> Result<Cow<'a, Path>> {
```
niklasmohrin: The home_path does not need to be an owned PathBuf here
```suggestion
fn expand_home<'a>(input_path: &'a Path, home_path: Option<&Path>) -> Result<Cow<'a, Path>> {
```
When calling this function, it suffices to change `as_ref` to `as_deref`

niklasmohrin (commented):
Everything looks great! I have one more comment, other than that we can merge 👍 

I would use the "squash and merge" option of Github anyways so there is no need to clean up the commits. This would use the PR title as a first line of the commit and the PR description (the first comment) as commit body. If you want, you can change these to a summary of the final changes we arrived at. You don't have to though, I will also make sure that the commit that lands on main has a sensible commit message etc. when merging :)

src/config.rs
ellsclytn: Good point, I hadn't bothered to dive into the underlying implementation of `is_absolute()`, I guess I just assumed they'd be doing something more efficient than me. But you're right, they're calling `.components()` just like I am, so I'll drop this bit.

src/config.rs
ellsclytn: Oops, good catch!

ellsclytn:
Thanks again, have addressed the comments and have adjusted the PR description.

niklasmohrin (approved):
Great, thank you!

PR #480, opened 2026-05-03
Remove `tldr-c` entry from benchmark results in README.md

xtqqczze:
`tldr-c-client` is unmaintained.


niklasmohrin (changes requested):
Can you also remove the line for the link `c-gh` link reference?

niklasmohrin (approved):

niklasmohrin:
Thanks!

xtqqczze:
It would be good to update the benchmarks at some point to include [`tlrc`](https://github.com/tldr-pages/tlrc), which is the replacement for `tldr-c`.

PR #481, opened 2026-05-03
Fix clippy lints on all targets

xtqqczze:


.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
            toolchain: stable
            components: clippy
        - name: run clippy lints
-         run: cargo clippy --features logging
+         run: cargo clippy --workspace --all-targets --features logging
```
niklasmohrin: Does `--workspace` do anything for us here?

xtqqczze: Since the repository currently contains only a single crate, it isn’t necessary at the moment, but it helps future-proof the workflow in case that changes later.

niklasmohrin: Yeah, I don't think we need to future-proof our ci config, this is just noise. Let's remove this

niklasmohrin:
Thanks :)

PR #483, opened 2026-06-02
feat: set a `User-Agent` when fetching

CosmicHorrorDev:
this sets an identifiable `User-Agent` when fetching assets for the cache, over the default `ureq/{VERSION}` value

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -351,6 +351,11 @@ impl Cache<'_> {
         let config = Agent::config_builder()
             .http_status_as_error(false) // because we want to handle them
             .tls_config(tls_builder.build())
+            .user_agent(concat!(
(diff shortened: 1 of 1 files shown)


niklasmohrin:
Hi @CosmicHorrorDev, what is the motivation for this change? We never had anyone complain about the ureq user agent and it does correctly reflect what implementation is performing the http request

CosmicHorrorDev:
it's good practice in general. the default user-agent isn't very identifying given that anything using that version of `ureq` shares it (if somone did have a problem then it would be hard for them to identify that e.g. `tealdeer` was the cause if all they recieve is `ureq/{VERSION}` of course)

CosmicHorrorDev:
i don't really have the energy to go through all of this at the moment. i'll just avoid contributing going forward

PR #484, opened 2026-06-02
deps: update dependencies

CosmicHorrorDev:
_update: takes bumping the MSRV a bit_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


niklasmohrin:
Hi @CosmicHorrorDev, thanks for the effort, but I will update the dependencies before creating a new release of tealdeer. Until then, I don't see a point in updating anything unless we want to use the new dependency versions for something. Do you have an existing issue in mind that needs these updates?

CosmicHorrorDev:
ah, updating dependencies (particularly including semver breaking changes) just before the release goes against the norm of what I'm used to, but if that's how you want to handle things then sure

PR #488, opened 2026-06-06
fix: emit indentation whitespace without style

algojogacor:
This is an independent contribution and is not associated with any hackathon or competition.

## Summary

Fixes indentation whitespace being styled with user-configured text styles like underline or background color.

## Root Cause

When the configurable indentation feature was introduced in #471, indentation whitespace was emitted as styled snippet types (e.g., `PageSnippet::NormalCode` for example code indentation). This caused styles configured in `[style.example_code]` — such as underline — to apply to the leading whitespace rather than only the text content.

## Fix

Introduced a `PageSnippet::Indent` variant that carries indentation whitespace. The renderer (`print_snippet`) writes this variant as plain text without any style. This restores the pre-#471 behavior where indentation was always plain.

## Testing

- All 51 existing tests pass.
- Verified output manually: indentation whitespace no longer carries ANSI styling codes.


niklasmohrin:
Someone else already submitted a PR for this in #487

PR #491, opened 2026-06-21
[codex] Render indents without styles

### LINKED ISSUE #485: Indents in output should use plain style
When implementing configurable indents in #471, we changed how indentation in the output is handled. Before, we would just print plain space characters followed by styled example code or text. Now, we emit a separate snippet of the corresponding type that just contains whitespace. As an example, by using the config
```toml
[style.example_code]
underline = true
```
and running `tldr tar`, we can see that the indents are underlined.

We should revert this behavior so that the indents are not styled. To do so, a new PageSnippet, `PageSnippet::Indent` should be introduced to trigger the printing of unstyled indents.

A69Nacional:
## Summary

- add the dedicated `PageSnippet::Indent` variant requested in #485
- render indent snippets directly instead of applying content styles
- add regression coverage for snippet generation and output
- update colored-output fixtures so leading spaces are outside ANSI style sequences

Closes #485.

## Root cause

Configurable indentation was emitted with the surrounding title, description, text, or code snippet type. That made whitespace inherit styles such as underline or background color.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --locked --all-targets --features logging -- -D warnings`
- 29/29 unit tests pass
- 48/51 integration tests pass on Windows, including every rendering and configuration test

The remaining three integration tests are unrelated editor tests that set `EDITOR=touch`; Windows cannot find the Unix `touch` executable. The project CI runs those tests on Ubuntu.


niklasmohrin:
Another PR (#487) is already open for this issue

PR #492, opened 2026-06-29
Add `updates.warn_cache_age` config option

karaolidis:
Adds a boolean `[updates] notify_outdated_cache` option (default `true`) to suppress the "The cache hasn't been updated for N days" warning.

This is useful when the cache is managed externally, e.g. provisioned from a Nix store path or by a package manager, where the directory's mtime doesn't reflect the cache's real age, causing a spurious warning on every invocation. Today the only way to silence it is `--quiet`, which must be passed every call and hides all other output too.

### DIFF
--- a/docs/src/config_updates.md
+++ b/docs/src/config_updates.md
@@ -30,6 +30,18 @@ auto_update = true
 auto_update_interval_hours = 24
 ```
 
+### `warn_cache_age`
+
+Controls when a warning is printed if the cache has not been updated in a while.
+By default, the warning is shown once the cache is older than 30 days. Set this
+to `"never"` to silence the warning. This is useful if, for some reason, the
+modification time does not reflect its actual age.
+
+```toml
+[updates]
+warn_cache_age = "never"
+```
+
 ## Download configuration
 
 ### `download_languages`
--- a/src/config.rs
+++ b/src/config.rs
@@ -228,6 +228,17 @@ fn default_archive_source() -> String {
     "https://github.com/tldr-pages/tldr/releases/latest/download".to_owned()
 }
 
+/// Controls when a warning about an outdated cache is printed.
+///
+/// Currently, the only nameable option is `"never"`. In the future, this may
+/// be extended to also accept a duration (e.g. `"60d"`), after which the
+/// warning should be shown.
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+enum RawWarnCacheAge {
+    Never,
+}
+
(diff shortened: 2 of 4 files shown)


niklasmohrin:
Hi @karaolidis, thank you for the contribution. Your changes look good, I have two thoughts:
1. I am unsure whether we want to make the duration itself configurable at some point. If so, it would be nice to have only one config option instead of two. With this in mind, I think a nicer naming would be something like `updates.warn_outdated_cache_age = "never"` (let's say, for now, that "never" is the only nameable option, and we add support for putting something like "60d" later; this should be doable by having an enum with only one option and putting `Option<TheEnum>` in rawconfig, the config can then have `Option<Duration>` or so). What do you think? And can you think of a better name than `warn_outdated_cache_age`, one that signals that usually a duration should be inserted, but were something like "never" also makes sense? Maybe `warn_cache_age` is better?
2. So far, I have been reluctant to bless any setup where the page cache is not managed by tealdeer itself (see https://github.com/tldr-pages/tldr/issues/19775#issuecomment-3696734687 and the following comments for some of my thoughts). Thus, I would cut back on the description in the docs a bit. Maybe something like "This is useful if, for some reason, the modification time does not reflect its actual age".

Can you tell me a bit more about this nix setup for the page cache? Do you just have a pinned clone of the repo in your home manager config? And is this something that is promoted somewhere and that people are using?

If you need any help in applying the changes, please let me know! I am looking forward to hear what you think :)

karaolidis:
Hey @niklasmohrin, Thanks for taking a look at this!

For the first point, I agree. I could implement this as part of this PR if we decide to go forward with it? The description changes sound reasonable as well.

You can find my tealdeer setup here: https://git.karaolidis.com/karaolidis/nix/commit/cb2067d147dd21318c8554a63e0e9f24b4dff280

niklasmohrin:
I see. I would still say that instantiating the cache like this is not officially supported, but I can see why you would want to do it like this. Regardless, let's add this change and maybe if I see more people trying to initialize tealdeer caches like this, I can start to think about how I want this to work officially. (Maybe, at some point, I will add a `tldr --update --archive-source 'file://...'` option to generate the cache in nix while still allowing tealdeer to control the layout of the cache directory; this would then be the blessed way of doing this. The alternative is to guarantee that tealdeer works with the layout of the zip directly, which also opens possibilities like more than one cache location using XDG_DATA_DIRS or so. Not sure about this though, and not relevant to this PR).

For the naming, let's go with `updates.warn_cache_age = "never"`, I think this avoids the redundant naming of `warn_outdated_cache_age` and still has all important words in there. Support for `warn_cache_age = "60d"` or should be added later.

niklasmohrin (approved):
Thanks, very nice feature :)

PR #494, opened 2026-07-19
fix: print indents without styling

### LINKED ISSUE #485: Indents in output should use plain style

deeferentleeg:
## Description

Closes #485.

When configurable indents were introduced in #471, the indentation of every line was emitted as a *styled* snippet of the surrounding line type (`Description`, `Text`, `NormalCode` or `Title`). As a result, configuring a style for one of those types also styled the leading indentation. For example, with

```toml
[style.example_code]
underline = true
```

running `tldr tar` would underline the indents as well as the code.

This PR reverts that behavior so that indents are never styled, as requested in the issue. It introduces a new `PageSnippet::Indent` variant that `highlight_lines` emits for the leading whitespace of each line. `print_snippet` renders `Indent` as plain, unstyled text, while all other snippets continue to be styled as before.

### Changes

- Added `PageSnippet::Indent(T)` and handled it in `map`, `PartialEq`, `is_empty` and `print_snippet`.
- `highlight_lines` now emits `PageSnippet::Indent(&base_indent)` / `PageSnippet::Indent(&command_indent)` instead of a typed snippet for the indentation of `Title`, `Description`, `ExampleText` and `ExampleCode` lines.
- Updated the rendered integration-test fixtures (`inkscape-default`, `inkscape-with-config`, `inkscape-with-title`, `apt.ja`) so the expected output no longer contains escape codes for the indents. The no-color fixtures were already unstyled and are unchanged.
- Added unit tests in `formatter::tests::highlight_lines` verifying that indents are emitted as `Indent` snippets, that the title indent is unstyled, and that a zero-width indent produces no output.

### Notes

- The behavior in `--color never` / no-color mode is unchanged: indents were already plain there because all styles default to unstyled.
- Empty indents (when `indent.base` / `indent.command` is `0`) are still filtered out by the existing `is_empty` check in the output closure, preserving previous behavior.

## AI usage disclosure

An AI assistant was used to help locate the relevant code and draft this change. I have reviewed and understand every line of the diff, and I am the sole author of this contribution.


niklasmohrin:
Comment violates AI policy

51 PRs on the same day, so strong doubts on your authorship declaration

PR #503, opened 2026-08-16
ci: build with `--locked`

meop:
## Summary

- add `--locked` to feature builds, tests, Clippy, and documentation generation
- add `--locked` to Linux, macOS, and Windows release builds
- ensure CI and published binaries use the committed dependency resolution

The project is a binary application with a committed `Cargo.lock`, so silently resolving a different dependency graph on a runner can hide manifest/lockfile drift and make releases less reproducible.

## Validation

- `cargo check --locked --all-targets --features logging`
- `cargo check --locked --features native-tls --no-default-features`

Prepared with assistance from OpenAI Codex; the commit includes a `Co-authored-by` trailer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,23 +28,23 @@ jobs:
       - run: mkdir artifacts
       - name: Build with default features
         run: |
-          cargo build
+          cargo build --locked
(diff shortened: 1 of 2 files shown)


niklasmohrin (approved):
Makes sense, thanks :)

PR #516, opened 2026-08-22
Add warn about unknown config keys

### LINKED ISSUE #511: Warn about unknown config keys
When deserializing the config file and when applying config overrides with `--override-config`, a warning should be printed if there are keys that do not map to actual config values. After a quick search, it seems that https://github.com/dtolnay/serde-ignored is a nice way to implement this, but I am not sure whether it is the best option at this time.

gui-alb:
Should close #511 

- Added `deny_unknown_fields` attribute to the config structs. Unknown configuration keys now result in an error
- Added two test cases: one covering the config.toml file and another covering `override-config`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -140,6 +140,7 @@ impl From<RawColor> for Color {
 }
 
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 struct RawStyle {
     pub foreground: Option<RawColor>,
     pub background: Option<RawColor>,
@@ -193,6 +194,7 @@ impl From<RawStyle> for Style {
 }
 
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 struct RawStyleConfig {
     #[serde(default)]
     pub description: RawStyle,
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -1022,6 +1029,19 @@ mod test {
...
 
+    #[test]
+    fn deny_unknown_config_keys() {
+        let config = "
+            [display]
+            compact = true
+            unknown = false
+            ";
+
+        let result: Result<RawConfig, _> = toml::from_str(config);
```
niklasmohrin: I think it would be more meaningful to test this as an integration test in tests/lib.rs; for this "easy" case here, I think we can trust that the attribute works

(same thing for the other test)

src/config.rs
niklasmohrin, line 1: Is there any way that we can specify `deny_unknown_fields` only once?

niklasmohrin (commented):
Looks mostly good :)

src/config.rs
gui-alb: i dont think so, if we add it only to `RawConfig` it wouldnt detect `style.idk` for example, so we need to specify it in every struct.

niklasmohrin: Makes sense!

niklasmohrin (approved):
Thanks, everything works as expected!

I added an additional check to the tests to ensure that we are failing _because of the unknown field_ :)

PR #522, opened 2026-09-15
List embedded tealdeer page in `--list` output

### LINKED ISSUE #521: Include `tealdeer` in `tldr -l`
Now that tealdeer renders an embedded page for itself, it would be nice if completions would pick up on it, which can be done if `tldr -l` lists it as well (currently it seems to be limited to listing pages present in cache).

ethanstoner:
Closes #521

AI disclosure: the code change, test updates, and this description were written with Claude Code (AI). I reviewed the diff and ran the checks below.

`tldr --list` only listed pages from the cache, so the embedded `tealdeer` page was missing. It is now added to the list (sorted, deduplicated). The exact-output `--list` tests were updated to include it.

Checked with `cargo fmt --check`, `cargo clippy --all-targets` (no warnings) and `cargo test` (all pass).


ethanstoner:
(AI-written comment) Withdrawing this: it was made with AI on a good-first-issue, which the README's AI policy asks to leave for new contributors. Sorry for the noise.

